PR #1, opened 2022-02-11
Bring in basic (mostly empty), fully-themed site

listx:
The reason why there are so many commits is because I merged in the Docsy (Hugo) theme's starter website (the recommended way to install/use the theme). The merge made it easier to bring over the submodules that come with the Docsy theme.

I went ahead and removed some of the built-in docs pages (see the commits near HEAD), and did a rough s/docsy/prow replacement to make the site make more sense. It also includes the `netlify.toml` from https://github.com/kubernetes/test-infra/pull/25042 with a minor tweak in the domain name.

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -1,0 +1,4 @@
+baseURL = 'http://example.org/'
+languageCode = 'en-us'
+title = 'My New Hugo Site'
+theme = "docsy"
--- a/site/config.toml
+++ b/site/config.toml
@@ -112,12 +112,12 @@ version = "0.0"
 
 # A link to latest version of the docs. Used in the "version-banner" partial to
 # point people to the main doc site.
-url_latest_version = "https://example.com"
+url_latest_version = "https://docs.prow.k8s.io"
 
 # Repository configuration (URLs for in-page links to opening issues and suggesting changes)
 github_repo = "https://github.com/kubernetes-sigs/prow"
 # An optional link to a related project repo. For example, the sibling repository where your product code lives.
-#github_project_repo = "https://github.com/google/docsy"
+#github_project_repo = "https://github.com/kubernetes/test-infra"
 
 # Specify a value here if your content directory is not in your repo's root directory
 # github_subdir = ""
--- a/site/config.toml
+++ b/site/config.toml
@@ -117,7 +117,7 @@ url_latest_version = "https://docs.prow.k8s.io"
 # Repository configuration (URLs for in-page links to opening issues and suggesting changes)
 github_repo = "https://github.com/kubernetes-sigs/prow"
 # An optional link to a related project repo. For example, the sibling repository where your product code lives.
-#github_project_repo = "https://github.com/kubernetes/test-infra"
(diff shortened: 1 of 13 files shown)


mrbobbytables:
You'll need to be a bit more selective with the files you want to copy over, the docsy examples includes things like the Google CLA info and Google copyrighted files (bash script). :x

listx:
On second thought I think I should just bring in the theme as a git submodule itself. That way when the theme updates we can also update by just updating the theme's commit SHA.

listx:
OK I just imported the theme as a submodule (so if you want to try out the site locally, you have to run the `make update-theme`) rule, followed by `make serve`.

site/config.toml
```diff
@@ -0,0 +1,194 @@
...
+version = "0.0"
+
+# A link to latest version of the docs. Used in the "version-banner" partial to
+# point people to the main doc site.
+url_latest_version = "https://example.com"
+
+# Repository configuration (URLs for in-page links to opening issues and suggesting changes)
+github_repo = "https://github.com/kubernetes-sigs/prow"
+# An optional link to a related project repo. For example, the sibling repository where your product code lives.
+#github_project_repo = "https://github.com/google/docsy"
```
cjwagner, line 120: Sounds like this could be https://github.com/kubernetes/test-infra?

site/config.toml
```diff
@@ -0,0 +1,194 @@
...
+archived_version = false
+
+# The version number for the version of the docs represented in this doc set.
+# Used in the "version-banner" partial to display a version number for the
+# current doc set.
+version = "0.0"
+
+# A link to latest version of the docs. Used in the "version-banner" partial to
+# point people to the main doc site.
+url_latest_version = "https://example.com"
```
cjwagner, line 115: Should this be https://docs.prow.k8s.io/?

cjwagner (approved):
I'm not familiar with the tools in use here so I can't be real confident in my review, but this looks ready to try out to me. I have a couple nits/questions about some links, feel free to `/unhold` if the existing values are right or should be updated later.
/hold

site/config.toml
listx: Done

site/config.toml
listx: Done

site/config.toml
```diff
@@ -112,12 +112,12 @@ version = "0.0"
...
 # A link to latest version of the docs. Used in the "version-banner" partial to
 # point people to the main doc site.
-url_latest_version = "https://example.com"
+url_latest_version = "https://docs.prow.k8s.io"
 
 # Repository configuration (URLs for in-page links to opening issues and suggesting changes)
 github_repo = "https://github.com/kubernetes-sigs/prow"
 # An optional link to a related project repo. For example, the sibling repository where your product code lives.
-#github_project_repo = "https://github.com/google/docsy"
+#github_project_repo = "https://github.com/kubernetes/test-infra"
```
cjwagner, line 120: This is still commented out.

cjwagner (approved):

PR #3, opened 2022-03-01
fix netlify preview

listx:
The "site" folder does not need to specified again for the "publish"
field because it is relative to the "base" field, as per the docs [1]:

    # Settings in the [build] context are global and are applied to
    # all contexts unless otherwise overridden by more specific contexts.
    [build]
      # Directory to change to before starting a build.
      # This is where we will look for package.json/.nvmrc/etc.
      # If not set, defaults to the root directory.
      base = "project/"

      # Directory that contains the deploy-ready HTML files and
      # assets generated by the build. This is relative to the base
      # directory if one has been set, or the root directory if
      # a base has not been set. This sample publishes the directory
      # located at the absolute path "root/project/build-output"

      publish = "build-output/"

[1]: https://docs.netlify.com/configure-builds/file-based-configuration/

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,6 +1,6 @@
 [build]
 base = "site/"
-publish = "site/public/"
+publish = "public/"
 command = "hugo"
 
 [build.environment]
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,7 +1,7 @@
 [build]
 base = "site/"
 publish = "public/"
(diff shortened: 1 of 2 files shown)


listx:
Looks like there are some other issues with the build (it's broken at the moment). Will dig into this and fix/add onto this PR.

cjwagner (approved):

PR #7, opened 2022-04-15
add architecture diagram

listx:
This also enables mermaidjs (which comes bundled with the Docsy theme that we use). While we're at it, we also update the Docsy theme itself.

/cc @mpherman2

### DIFF
--- a/site/content/en/docs/Overview/architecture.md
+++ b/site/content/en/docs/Overview/architecture.md
@@ -1,0 +1,83 @@
+---
+title: "Architecture"
+linkTitle: "Architecture"
+weight: 1
+description: >
+  Prow is made up of a collection of microservices (aka "Prow components") that work together in a **service cluster**, leveraging one or more **build clusters** to schedule Prow Jobs (or just "jobs").
+---
+
+## Prow in a Nutshell
+
+Prow creates jobs based on various types of events, such as:
+
+- GitHub events (e.g., a new PR is created, or is merged, or a person comments
+"/retest" on a PR),
+
+- Pub/Sub messages,
+
+- time (these are created by **Horologium** and are called **periodic** jobs),
+and
+
+- retesting (triggered by **Tide**).
+
+Jobs are created inside the Service Cluster as Kubernetes Custom Resources.  The
+**Prow Controller Manager** takes triggered jobs and schedules them into a build
+cluster, where they run as Kubernetes pods. **Crier** then reports the results
+back to GitHub.
+
+```mermaid
+flowchart TD
+
+    classDef yellow fill:#ff0
+    classDef cyan fill:#0ff
+    classDef pink fill:#f99
+
+    subgraph Service Cluster["<span style='font-size: 40px;'><b>Service Cluster</b></span>"]
+        Deck:::cyan
+        Prowjobs:::yellow
+        Crier:::cyan
+        Tide:::cyan
+        Horologium:::cyan
+        Sinker:::cyan
+        PCM[Prow Controller Manager]:::cyan
+        Hook:::cyan
+            subgraph Hook
+                WebhookHandler["Webhook Handler"]
+                PluginCat(["'cat' plugin"])
+                PluginTrigger(["'trigger' plugin"])
+            end
+    end
+
+    subgraph Build Cluster[<b>Build Cluster</b>]
+        Pods[(Pods)]:::yellow
+    end
+
+    style Legend fill:#fff,stroke:#000,stroke-width:4px
+    subgraph Legend["<span style='font-size: 20px;'><b>LEGEND</b></span>"]
+        direction LR
+        k8sResource[Kubernetes Resource]:::yellow
+        prowComponent[Prow Component]:::cyan
+        hookPlugin([Hook Plugin])
+        Other
+    end
+
+    Prowjobs <-.-> Deck <-----> |Serve| prow.k8s.io
+    GitHub ==> |Webhooks| WebhookHandler
+    WebhookHandler --> |/meow| PluginCat
+    WebhookHandler --> |/retest| PluginTrigger
+    Prowjobs <-.-> Tide --> |Retest and Merge| GitHub
+    Horologium ---> |Create| Prowjobs
+    PluginCat --> |Comment| GitHub
+    PluginTrigger --> |Create| Prowjobs
+    Sinker ---> |Garbage collect| Prowjobs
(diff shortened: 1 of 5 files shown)


site/content/en/docs/Overview/architecture.md
```diff
@@ -0,0 +1,74 @@
...
+title: "Architecture"
+linkTitle: "Architecture"
+weight: 1
+description: >
+  Prow is made up of a collection of microservices (aka "Prow components") that work together in a **service cluster**, leveraging one or more **build clusters** to schedule Prow Jobs (or just "jobs").
+---
+
+## Prow in a Nutshell
+
+Prow creates **Prow Jobs** ("jobs") based on either GitHub events (e.g., a new
```
cjwagner, line 11: "on either GitHub events" or what else?
Other triggering events include gerrit events, pubsub messages, periodic jobs, and tide retesting.

site/content/en/docs/Overview/architecture.md
```diff
@@ -0,0 +1,74 @@
...
+PR is created, or is merged, or a person comments "/retest" on a PR). These jobs
+are created inside the Service Cluster as Kubernetes Custom Resources. The
+**Prow Controller Manager** takes pending jobs and schedules them into a build
```
cjwagner, line 14:
```suggestion
**Prow Controller Manager** takes triggered jobs and schedules them into a build
```

site/content/en/docs/Overview/architecture.md
```diff
@@ -0,0 +1,74 @@
...
+    PCM -.-> |List and update| Prowjobs
+    PCM ---> |Report| Prowjobs
+    PCM ==> |Create and Query| Pods
+    Prowjobs <-.-> |Inform| Crier --> |Report| GitHub
+```
+
+## Notes
+
+Note that Prow can also work with Gerrit, not just GitHub as in the diagram
+above.
```
cjwagner, line 74: I think this is misleading. Most of the interactions with GitHub shown in the diagram are not supported with Gerrit. Specifically, neither Tide nor Hook work with Gerrit yet.

site/content/en/docs/Overview/architecture.md
```diff
@@ -0,0 +1,74 @@
...
+    hookPlugin([Hook Plugin])
+    Other
+    end
+
+    Prowjobs <-.-> Deck -----> |Serve| prow.k8s.io
+    GitHub ==> |Webhooks| Hook
+    Prowjobs <-.-> Tide --> |Retest and Merge| GitHub
+    Horologium ---> |Create| Prowjobs
+    Hook --> |/meow| PluginCat
+    Hook --> |/retest| PluginTrigger
```
cjwagner, line 60: nit: What about moving the plugins inside the "Hook" box to indicate they are part of the hook component? Doesn't matter much.

cjwagner:
Just nits, this looks like a good detail level for introducing Prow, thanks!

site/content/en/docs/Overview/architecture.md
listx: Reworded.

site/content/en/docs/Overview/architecture.md
listx: Done

site/content/en/docs/Overview/architecture.md
listx: Done

site/content/en/docs/Overview/architecture.md
listx: On second thought, doesn't Hook work with Gerrit? People can comment `/retest` and jobs will get triggered there, right?

site/content/en/docs/Overview/architecture.md
```diff
@@ -0,0 +1,82 @@
...
+        hookPlugin([Hook Plugin])
+        Other
+    end
+
+    Prowjobs <-.-> Deck <-----> |Serve| prow.k8s.io
+    GitHub ==> |Webhooks| Hook
+    Prowjobs <-.-> Tide --> |Retest and Merge| GitHub
+    Horologium ---> |Create| Prowjobs
+    GitHub ---> |/meow| PluginCat
+    GitHub --> |/retest| PluginTrigger
```
cjwagner, line 68: These comments are not received separately from webhooks as currently shown and they are routed to all plugins, not just the ones that react.
To indicate these are webhooks I think it would be clearer to have the lines start at the `Webhooks` box rather than `GitHub`. A dotted line or different color might be better than a solid line too since these are examples?

cjwagner (approved):
> On second thought, doesn't Hook work with Gerrit? People can comment `/retest` and jobs will get triggered there, right?

Hook only handles GitHub webhooks, Gerrit does not send webhooks. The Gerrit adapter scrapes Gerrit instances and is responsible for triggering jobs based on new commits or job triggering comments like `/retest`.
Some code is shared for the comment matching and determining which job to run, but the components are otherwise quite different.

I have one remaining bit of feedback, but it isn't critical.
/hold

site/content/en/docs/Overview/architecture.md
listx: To clarify, you'd prefer the lines to look more like this?

```
    GitHub --> Webhook ---> |/meow| PluginCat
    GitHub --> Webhook --> |/retest| PluginTrigger
```

Aside: currently in the diagram the `Webhooks` is not a node (box) but rather just a labeled edge, so that would need to change as well.

BTW this architecture diagram is mostly copied from one of the existing architecture diagrams I found internally.

> > On second thought, doesn't Hook work with Gerrit? People can comment `/retest` and jobs will get triggered there, right?
> 
> Hook only handles GitHub webhooks, Gerrit does not send webhooks. The Gerrit adapter scrapes Gerrit instances and is responsible for triggering jobs based on new commits or job triggering comments like `/retest`. Some code is shared for the comment matching and determining which job to run, but the components are otherwise quite different.
> 
> I have one remaining bit of feedback, but it isn't critical. /hold

Makes sense

cjwagner: > To clarify, you'd prefer the lines to look more like this?

That seems closer to reality to me. I mainly just don't like that the diagram currently shows `/meow` and `/retest` as completely separate from (and parallel to) webhooks.

listx: Done

listx:
@cjwagner Can you lgtm as well?

PR #10, opened 2022-05-31
Migration

listx:
/assign @cjwagner 

This builds on #9.


listx:
I think I'll add comments in the original files in kubernetes/test-infra/prow/*.md so that people know to not modify those old files (and instead update them here). Another alternative is to simply delete the original files, but that seems like it will be a very disruptive transition.

site/content/en/docs/Legacy Snapshot/migrate.sh
```diff
@@ -0,0 +1,122 @@
...
+    echo "Finished migrating files."
+}
+
+function main() {
+    if [[ -z "${1:-}" ]]; then
+        echo >&2 "Usage: $0 [kubernetes/test-infra directory]"
+    fi
+
+    migrate "${1}"
+    add_warnings "${1}"
```
cjwagner, line 119: This does not appear to be defined anywhere.

site/content/en/docs/Legacy Snapshot/_index.md
```diff
@@ -0,0 +1,21 @@
...
+## Legacy Snapshot
+
+Due to historical reasons, almost all of Prow's documentation resides in the
+[kubernetes/test-infra](k/t-i) repo. As part of the [migration
+effort](https://github.com/kubernetes-sigs/prow/issues/4), this page captures
+Prow documentation in the [kubernetes/test-infra](k/t-i) repository as of [May
+31,
+2022](https://github.com/kubernetes/test-infra/tree/28269f0ce52487379134afe7738393557d577a58).
+All Markdown files have been copied in along with their directory structure
+here.
```
cjwagner, line 18: I think you mentioned your intentions for these files to me, can we capture that here or in the "Contribution Guidelines" page? Specifically when one of the legacy docs needs to be updated how would we like that to happen? Should we move it out of the legacy snapshot directory at that time?

cjwagner (approved):
/hold

site/content/en/docs/Legacy Snapshot/migrate.sh
listx: Done.

site/content/en/docs/Legacy Snapshot/_index.md
listx: > I think you mentioned your intentions for these files to me, can we capture that here or in the "Contribution Guidelines" page?

Done (added instructions to Contribution Guidelines).

> Specifically when one of the legacy docs needs to be updated how would we like that to happen? Should we move it out of the legacy snapshot directory at that time?

Yes, we should move it out of Legacy Snapshot and move it to a more appropriate location (the instructions in the Contribution Guidelines say this now).

cjwagner (approved):

PR #17, opened 2022-08-24
Build cluster setup demystified

listx:
Putting this up for initial review before I add too much (incorrect) technical details. I will probably also add a (simplified?) diagram WRT `gencred` and `kubernetes-external-secrets`, but that can come later.

/cc @chaodaiG


site/content/en/docs/build-clusters.md
```diff
@@ -0,0 +1,146 @@
+---
+title: "GKE Build Clusters"
```
chaodaiG, line 2: Note that for prow, a build cluster could be any k8s cluster, could be GKE, AKS, or EKS. I think it's fine to only document GKE build cluster for now. But should also keep in mind that build cluster from other providers might also be documented by someone else.

chaodaiG (commented):
Overall LGTM, left two comments

site/content/en/docs/build-clusters.md
```diff
@@ -0,0 +1,146 @@
...
+freely create, update, and delete jobs (pods) in your build cluster as
+necessary.
+
+In order for Prow's components to authenticate as the cluster-admin KSA in your
+build cluster, they use a token. You can think of this token simply as a
+password, that when provided to your build cluster's API server along with the
+KSA name, grants authentication as this very same KSA.
+
+The question now turns to how we can generate this token. This can be done
+manually, but the token cannot be valid for longer than X days for security
```
listx, line 59: @chaodaiG Do you know how long this is? was it 2 days?

site/content/en/docs/build-clusters.md
listx: Exactly.

listx:
@chaodaiG @cjwagner Can you have a double-take on for the diagram preview here? https://deploy-preview-17--k8s-prow.netlify.app/docs/build-clusters/

site/content/en/docs/build-clusters.md
chaodaiG: 2 days max

chaodaiG: To be more accurate, 2 days for GKE, the value might be different for other k8s providers

chaodaiG:
> @chaodaiG @cjwagner Can you have a double-take on for the diagram preview here? https://deploy-preview-17--k8s-prow.netlify.app/docs/build-clusters/#grant-gcs-access-for-your-jobs-with-workload-identity

might just be me, although knowing the contexts already, it still feels not very easy to understand the diagram. what do you think about describing the sequence of events with bullets? something like:

1. KSA associated short-lived token generated by gencred
2. token mount onto Prow components, such as PCM
3. PCM runs `kubectl apply` equivalent that is authenticated with the token

listx:
> Member

I can add those things, although I don't think adding numbers to indicate the sequence is a good idea (because not all of the arrows should be numbered).

listx:
@chaodaiG QQ: why do we need this line https://github.com/kubernetes/test-infra/blob/a25542c55d8452bc470f847421c53a6b1710f84c/prow/create-build-cluster.sh#L270 ? Is it because the `gencred` binary needs it when it runs inside the `ci-test-infra-gencred-refresh-kubeconfig` job inside the trusted cluster? If so, does gencred spin up a new container inside the user-provided build cluster to generate the 2-day token and store it in the user's GCP Secret Manager?

chaodaiG:
> @chaodaiG QQ: why do we need this line https://github.com/kubernetes/test-infra/blob/a25542c55d8452bc470f847421c53a6b1710f84c/prow/create-build-cluster.sh#L270 ? Is it because the `gencred` binary needs it when it runs inside the `ci-test-infra-gencred-refresh-kubeconfig` job inside the trusted cluster?

yes

> If so, does gencred spin up a new container inside the user-provided build cluster to generate the 2-day token and store it in the user's GCP Secret Manager?

It does not. `gencred` interacts directly with the api server of build clusters to generate the token

listx:
Can you go into a little more detail about why the gencred-refresher GSA needs `--role="roles/container.admin"` on the build cluster's GCP project? For example, what container does it need to create/delete for it to run?

chaodaiG:
> Can you go into a little more detail about why the gencred-refresher GSA needs `--role="roles/container.admin"` on the build cluster's GCP project? For example, what container does it need to create/delete for it to run?

I see the confusion, this is one of the predefined GKE roles for managing GKE https://cloud.google.com/kubernetes-engine/docs/how-to/iam#predefined, they are recommended roles for interacting with GKE clusters

listx:
> I see the confusion, this is one of the predefined GKE roles for managing GKE https://cloud.google.com/kubernetes-engine/docs/how-to/iam#predefined, they are recommended roles for interacting with GKE clusters

IIUC, `gencred` binary needs to make the build cluster's K8s API server generate a new token for the "root" (cluster-admin KSA role) KSA on the build cluster. `gencred` does this by running as a GSA that has the `roles/container.admin` role defined by GCP. Is this correct?

chaodaiG:
> > I see the confusion, this is one of the predefined GKE roles for managing GKE https://cloud.google.com/kubernetes-engine/docs/how-to/iam#predefined, they are recommended roles for interacting with GKE clusters
> 
> IIUC, `gencred` binary needs to make the build cluster's K8s API server generate a new token for the "root" (cluster-admin KSA role) KSA on the build cluster. `gencred` does this by running as a GSA that has the `roles/container.admin` role defined by GCP. Is this correct?

correct https://github.com/kubernetes/test-infra/blob/408f66a2a9199fbcaafc55aa35f72a49a06da26c/gencred/pkg/serviceaccount/serviceaccount.go#L70

listx:
PTAL

site/content/en/docs/build-clusters.md
```diff
@@ -0,0 +1,181 @@
...
+
+*Note: This page discusses build clusters that use GKE. Technically speaking, a
+build cluster could be any Kubernetes cluster (not just GKE) because the only
+thing Prow needs is the ability to authenticate as a Kubernetes Service Account
+with cluster-admin role permissions to the build cluster.*
+
+## Overview
+
+By default Prow will schedule jobs in the same Kubernetes cluster as itself, but
+under the `test-pods` namespace. Setting up a separate build cluster will allow
```
chaodaiG, line 16: This is not necessarily true. For example "default" cluster in k8s prow is the build cluster located in `k8s-prow-builds` project instead of `k8s-prow` project. Consider something like:

> By default Prow will schedule jobs in the cluster that maps to kubeconfig alias "default" (Imagine running `kubectl config set-context "default" --cluster=<CLUSTER_CONTEXT>`, the <CLUSTER_CONTEXT> could be any cluster)

chaodaiG: Speaking of this, I think this document should also include infos related to context aliases. For example a GKE cluster named `cluster-foo` in project `project-bar` under `us-central1` could be named something totally unrelated, such as `the-most-awesome-build-cluster`. And whenever a user wants to configure something for this cluster they should use `the-most-awesome-build-cluster` instead of `cluster-foo`

site/content/en/docs/build-clusters.md
```diff
@@ -0,0 +1,181 @@
...
+into Prow and mount it into the various Prow components that need it; this
+one-way sync is performed by the `kubernetes-external-secrets` component. The
+Prow components' configurations don't have to be updated though, because your
+build cluster's token is combined with other secrets into a composite file.
+
+### Let your jobs report their status to GCS
+
+Your jobs in your build cluster must have GCS access in order to upload critical
+job metadata, such as a `finished.json` file to indicate the status of your job
+(whether it passed or failed). The GCS bucket location depends on how the Prow
```
chaodaiG, line 87: the GCS bucket could be configured on several different levels in Prow, and is not the same for a single Prow instance, for example a user can configure bucket for a build cluster (https://github.com/GoogleCloudPlatform/oss-test-infra/blob/d46723826c93a03f1d6e30d603bd5d0b0f723212/prow/oss/config.yaml#L138). Rarely user might want to use a different bucket for a job, which is also allowed

site/content/en/docs/build-clusters.md
```diff
@@ -0,0 +1,181 @@
...
+by [Deck](../components/deck), in order to report job status (e.g., "passing" or
+"failing").
+
+We look at both requirements below.
+
+### Let Prow schedule your jobs into your build cluster
+
+Prow is a Kubernetes cluster. So is your build cluster. In order for Prow to
+schedule jobs (i.e., create Kubernetes pods) into your build cluster, it must be
+able to authenticate as a KSA defined in your build cluster that has a
```
chaodaiG, line 58: KSA is not a hard requirement, it's just an easier way to generate kubeconfig for authenticating with a cluster. Being able to manage `Pod` resource on the build cluster might be more accurate, for more detail see gencred code

site/content/en/docs/build-clusters.md
```diff
@@ -0,0 +1,181 @@
...
+[cluster-admin](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles)
+Kubernetes role. This way, Prow (with its various individual components) can
+freely create, update, and delete jobs (pods) in your build cluster as
```
chaodaiG, line 61: not `jobs`, `prowjob` CR doesn't exist in build clusters

site/content/en/docs/build-clusters.md
```diff
@@ -0,0 +1,181 @@
...
+KSA name, grants authentication as this very same KSA.
+
+The question now turns to how we can generate this token. This can be done
+manually, but the token cannot be valid for longer than 2 days for security
+reasons (a GKE restriction), and must be rotated regularly. Fortunately, there
+is a tool called `gencred` that automates the generation of the token. We just
+need to have it be run periodically --- and so we need to add a Prow job that
+regularly invokes `gencred`.
+
+Once the token is generated, we can store it in the GCP Secret Manager for your
```
chaodaiG, line 76: it's in the GCP project where prow is in, so the GSA doesn't need GCP secret manager ACL in build cluster project

chaodaiG (changes requested):

site/content/en/docs/build-clusters.md
listx: I don't want to make this doc any longer. Let me try rephrasing and/or deleting the sentence you mentioned.

chaodaiG: if you meant to not extending this PR I would agree. Otherwise imo the `alias` vs real `cluster_name` is where our users got tripped most, and it's worth documented

site/content/en/docs/build-clusters.md
listx: But what about https://github.com/kubernetes/test-infra/blob/408f66a2a9199fbcaafc55aa35f72a49a06da26c/gencred/pkg/serviceaccount/serviceaccount.go#L78-L79 ? Seems to contradict what you just said. Unless there's other `gencred` codepaths involved here.

I will simply include your words verbatim into the doc.

listx: I'm confused because this line seems to condtradict what you're saying: https://github.com/kubernetes/test-infra/blob/408f66a2a9199fbcaafc55aa35f72a49a06da26c/gencred/pkg/serviceaccount/serviceaccount.go#L78

But also, I'll include your comment in the doc directly, because it will be faster than me reading through gencred.

chaodaiG: I don't think there is a conflict here. The point I'd like to make, is that "it must be able to authenticate as a KSA" is not an accurate statement, what Prow really must have is "it must be able to `kubectl get pods`, `kubectl create pods`, `kubectl * pods` etc", and other than creating a KSA, there are multiple other ways this can be achieved. At least gencred supports two different ways https://github.com/kubernetes/test-infra/blob/408f66a2a9199fbcaafc55aa35f72a49a06da26c/gencred/cmd/gencred/main.go#L236

chaodaiG: > I'm confused because this line seems to condtradict what you're saying: https://github.com/kubernetes/test-infra/blob/408f66a2a9199fbcaafc55aa35f72a49a06da26c/gencred/pkg/serviceaccount/serviceaccount.go#L78
> 
> But also, I'll include your comment in the doc directly, because it will be faster than me reading through gencred.

Well please hold off a few minutes if this is what you plan to do, I'll need to read through gencred code to make sure that I didn't miss something especially the RBAC stuff

chaodaiG: > Well please hold off a few minutes if this is what you plan to do, I'll need to read through gencred code to make sure that I didn't miss something especially the RBAC stuff

So according to https://github.com/kubernetes/test-infra/blob/408f66a2a9199fbcaafc55aa35f72a49a06da26c/gencred/pkg/serviceaccount/serviceaccount.go#L36 that this KSA is a "cluster-admin" super user. I'm not entirely convinced that it's a requirement, feels like a security hole to me. @cjwagner , do you know what resource Prow needs to manage other than `Pod` resource on build clusters?

site/content/en/docs/build-clusters.md
listx: > could be named something totally unrelated

Where is this definition created? I want to add a link to an example.

site/content/en/docs/build-clusters.md
listx: done

site/content/en/docs/build-clusters.md
listx: done

site/content/en/docs/build-clusters.md
listx: done

site/content/en/docs/build-clusters.md
chaodaiG: > > could be named something totally unrelated
> 
> Where is this definition created? I want to add a link to an example.

See https://github.com/kubernetes/test-infra/blob/408f66a2a9199fbcaafc55aa35f72a49a06da26c/gencred/cmd/gencred/main.go#L117, this is the `alias`

listx:
Spoke offline with Chao and agreed to punt some of the additional details to a future PR if necessary. Can I get LGTM please?

chaodaiG (approved):

chaodaiG:
Thank you Linus!

site/content/en/docs/build-clusters.md
listx: Oh interesting. I was expecting a declarative YAML config somewhere, but it is defined in gencred first. So I: guess it's accurate to say that context aliases start out from gencred (makes sense to me, because kubeconfig seems to be the source of truth WRT how prow is able to talk to other build clusters, and as I now know, the token(s) in the kubeconfig need to be refreshed by gencred).

PR #22, opened 2022-10-26
Update legacy snapshot

listx:
Broken links have been fixed. This update includes some other fixes in general (see each commit for more information).


listx:
Note that the broken links were fixed by pointing to the old k/t-i repo. This was done mechanically to get us to a "good" state. We still have to update these docs and reorganize them out of the Legacy Snapshot.

Issue #34, opened 2022-12-15
Add a search bar to the docs site.


cjwagner:
Is there an easy way to enable a search bar for the documentation site to facilitate navigation?
cc @listx

listx:
Looks like there are many options: https://gohugo.io/tools/search/

jihoon-seo:
Some live examples (FYI: these websites are also using Docsy):
- https://kubernetes.io/docs/home/
- https://glossary.cncf.io/
- ...

listx:
If I run hugo locally I can just edit this line https://github.com/kubernetes-sigs/prow/blob/1005b0be14d394fcb2dca13462951433a79f042a/site/config.toml#L136 to `true` and the sidebar now has a search box. I think this might be sufficient for now. The other option is to create the Google-backed search which is probably better, but then I think we need someone in the community to do the creation of this. @mrbobbytables is [this](https://www.docsy.dev/docs/adding-content/navigation/#setting-up-site-search) something you've done before?

cjwagner:
@listx Fixed this in https://github.com/kubernetes-sigs/prow/pull/38
/close

PR #38, opened 2022-12-21
add search in sidebar

listx:
Later we can upgrade to Google-indexed search, as outlined here: https://www.docsy.dev/docs/adding-content/navigation/#setting-up-site-search

Temporarily addresses #34

/cc @jihoon-seo @cjwagner

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -133,7 +133,7 @@ github_branch= "main"
 algolia_docsearch = false
 
 # Enable Lunr.js offline search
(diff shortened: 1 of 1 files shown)


jihoon-seo:
LGTM from my side.
(Personally this offline search looks better than Google-backed search. 🙂)

cjwagner (approved):
Thanks Linus. I think this makes the doc site way more useful.

PR #42, opened 2022-12-29
Finish incomplete lenses description

dobesv:
Wasn't sure how this sentence was meant to end so I just put something in here that seems reasonable


petr-muller (approved):
Thanks!

PR #43, opened 2022-12-29
Try to clarify the postsubmit / presubmit distinction a bit

dobesv:
I was a bit confused by this, so hopefully this helps others.  Assuming my interpretation is correct, that is!

### DIFF
--- a/site/content/en/docs/jobs.md
+++ b/site/content/en/docs/jobs.md
@@ -69,7 +69,7 @@ postsubmits:
 
 Postsubmits are run when a push event happens on a repo, hence they are
 configured per-repo. If no `branches` are specified, then they will run against
-every branch, including for open pull requests.
+every branch.
 
 Presubmit config looks like so (see [GoDocs](https://pkg.go.dev/k8s.io/test-infra/prow/config#Presubmit) for complete config):
 
--- a/site/content/en/docs/jobs.md
+++ b/site/content/en/docs/jobs.md
@@ -72,7 +72,7 @@ configured per-repo. If no `branches` are specified, then they will run on every
 every branch on the given repo.
 
 Postsubmit jobs apply `run_if_changed` and `skip_if_only_changed` filters based on which
-files were modified by the commits included in the specific push even from github.
+files were modified by the commits included in the specific push event from github.
 
 Presubmit config looks like so (see [GoDocs](https://pkg.go.dev/k8s.io/test-infra/prow/config#Presubmit) for complete config):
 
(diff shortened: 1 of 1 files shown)


site/content/en/docs/jobs.md
```diff
@@ -69,7 +69,7 @@ postsubmits:
 
 Postsubmits are run when a push event happens on a repo, hence they are
 configured per-repo. If no `branches` are specified, then they will run against
-every branch.
+every branch, including for open pull requests.
```
jihoon-seo, line 72: AFAIK postsubmits will not run for pull requests which are "open".
Instead, when a PR gets merged into a certain branch, it then means that new commits are pushed to that branch, so postsubmits will run.
Please let me know if I am wrong.

dobesv: I am seeing `postsubmits` running on every commit, regardless of whether they are in an open PR or not.  In fact, as soon as I push a commit the `postsubmits` will run, even before the PR is open.  Then when I open the PR it would run the `presubmits`.

I think to prevent `postsubmits` from running on PRs you would set branch filters on them, e.g. `branches: ['master']`

That said, I'm a prow newbie so maybe I'm missing something!

dobesv: Maybe it's just that normally people are setting a branch filter on `postsubmits` so they never notice it running on PR commits or commits to feature branches without a PR.

jihoon-seo: Hmm, let's suppose a situation that a contributor (he/him) wants to contribute to this repo (`kubernetes-sigs/prow`).

- If the feature branch that he created (say `my-feat-branch`) is on this repo (`kubernetes-sigs/prow`), and no `branches` are specified for postsubmits, then you are right, even before he opens a PR, postsubmits will run, because he pushed the new feature branch (`my-feat-branch`).
- But IMO in most cases contributors will create or push a new feature branch to their forked repo (e.g., `jihoon-seo/prow`), not to the upstream repo, because
  - most ordinary contributors do not have a permission to create or push a new feature branch to the upstream repo,
  - and even if for whom that have the permission, IMO it would be better to create or push a new feature branch to their forked repo, not to the upstream repo, to open a PR.

And it's highly likely that the dedicated Prow instance is watching and acts on the upstream repo, but not for forked repos.
This is why I said like [this](https://github.com/kubernetes-sigs/prow/pull/43#discussion_r1058732585).

dobesv, line 72:
```suggestion
every branch.
```

dobesv: Ah right, I think I get the confusion now.  My wording implied it would build commits in any pull request, but those commits would only be built if the pull request is to and from the same repo.  I've tried to clean it up a bit, let me know if that helps.

PR #45, opened 2023-01-11
Document Crier's webhook reporter

pierreprinetti:
Based on https://github.com/kubernetes/test-infra/pull/28341

### DIFF
--- a/site/content/en/docs/components/core/crier.md
+++ b/site/content/en/docs/components/core/crier.md
@@ -198,6 +198,158 @@ postsubmits:
               - echo
 ```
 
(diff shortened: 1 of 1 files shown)


pierreprinetti:
/hold
until the logic being documented here is merged (https://github.com/kubernetes/test-infra/pull/28341)

smg247:
I wonder if you can consolidate some of the documentation that is the same as the slack reporter. It might be nice to have a section that discusses the similar configuration of both reporters, and then another for each that talks about the unique parts. Just something to think about, I am fine with how it is too.

PR #52, opened 2023-04-14
Add example value for gcs_browser_prefix for AWS S3

dobesv:
I didn't realize it was possible to setup `gcs_browser_prefix` for S3, but I found a way to do it and thought it would be nice to share that with others in the documentation.


cjwagner (approved):
Thank you, Dobes.

PR #54, opened 2023-05-18
Update announcements.md for blobless cloning

airbornepony:
Announce the work done for kubernetes/test-infra/issues/26590.


BenTheElder (commented):
unrelated to this PR: this approach to breaking changes announcements seems awkward, I feel like we should be linking to the PR or something else more directly identifying than dates.

this specific change looks good

PR #55, opened 2023-05-23
Revert "Update announcements.md for blobless cloning"

airbornepony:
Reverts kubernetes-sigs/prow#54

The blobless fetches were reverted for now due to unforeseen merge issues.


mpherman2 (approved):

petr-muller (approved):
Maybe it would be better to add a line about _the blobless fetch support had to be removed_ rather than removing a line, but that's just a nit

PR #59, opened 2023-06-30
Add gardener prow instance to list

timebertt:
Add gardener prow instance to the list of prow instances in the wild :) 

cc @oliver-goetz


listx:
Thanks!

/lgtm
/approve

PR #70, opened 2023-11-16
Add go.mod file

timwangmusic:
We decided to use module name `sigs.k8s.io/prow`. The dependencies in the file are copied from k8s.io/test-infra and run `go mod tidy` to clean up the dependencies not used.

/assign @cjwagner @listx @airbornepony


matthyx (approved):

cjwagner:
This `go.mod` file doesn't appear to be valid for this repo. Did you confirm that this file actually works and the code builds?
Notably, we still have many go files that import from `k8s.io/test-infra` (which needs to be replaced with `sigs.k8s.io/prow`), but `k8s.io/test-infra` is not present in the go.mod.

timwangmusic:
> This `go.mod` file doesn't appear to be valid for this repo. Did you confirm that this file actually works and the code builds? Notably, we still have many go files that import from `k8s.io/test-infra` (which needs to be replaced with `sigs.k8s.io/prow`), but `k8s.io/test-infra` is not present in the go.mod.

Yes, in a follow-up PR I will fix the occurrences of  k8s.io/test-infra and start fixing the test cases.

cjwagner:
Seems like this PR was unnecessary then? Merging this invalid go.mod file doesn't help us does it? We'll still need to `go mod tidy` (or simply `go mod init` without copying the mod file from k/t-i) after the changes I mentioned are made.

PR #67, opened 2023-10-26
Add Prow as a subdirectory

timwangmusic:
The Prow directory contains all the current source code from k8s/test-infra with commit history.

Steps to recreate the change:
* Install the [git-filter-repo](https://github.com/newren/git-filter-repo) tool to your machine
* Execute the following shell script
```  
git clone https://github.com/kubernetes/test-infra.git
cd test-infra
git-filter-repo --path prow/
# $YOUR_PROW_FORK is your forked repo for https://github.com/kubernetes-sigs/prow
git remote add target $YOUR_PROW_FORK
git checkout main
# You are now on the target/main branch and then merge the local master branch
git merge master --allow-unrelated-histories
git push target main
```

/assign @cjwagner @listx @airbornepony

PR #71, opened 2023-12-04
Use new Go module name in Prow source files

timwangmusic:
The new Go module name is sigs.k8s.io/prow. In this change:
- Replace module names from k8s.io/test-infra to sigs.k8s.io/prow for Prow source files
- Added the tools package (copied from the k8s.io/test-infra repo)

/assign: @cjwagner @listx @airbornepony


cjwagner:
Just as in https://github.com/kubernetes-sigs/prow/pull/67 we'll need to be able to redo this to sync the latest Prow changes before we make this repo the source of truth. So the same feedback of making this repeatable via a shell script applies here. Ideally when we're ready to sync the latest Prow changes and make the switch, we can run one script to redo all the setup.

listx:
Taking a step back from this PR, I'm interested in how to do the actual switch over to this repo as the source of truth. What are the exact steps we need to do in order to make the switch happen? Could you explain in the PR description what will need to happen next, etc?

timwangmusic:
closing due to out of date.

PR #73, opened 2024-01-10
Add codegen related tools and scripts

timwangmusic:
`hack/make-rules/update/codegen.sh` is responsible for installing and running code generations. We copy this file from k8s.io/test-infra repo and made modifications with the new go module name.

- copy `hack/make-rules/update/codegen.sh` file from k8s.io/test-infra
- replace all occurrences of `k8s.io/test-infra` with `sigs.k8s.io`
- add associated files such as `hack/third_party/gimme`
- add prow job crd file


listx:
How can we check that these code generation scripts work correctly? Or, are we OK with accepting the possibility that there may be additional polishing needed in follow-up PRs?

listx:
Also, do you need these scripts in order to start building Prow images from this repo? Or are there still other dependencies we need to port over from the old repo?

/lgtm
/hold

timwangmusic:
> How can we check that these code generation scripts work correctly? Or, are we OK with accepting the possibility that there may be additional polishing needed in follow-up PRs?

Added a codegen.sh under `hack/make-rules/verify/`. This internally runs the other codegen.sh and verifies the output.

timwangmusic:
> Also, do you need these scripts in order to start building Prow images from this repo? Or are there still other dependencies we need to port over from the old repo?
> 
> /lgtm /hold

No, these scripts are purely for making sure we can run `hack/make-rules/update/codegen.sh`, I'll verify building the images later.

listx:
>  I'll verify building the images later.

SG. Is that coming after this PR, or is there some other work needed before you can start working on that?

/lgtm

timwangmusic:
> > I'll verify building the images later.
> 
> SG. Is that coming after this PR, or is there some other work needed before you can start working on that?
> 
> /lgtm

Yes, the next PR is going to replace the module paths with these generators and verify it is working by trying to build Prow images.

PR #75, opened 2024-01-19
Replace go module name

timwangmusic:
In this change, we replace all the occurrences of `k8s.io/test-infra` in the Go source files with the new module name prefix `sigs.k8s.io` with command `find . -type f -name "*.go" -exec sed -i 's/k8s.io\/test-infra/sigs.k8s.io/g' {} \;`

The following imports need to be restored after the replacement command:
- "k8s.io/test-infra/greenhouse/diskutil"
- "k8s.io/test-infra/pkg/flagutil"
- "k8s.io/test-infra/maintenance/migratestatus/migrator"
- "k8s.io/test-infra/experiment/clustersecretbackup/secretmanager"

/assign: @cjwagner @listx @airbornepony


prow/client/clientset/versioned/clientset.go
airbornepony, line 1: How were generated files updated? Seems important that these are produced using "make update-codegen" rather than sed edits.

config/prow/config.yaml
airbornepony, line 1: Can this commit be elided rather than reverted by the final commit in this change? Or is it needed for some reason?

prow/go.mod
airbornepony, line 1: This change looks distinct from the module renames. Should it be in a separate PR?

airbornepony (changes requested):
It's not clear why there are many commits in this PR if there's a single set of commands being run to produce the change across all files. (Sans the other changes which I didn't see in the description.)

Seems cleaner to make a change of this type with a single commit and a complete list of all commands executed, including the commands used to undo those specific unwanted edits.

That's essential for such wide-ranging changes as this; we humans aren't capable of reading such large changes and determining they are correct. Much easier to review the exact commands used to make the change. Thanks!

prow/client/clientset/versioned/clientset.go
timwangmusic: These are actually generated with `update-codegen.sh` script

timwangmusic:
> It's not clear why there are many commits in this PR if there's a single set of commands being run to produce the change across all files. (Sans the other changes which I didn't see in the description.)
> 
> Seems cleaner to make a change of this type with a single commit and a complete list of all commands executed, including the commands used to undo those specific unwanted edits.
> 
> That's essential for such wide-ranging changes as this; we humans aren't capable of reading such large changes and determining they are correct. Much easier to review the exact commands used to make the change. Thanks!

The commit history is skewed. I'll close this one and make a new PR for replacement of non-generated files only.

PR #76, opened 2024-01-23
Use new Go module name for generated files

timwangmusic:
These are generated with `bash hack/make-rules/update/codegen.sh` command.

/assign: @airbornepony

### DIFF
--- a/prow/config/prow-config-documented.yaml
+++ b/prow/config/prow-config-documented.yaml
@@ -359,11 +359,6 @@ deck:
         # each spyglass page. Using HTML in the template is acceptable.
         # Currently the only variable available is .ArtifactPath, which contains the GCS path for the job artifacts.
         announcement: ' '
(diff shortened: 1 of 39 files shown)


prow/plugins/plugin-config-documented.yaml
airbornepony, line 1: This is weird. I recall this change--it should be a move of this block to a different location in the generated file, not a strict delete.

Was the block deleted after https://github.com/kubernetes/test-infra/pull/31619/files was merged?

airbornepony: I just confirmed this block is still present. Not sure why it's being deleted.

prow/config/prow-config-documented.yaml
airbornepony, line 1: These deletions don't seem right.

timwangmusic:
See #82

PR #74, opened 2024-01-16
Sync Prow commits from the test-infra repo

timwangmusic:
Extract Prow commit history from test-infra up to 1/14/2024 and merge them to the Prow repo.

Script to reproduce this PR:
```
git clone https://github.com/kubernetes/test-infra.git
cd test-infra
git-filter-repo --path prow/
# $YOUR_PROW_FORK is your forked repo for https://github.com/kubernetes-sigs/prow
git remote add target $YOUR_PROW_FORK
git fetch target
# move to the target/main branch
git checkout target/main
git switch -c tmp
# merge the local master branch
git merge master --allow-unrelated-histories
git push --set-upstream target tmp
```

/assign: @listx @cjwagner @airbornepony

PR #78, opened 2024-01-25
Add script to sync Prow commits from test-infra repo

timwangmusic:
Add a script to sync Prow commits from test-infra repo to the current repo. The script is based on this [comment](https://github.com/kubernetes-sigs/prow/pull/74#issuecomment-1907347751) from @listx with modifications.

Running the script assumes installation of `git` and `git-filter-repo` tools.

### DIFF
--- a/hack/scripts/sync-prow.sh
+++ b/hack/scripts/sync-prow.sh
@@ -2,6 +2,15 @@
 
 set -euo pipefail
 
+check_deps() {
+  if ! which git-filter-repo > /dev/null 2>&1; then
+    echo "please install the git-filter-repo tool per https://github.com/newren/git-filter-repo"
+    exit 1
+  fi
+}
+
+check_deps
+
 old_prow="$PWD/test-infra"
 new_prow="$PWD/prow"
 
--- a/hack/scripts/sync-prow.sh
+++ b/hack/scripts/sync-prow.sh
@@ -19,17 +19,16 @@ if [[ ! -d "$old_prow" ]]; then
     git clone https://github.com/kubernetes/test-infra.git
 fi
 
-# Strip down the master branch of $old_prow to only have commits that touch the the /prow directory.
-pushd "$old_prow"
-git-filter-repo --path prow
-popd
-
 # Clone the new prow repo (this should be your fork).
 if [[ ! -d "$new_prow" ]]; then
     git clone https://github.com/kubernetes-sigs/prow.git
 fi
-cd "$new_prow"
 
+# Strip down the master branch of $old_prow to only have commits that touch the the /prow directory.
+cd "$old_prow"
+git-filter-repo --path prow
+
+cd "$new_prow"
 # Check out a new "sync" branch that we want to create a PR from.
 git branch sync origin/main
 git checkout sync
--- a/hack/scripts/sync-prow.sh
+++ b/hack/scripts/sync-prow.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+old_prow="$PWD/test-infra"
+new_prow="$PWD/prow"
+
+# Clone the old prow repo.
+if [[ ! -d "$old_prow" ]]; then
+    git clone https://github.com/kubernetes/test-infra.git
+fi
+
+# Strip down the master branch of $old_prow to only have commits that touch the the /prow directory.
+pushd "$old_prow"
+git-filter-repo --path prow
+popd
+
+# Clone the new prow repo (this should be your fork).
+if [[ ! -d "$new_prow" ]]; then
+    git clone https://github.com/kubernetes-sigs/prow.git
+fi
+cd "$new_prow"
+
+# Check out a new "sync" branch that we want to create a PR from.
+git branch sync origin/main
+git checkout sync
+
+# Import commits from old_prow/master into "sync" inside $new_prow.
+git remote add old_prow "$old_prow"
+git fetch old_prow
+git merge old_prow/master --no-edit


hack/scripts/sync-prow.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
listx, line 1: @timwangmusic FTR this script comes from https://github.com/kubernetes-sigs/prow/pull/74#issuecomment-1907347751. Typically when code is copied from somewhere, you should cite the source to include the additional context of how/why the script came to be. I don't expect you to cite the URL in the code directly, but having it in the PR description would have been helpful for reviewers (for example, to compare against the previous script that you already used to import commits into this repo, which had a different merge direction than this one).

hack/scripts/sync-prow.sh
```diff
@@ -0,0 +1,39 @@
...
+
+set -euo pipefail
+
+# Check if git-filter-repo tool is installed
+if which git-filter-repo > /dev/null 2>&1; then
+  echo "git-filter-repo is installed"
+else
+  echo "please install the git-filter-repo tool per https://github.com/newren/git-filter-repo"
+  exit 1
+fi
```
listx, line 11: This isn't really necessary because we already do `set -e` so if the tool isn't on the `$PATH` it will die.

You can use the `if ! command; then ... fi` idiom to only display the error if the command cannot be found or failed for whatever reason. I think it would be simpler to move this there instead.

airbornepony: Sorry, I requested this. I'd prefer to know about missing non-default tools before starting git ops.

Would be cleaner if encapsulated in a check_deps() shell function.

hack/scripts/sync-prow.sh
```diff
@@ -0,0 +1,39 @@
...
+new_prow="$PWD/prow"
+
+# Clone the old prow repo.
+if [[ ! -d "$old_prow" ]]; then
+    git clone https://github.com/kubernetes/test-infra.git
+fi
+
+# Strip down the master branch of $old_prow to only have commits that touch the the /prow directory.
+pushd "$old_prow"
+git-filter-repo --path prow
```
airbornepony, line 23: How do we expand this to include the hack/ and maybe other directories needed to build and test prow? It would be good to sync those at the same time to ensure consistency between code and tools.

hack/scripts/sync-prow.sh
timwangmusic: Added the reference in the PR description, thanks for pointing this out.

hack/scripts/sync-prow.sh
timwangmusic: Moved the checks to `check_deps` function.

hack/scripts/sync-prow.sh
timwangmusic: Currently the stuff in `/hack` do not tend to change frequently. We can address this when needed in the future.

airbornepony: If there are no changes, it's a NOP. The point is that when it does change, it's likely quite important to propagate at the same time. The sort of skew that could otherwise arise may be very difficult to troubleshoot.

hack/scripts/sync-prow.sh
```diff
@@ -0,0 +1,40 @@
...
+popd
```
airbornepony, line 25: It's not clear why pushd/popd are needed, and it sort of confused me into thinking $new_prow is a relative path to the original directory.

Can use plan "cd" above and omit popd since both $old_prow and $new_prow are absolute by virtue of $PWD.

hack/scripts/sync-prow.sh
airbornepony: We can follow up on this issue outside this PR.

hack/scripts/sync-prow.sh
timwangmusic: Done.

airbornepony:
/lgtm

Thanks, this looks pretty clean now!

hack/scripts/sync-prow.sh
listx: > Sorry, I requested this. I'd prefer to know about missing non-default tools before starting git ops.

Ah, OK. I would have gone for this option also if this script was going to be used long-term, but as this script is only temporary and there is only 1 niche tool involved I thought dedicating a separate stanza for this felt a bit like overkill. But I'm fine with check_deps() now that we have it.

hack/scripts/sync-prow.sh
listx: > It's not clear why pushd/popd are needed, and it sort of confused me into thinking $new_prow is a relative path to the original directory.

It's because when I originally wrote the script I didn't use `$old_prow` or other variables. Thanks for the suggestion.

hack/scripts/sync-prow.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+check_deps() {
+  if ! which git-filter-repo > /dev/null 2>&1; then
```
listx, line 6: FYI Bash comes with the `command` builtin command which is better than running `which` as that spawns a subprocess. So this could have been just `if ! command -v git-filter-repo; then`.

hack/scripts/sync-prow.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+check_deps
+
+old_prow="$PWD/test-infra"
+new_prow="$PWD/prow"
+
+# Clone the old prow repo.
+if [[ ! -d "$old_prow" ]]; then
+    git clone https://github.com/kubernetes/test-infra.git
```
listx, line 19: Another way to make this perhaps more obvious is to tell git to use a directory name directly, like this:

```
old_prow="$PWD/old_prow"
git clone https://github.com/kubernetes/test-infra.git $old_prow || true
```

That way it can just be a 1-liner, and `old_prow` will just be named `old_prow` to match the git remote name we give it on line 37.

listx:
@timwangmusic Made some minor suggestions for future improvements, should you decide to tweak it again. Thanks.

hack/scripts/sync-prow.sh
airbornepony: I had suggested using 'type' also builtin. I had not heard of command. Seems like with -v it's very similar to type?

listx: They are similar. I think `command -v` is basically the same as `type -P` which forces type to consider only those things in the PATH (to disregard any available shell aliases, for example).

I also thought of `type` sometime after writing my comment, but probably `command` is more obvious to folks not familiar with Bash, simply because "type" is such a generic word as to have no meaning without context. At least "command" gets us *some* context about what it does.

PR #82, opened 2024-01-30
Use new go module name

timwangmusic:
- The first commit (c9dbf0) is created via command `find . -type f -name "*.go" -exec sed -i 's/k8s.io\/test-infra/sigs.k8s.io\/prow/g' {} \;`
- The second commit(5e98a6) restores some imports to the original. We need to restore these imports from having `sigs.k8s.io/prow` prefix:
  - “k8s.io/test-infra/ghproxy/ghcache"
  - "k8s.io/test-infra/pkg/genyaml"
  - "k8s.io/test-infra/pkg/flagutil"
  - “"k8s.io/test-infra/greenhouse/diskutil"
  - "k8s.io/test-infra/robots/pr-creator/updater"
  - "k8s.io/test-infra/maintenance/migratestatus/migrator"
  - "k8s.io/test-infra/experiment/clustersecretbackup/secretmanager"

- The third commit updates codegen.sh file and added changes from running the script.


airbornepony:
How were the imports sorted in the second commit? Was there also a gofmt pass?

I'd suggest:
1. Document the commands used in each commit description so they make sense in git log.
2. Write and submit a script for these renames that is idempotent and can be rerun following future code moves.

timwangmusic:
> How were the imports sorted in the second commit? Was there also a gofmt pass?
> 
> I'd suggest:
> 
> 1. Document the commands used in each commit description so they make sense in git log.
> 2. Write and submit a script for these renames that is idempotent and can be rerun following future code moves.

Please see #84

PR #86, opened 2024-02-06
Copy packages from test-infra repo

timwangmusic:
I am trying to preserve the commit history for these packages Prow depends on. But it looks like some email addresses in commits maybe missing.

/cc @cjwagner

### DIFF
--- a/ghproxy/ghcache/README.md
+++ b/ghproxy/ghcache/README.md
@@ -1,0 +1,18 @@
+# ghCache
+
+## What?
+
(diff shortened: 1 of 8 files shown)


cjwagner:
This looks fine to me and I'll bypass the CLA requirement given that these commits already exist in k/t-i.

However, this appears to have been manually done. I do not see any script updates that do this sync. If there are changes to these libraries before we are ready to make the switch those changes need to be synced. As requested before, please don't do any manual steps, instead add everything to the sync script to make it easily and quickly reproducible.

PR #87, opened 2024-03-13
Add assorted files from kubernetes/test-infra that are needed to build and test.

cjwagner:
/assign @listx @airbornepony


hack/make-rules/update/codegen.sh
```diff
@@ -159,14 +159,14 @@ gen-deepcopy() {
   echo "Generating DeepCopy() methods..." >&2
   "$deepcopygen" \
     --go-header-file hack/boilerplate/boilerplate.generated.go.txt \
-    --input-dirs sigs.k8s.io/prow/apis/prowjobs/v1 \
+    --input-dirs k8s.io/test-infra/prow/apis/prowjobs/v1 \
     --output-file-base zz_generated.deepcopy \
-    --bounding-dirs sigs.k8s.io/prow/apis
+    --bounding-dirs k8s.io/test-infra/prow/apis
```
cjwagner, line 164: These path names were updated incorrectly so I just overwrote the whole file with the one from k/t-i. The module rename will be handled in bulk anyways by an improvement to the sync script (about to push a PR).

PR #88, opened 2024-03-14
Update sync-prow.sh script to sync all needed code and handle the module rename.

cjwagner:
I've tested this and it seems to be working properly now.

### DIFF
--- a/hack/scripts/sync-prow.sh
+++ b/hack/scripts/sync-prow.sh
@@ -3,30 +3,40 @@
 set -euo pipefail
 
 check_deps() {
-  if ! which git-filter-repo > /dev/null 2>&1; then
+  if ! command -v git-filter-repo &>/dev/null; then
     echo "please install the git-filter-repo tool per https://github.com/newren/git-filter-repo"
+    exit 1
+  fi
+
+  if [[ -z "${1}" ]]; then
+    echo "please specify your github username as a positional parameter."
+    echo "${1}"
     exit 1
   fi
 }
 
-check_deps
+# $1 must be your GitHub username.
+check_deps "${1:-}"
 
-old_prow="$PWD/test-infra"
+gh_username="${1}"
+old_prow="$PWD/old_prow"
 new_prow="$PWD/prow"
 
 # Clone the old prow repo.
-if [[ ! -d "$old_prow" ]]; then
-    git clone https://github.com/kubernetes/test-infra.git
-fi
+git clone https://github.com/kubernetes/test-infra.git $old_prow || true
 
 # Clone the new prow repo (this should be your fork).
-if [[ ! -d "$new_prow" ]]; then
-    git clone https://github.com/kubernetes-sigs/prow.git
-fi
+git clone "https://github.com/${gh_username}/prow.git" $new_prow || true
 
-# Strip down the master branch of $old_prow to only have commits that touch the the /prow directory.
+# Strip down the master branch of $old_prow to only have commits that touch the prow related directories.
 cd "$old_prow"
-git-filter-repo --path prow
+git-filter-repo \
+  --path prow \
+  --path ghproxy \
+  --path pkg/flagutil \
+  --path pkg/genyaml \
+  --path hack/prowimagebuilder \
+  --path hack/ts-rollup
 
 cd "$new_prow"
 # Check out a new "sync" branch that we want to create a PR from.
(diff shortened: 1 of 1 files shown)


hack/scripts/sync-prow.sh
```diff
@@ -36,4 +39,14 @@ git checkout sync
 # Import commits from old_prow/master into "sync" inside $new_prow.
 git remote add old_prow "$old_prow"
 git fetch old_prow
-git merge old_prow/master --no-edit
+# We expect this to fail with some merge conflicts due to documentation being moved to the site directory.
+git merge old_prow/master --no-edit --strategy-option theirs || true
+# Fix the merge conlflicts, these should be exclusively files that are considered [renamed/deleted] (i.e. moved).
```
michelle192837, line 44: nit: conflicts?

hack/scripts/sync-prow.sh
```diff
@@ -36,4 +39,14 @@ git checkout sync
...
+git add -u
+git commit -m "Sync Prow source from kubernetes/test-infra."
+
+# Rename go module paths
+find . -type f -exec sed -i 's/k8s.io\/test-infra/sigs.k8s.io\/prow/g' {} \;
```
listx, line 49:
```suggestion
find . -type f -exec sed -i 's,k8s.io/test-infra,sigs.k8s.io/prow,g' {} \;
```

hack/scripts/sync-prow.sh
```diff
@@ -3,30 +3,33 @@
...
-if [[ ! -d "$new_prow" ]]; then
-    git clone https://github.com/kubernetes-sigs/prow.git
-fi
+git clone "https://github.com/${gh_username}/prow.git" $new_prow || true
 
-# Strip down the master branch of $old_prow to only have commits that touch the the /prow directory.
+# Strip down the master branch of $old_prow to only have commits that touch the prow related directories.
 cd "$old_prow"
-git-filter-repo --path prow
+git-filter-repo --path prow --path ghproxy --path pkg/flagutil --path pkg/genyaml --path hack/prowimagebuilder --path hack/ts-rollup
```
listx, line 32:
```suggestion
git-filter-repo \
  --path ghproxy \
  --path hack/prowimagebuilder \
  --path hack/ts-rollup \
  --path pkg/flagutil \
  --path pkg/genyaml \
  --path prow
```

hack/scripts/sync-prow.sh
```diff
@@ -3,30 +3,33 @@
...
+
+  if [[ -z "${1}" ]]; then
+    echo "please specify your github username as a positional parameter."
+    echo "${1}"
+    exit 1
+  fi
 }
 
-check_deps
+check_deps "${1:-}"
```
listx, line 18:
```suggestion
# $1 must be your GitHub username.
check_deps "${1:-}"
```

hack/scripts/sync-prow.sh
cjwagner: Yeah that's more readable. I had just copied this from a prior PR's description. Updated.

cjwagner:
Comments addressed.

hack/scripts/sync-prow.sh
```diff
@@ -3,30 +3,40 @@
...
-old_prow="$PWD/test-infra"
+gh_username="${1}"
+old_prow="$PWD/old_prow"
 new_prow="$PWD/prow"
 
 # Clone the old prow repo.
-if [[ ! -d "$old_prow" ]]; then
-    git clone https://github.com/kubernetes/test-infra.git
-fi
+git clone https://github.com/kubernetes/test-infra.git $old_prow || true
```
airbornepony, line 26: It seems git clone could fail for a variety of reasons, some fatal (e.g. network error?) and the "true" alternation would mask such events. It seems the predicate approach is better since it only tries if not already present, which I presume is the error case being considered. (Same below.)

cjwagner: If the repo fails to clone subsequent steps will fail in a way that makes that clear. This sync script only needs to be run a few times so a network error is unlikely to occur. I don't really want to spend time implementing graceful error handling or fixing more typos.
I only changed this due to the requests that were made after the PR that added this script had already merged: https://github.com/kubernetes-sigs/prow/pull/78#pullrequestreview-1846560173

PR #89, opened 2024-03-14
Fix spelling check to actually return an error.

cjwagner:
/cc @airbornepony @michelle192837 @listx 

I'll make a PR to fix this upstream as well. I wonder how many misspellings we'll find upstream :upside_down_face:


cjwagner:
Upstream fix: https://github.com/kubernetes/test-infra/pull/32266

michelle192837 (approved):
/hold

Feel free to unhold at will!

PR #92, opened 2024-03-16
Add missing build dependencies from test-infra for deck image build

simony-gke:
The configs in top level folder and hack/ are hidden dependencies for gather-static.sh (in prow/cmd/deck, which is called by prowimagebuilder) and ts_rollup.

Tested prow/cmd/deck can be successfully built with these changes.

### DIFF
--- a/hack/build/ensure-node_modules.sh
+++ b/hack/build/ensure-node_modules.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# this script ensures node_modules is up to date with yarn in docker
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+
+cache() {
+    local action="$1"
+    cd "${REPO_ROOT}/hack/tools"
+    go build -o "${REPO_ROOT}/_bin/gcs-cacher" github.com/sethvargo/gcs-cacher
+
+    "${REPO_ROOT}/_bin/gcs-cacher" \
+        -bucket "k8s-prow-build-cache" \
+        -$action "yarn-deps-$(md5sum ${REPO_ROOT}/yarn.lock  | awk '{ print $1 }')" \
+        -dir "${REPO_ROOT}/node_modules" \
+        -allow-failure
+}
+
+# Download cache if running in CI
+if [[ "${CI:-}" == "true" ]]; then
+    cache restore
+fi
+
+cd "${REPO_ROOT}"
+hack/run-in-node-container.sh \
+    yarn install
+
+# Optional creates cache if this is supplied
+if [[ "${CREATE_TYPESCRIPT_CACHE:-}" == "true" ]]; then
+    cache cache
+fi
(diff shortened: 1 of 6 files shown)


hack/scripts/sync-prow.sh
```diff
@@ -33,6 +33,7 @@ cd "$old_prow"
 git-filter-repo \
   --path prow \
   --path ghproxy \
+  --path gopherage \
```
listx, line 36: Since this PR is bringing in more dependencies (some that are even at the top-level), I wonder if we need to modify this sync script to include these additional paths. We're going to have to do a force-push of master branch anyway because the initial import of just `prow` was not enough to get things working, so we might as well just add more paths here as needed --- and then when things are working well enough, do the sync all over again to clean up the history.

simony-gke: What additional paths are you referring to? I think Cole want to avoid the whole /hack path as he put only hack/prowimagebuilder here. The top-level configs can't be moved into a subfolder unless we want to change some of the ts-rollup code.

listx: > What additional paths are you referring to?

The files that are added in this PR. I was assuming that git-filter-repo could take file paths (not just directory paths).

Also, the more I think about this I think we should consider doing a "deny list" (negative) approach instead of allow-listing certain paths. E.g., import entire k/t-i except for the job config folders. This way we don't have to dance around which paths need inclusion, etc, while still removing probably 90% of the commits that have nothing to do with Prow.

cjwagner: I don't think we should be including parts of k/t-i that are not related to Prow. Including top level files and things from the hack directory that are needed for building is fine (and we can move these around within this repo later if desired), but we should not be including unrelated tools like gopherage.
If Deck has a build dependency on gopherage ([here?](https://github.com/kubernetes/test-infra/blob/546c34e1df034ee3683ee7584368e4ad3bbf4dff/prow/spyglass/lenses/coverage/coverage.ts#L19)), we should remove that dependency in k/t-i like I did for all the go dependencies. In particular, we can just copy and paste the needed gopherage code somewhere in the prow/ directory and then change the dependency to reference the prow/ internal copy.

>  E.g., import entire k/t-i except for the job config folders.

This would include a ton of non-prow test infrastructure which I would prefer not to include or move. It seems preferable to just fix this gopherage dependency issue in k/t-i before we do the next source sync.

listx: I realize I forgot to add the full picture of what I envisioned could happen in my last response. This was the big picture I was thinking of:

1.  Import most of k/t-i to quickly set up the Prow-related infra in the fork ("deny list" approach)
2. Migrate Prow build/test jobs to the fork.
3. Selectively delete non-Prow things (create a PR that deletes any remaining non-Prow things)
4. Run git-filter-repo when (2) is done to declutter history.
5. Force-push to the fork (technically the only time we need to do this force-push).

(1) would let us quickly get up to speed with the Prow-related build/test jobs that could work unmodified. And also this way we leave the force-pushing as the very last step (5).

OTOH, the approach you described will also work. I was mainly concerned about (1) and (2) taking a long time if we hit more snags about what to import, if we use the "allow list" (current) approach for importing k/t-i in (1). But I spoke with @simony-gke yesterday and I think my concern might have been premature.

cjwagner: That is an interesting idea, but I think we should continue with the allowlist approach instead for the following reasons:
1. Most importantly it looks like we've addressed most of the dependency problems already if we're just having problems creating the Deck image because of some typescript stuff. In other words, I think we're very close to having build and test jobs working via the allow list approach so I'd lean against switching now.
1. I'm not sure that it is true that "(1) would let us quickly get up to speed with the Prow-related build/test jobs that could work unmodified". While that would provide all Prow dependencies, it would also introduce a lot of non-prow code that may have additional dependencies that we'll need to provide before  `go test` and `go build` will work across the entire repo. We could modify the jobs and the relevant Make targets to exclude this, but that could be as much work.

If my reason turns out to be wrong and we find more problems then I think it would be worth reconsidering the deny list approach, but if this is the last piece we need to be building the Prow images successfully then it really seems like we're about to cross the finish line.

cjwagner: I made a PR to avoid the dependency in upstream: https://github.com/kubernetes/test-infra/pull/32284
If we're ok with continuing with the allowlist approach, we can just update this PR to not modify `sync-prow.sh` and then it will be ready to merge.

simony-gke: Sounds good to me. I've removed the change to sync-prow.sh.

cjwagner (approved):

PR #93, opened 2024-03-16
Remove non-prow images in .ko.yaml file.

simony-gke:
Remove robots/issue-creator and testgrid/cmd/configurator from .ko.yaml.


cjwagner (approved):
Oops I missed these because they weren't under the `# external` section.  Thanks.

PR #94, opened 2024-03-22
Fix sync-prow.sh to avoid duplicating commit history.

cjwagner:
/assign @listx @simony-gke 

I used this script to generate this branch: https://github.com/cjwagner/prow/tree/sync
The branch will not merge cleanly so we'll need to refresh the sync and force push this to the main branch when we're ready. I don't think subsequent syncs will require a force push, but I am not certain.

### DIFF
--- a/hack/scripts/sync-prow.sh
+++ b/hack/scripts/sync-prow.sh
@@ -22,35 +22,43 @@ gh_username="${1}"
 old_prow="$PWD/old_prow"
 new_prow="$PWD/prow"
 
+rm -rf $old_prow $new_prow
 # Clone the old prow repo.
 git clone https://github.com/kubernetes/test-infra.git $old_prow || true
 
 # Clone the new prow repo (this should be your fork).
 git clone "https://github.com/${gh_username}/prow.git" $new_prow || true
 
-# Strip down the master branch of $old_prow to only have commits that touch the prow related directories.
-cd "$old_prow"
-git-filter-repo \
+# These are the --path arguments that define the set of paths we want to migrate:
+path_args="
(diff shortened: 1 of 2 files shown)


hack/scripts/sync-prow.sh
```diff
@@ -22,35 +22,43 @@ gh_username="${1}"
...
 
 # Clone the new prow repo (this should be your fork).
 git clone "https://github.com/${gh_username}/prow.git" $new_prow || true
 
-# Strip down the master branch of $old_prow to only have commits that touch the prow related directories.
-cd "$old_prow"
-git-filter-repo \
+# These are the --path arguments that define the set of paths we want to migrate:
+path_args="
   --path prow \
```
listx, line 34: The gopherage paths that Simon wanted to add earlier are not here because we took care of that in k/t-i. Makes sense.

hack/scripts/sync-prow.sh
```diff
@@ -22,35 +22,43 @@ gh_username="${1}"
...
 cd "$new_prow"
 # Check out a new "sync" branch that we want to create a PR from.
 git branch sync origin/main
 git checkout sync
 
+# Remove any existing history for the directories we care about to avoid duplicate commits.
+git-filter-repo \
+  --invert-paths \
+  --force \
+  ${path_args}
```
listx, line 56: Ah, very nice discovery. I admire your courage in diving into the wall of text for the docs around git-filter-repo.

listx (commented):
/lgtm

Wonderful (I appreciate the splitting up of commits also). Thanks!

PR #99, opened 2024-04-09
Update README to reflect the source code migration.

cjwagner:
This should trigger the image publishing postsubmit added in https://github.com/kubernetes/test-infra/pull/32417

If this works we can start removing Prow related things from kubernetes/test-infra


cjwagner:
This is part of https://github.com/kubernetes/test-infra/issues/31728

PR #101, opened 2024-04-11
move pkg, cmd & test folders in inside prow to the root of the repo

upodroid:


upodroid:
We need to merge the linked PR in test-infra to run the failing integration/image-build job.

danilo-gemoli:
If we want to redefine the folders structure I think we should take `ghproxy` into account as well. With the previous structure it was clear that Prow and GHProxy were separate as they were at the same level:
```
prow/
ghproxy/
```
Now we are going to have such a structure:
```
ghproxy/
cmd/
pkg/
...
```
that is a bit incosistent IMHO.
I'm wondering whether we should consider moving `ghproxy` into the `cmd/` folder.

cjwagner:
I'm assuming this is following the guidelines from this page: https://go.dev/doc/modules/layout, specifically this section seems to be most applicable to the contents of our repo: https://go.dev/doc/modules/layout#packages-and-commands-in-the-same-repository

We'll also need to create an announcement and include a migration guide since this is a backwards incompatible change for consumers that depend on these go packages. Perhaps you can provide a `sed` command that updates import paths according to where/how they are moved by this PR?

upodroid:
> We'll also need to create an announcement and include a migration guide since this is a backwards incompatible change for consumers that depend on these go packages.

I can send another email to the existing announcement about the updated import paths. Prow lib consumers will hopefully run into this problem +/-7 days from this comment.

> I'm assuming this is following the guidelines from this page: https://go.dev/doc/modules/layout, specifically this section seems to be most applicable to the contents of our repo: https://go.dev/doc/modules/layout#packages-and-commands-in-the-same-repository

That is right, all the programs will live in cmd/* and libraries will live in pkg/*, we don't have internal libraries yet but we should start doing that soon.

I'll push a separate commit moving ghproxy to the cmd folder

Can you approve the PR in test-infra to fix the tests?

cjwagner:
> I'll push a separate commit moving ghproxy to the cmd folder
>
> Can you approve the PR in test-infra to fix the tests?

Yep, I'll approve the test-infra PR. Please don't merge it until this PR has been updated to move ghproxy as well and is also ready to merge. These PRs need to merge together so if there is a problem with either we'll need to quickly revert.

upodroid:
@cjwagner Image builds are hanging for some reasons, any ideas about what is happening?

upodroid:
This is ready to be merged.

.gitignore
```diff
@@ -1,6 +1,8 @@
 _bin/
 _output/
 .python_virtual_env
+/yarn.lock
```
jihoon-seo, line 4: I suggest to remove this line, to version control the `yarn.lock` file;
please see https://github.com/yarnpkg/yarn/issues/1583

.gitignore
```diff
@@ -1,6 +1,8 @@
...
+/node_modules
```
jihoon-seo, line 5: Since `node_modules` is a directory, I suggest to modify this line slightly as follows:
→ `node_modules/`

jihoon-seo (commented):
@upodroid PTAL!

jihoon-seo:
(If someone feels that squash merge is not appropriate for this PR, then please feel free to remove the label.)

.gitignore
upodroid: When we are building images or running deck locally, we use yarn to generate the node_modules and yarn.lock and run `hack/ts-rollup` to create the final js files that go in the container. When the build fails, we get a dirty git tree but these folders should never be committed.

.gitignore
upodroid: ^

.gitignore
jihoon-seo: I am not objecting to, actually am agreeing to, add 'node_modules' directory to gitignore.
What I am suggesting is that we SHOULD version control the 'yarn.lock' file, and it's just a single file, not a directory.

cmd/deck/runlocal
```diff
@@ -77,6 +77,6 @@ fi
     --static-files-location="$DIR/kodata/static" \
     --template-files-location="$DIR/kodata/template" \
     --spyglass-files-location="$DIR/kodata/lenses" \
-    --config-path "${DIR}/../../../config/prow/config.yaml" \
+    --config-path "${DIR}/../../test/integration/config/prow/deck.yaml" \
```
upodroid, line 80: When deck is being developed locally, it assumes the prow config at https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml is in the repo which is no longer true. I copied the prod config and stored it next to the integration config.yaml. I can confirm that running `./cmd/deck/runlocal` runs deck locally on my laptop and it works nicely.

.gitignore
```diff
@@ -1,2 +1,39 @@
...
+cmd/sub/sub
+cmd/tackle/tackle
+cmd/tide/tide
+cmd/tot/fallbackcheck/fallbackcheck
+cmd/tot/tot
+cmd/external-plugins/cherrypicker/cherrypicker
+cmd/external-plugins/needs-rebase/needs-rebase
+cmd/external-plugins/refresh/refresh
+cmd/ghproxy/ghproxy
+pj.yaml
```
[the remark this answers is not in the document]
petr-muller: Patricia has that change in https://github.com/kubernetes-sigs/prow/pull/106, if we add it here she'll need to respin. Let's just merge this and then the dep bump. :+1:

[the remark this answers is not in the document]
upodroid: Yeah, also not sure why we need to ignore vendor. I'll take a further look at 106

petr-muller (approved):

petr-muller:
CI changes are in place, this passes tests => :shipit:

.gitignore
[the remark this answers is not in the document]
petr-muller: We don't vendor in Prow so having it in gitignore helps people realize that before submitting dep bump PRs

PR #110, opened 2024-04-16
Fix flake unit test TestAddWithParser

danilo-gemoli:
This PR is an attempt to fix https://github.com/kubernetes/test-infra/issues/31313.
As far as I was able to understand the test pulled out the value from the generator ([here](https://github.com/kubernetes/test-infra/blob/c76de018691234a5bfa8bb2c0e5586149c32db0a/prow/config/secret/agent_test.go#L159-L161)) before the agent had the time to update it ([here](https://github.com/kubernetes/test-infra/blob/c76de018691234a5bfa8bb2c0e5586149c32db0a/prow/config/secret/reloader.go#L78-L81)).
This solution might not be ideal still, due to the 10s time.Sleep but we need to go a pretty extensive refactor to make the test 100% deterministic.
Anyway, as far as I know the sleep has never been a problem so far, therefore we might be good this way.

The original PR is https://github.com/kubernetes/test-infra/pull/32385

### DIFF
--- a/pkg/config/secret/agent_test.go
+++ b/pkg/config/secret/agent_test.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -122,17 +123,26 @@ func testAddWithParser(t *testing.T) {
 		t.Fatalf("failed to write initial content of secret: %v", err)
 	}
 
-	vals := make(chan int, 3)
-	errs := make(chan error, 3)
+	newValAvailable := make(chan struct{})
+	// The first time the secret gets parsed synchronously, hence we don't want
+	// to block because nobody will listen for any notification.
+	firstRead := atomic.Bool{}
+	errs := make(chan error)
 	generator, err := AddWithParser(
 		secretPath,
 		func(raw []byte) (int, error) {
 			val, err := strconv.Atoi(string(raw))
 			if err != nil {
-				errs <- err
+				// This compare n swap shouldn't be needed as we don't expect any error the first time.
+				// It's just for extra safety.
+				if !firstRead.CompareAndSwap(false, true) {
+					errs <- err
+				}
 				return val, err
 			}
-			vals <- val
+			if !firstRead.CompareAndSwap(false, true) {
+				newValAvailable <- struct{}{}
+			}
 			return val, err
 		},
 	)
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
I'm having a hard time understanding the point of this unit test - that `AddWithParser` correctly returns the values from the parsing func? Why is this useful? I also think that moving from channels to mutexes is almost always the wrong path in Go, so I'd like to see a discussion on a) the root cause of the flake and b) whether we need to keep this test or exercise `AddWithParser` better. For instance - we know our parsing func is `strconv.Atoi` ... we write `"2"` and expect an integer. Why do we care about the internals?

danilo-gemoli:
Updated after a conversation with @stevekuznetsov

petr-muller:
/ok-to-test

I hope this comment also makes `needs-rebase` go away, that piece of GH had an incident earlier today.

pkg/config/secret/agent_test.go
```diff
@@ -142,22 +152,22 @@ func testAddWithParser(t *testing.T) {
...
-		select {
-		case v := <-vals:
-			if v != expected {
-				t.Errorf("expected value to get updated to %d but got updated to %d", expected, v)
-			}
-			break
-		case err := <-errs:
-			if e == nil || e.Error() != err.Error() {
-				t.Fatalf("expected error %v, got %v", e, err)
+	perpetualCheck:
```
stevekuznetsov, line 155: I'm not sure I understand the intent here - without the label and explicit `break`s wouldn't we have the same execution flows?

danilo-gemoli: I'm not sure, unless we replace `break`s with `return`s.
A "regular" `break` doesn't interrupt the outer loop:
```golang
for {
  select {
    case ...:
      // Stop `select`-ing only, skip to the next `for` iteration
      break
  }
}
```
therefore, with no labeled `break`, we have no way to exit cleanly from that function as it would:
- either time out and fail: `t.Fatalf("timed out waiting for value %d and error %d", expected, e)`.
- or fail in case of unexpected errors: `t.Fatalf("expected error %v, got %v", e, err)`

stevekuznetsov: You're in an anonymous function block, so prefer to use the `return` - labels and GOTO are not ideal.

stevekuznetsov: Also, if you `break` out of the `select`, the `for` has nothing else to do and you will go to another loop - so ... what is the difference there between breaking out of the `for` and re-running it (while only ever allowing it to have one iteration) and just running another iteration? That was my original comment.

droslean:
@danilo-gemoli Is this PR still stands?

danilo-gemoli:
@stevekuznetsov I don't remember anymore where we were at with this PR. Are you ok with it?

Issue #115, opened 2024-04-18
Mass job status change API


Prucek:
Original issue: https://github.com/kubernetes/test-infra/issues/30108

**What would you like to be added:**
A simple API to mass update statuses of jobs based on filtering criteria:

- cluster
- state
- started_after
- started_before
- job type

There would be a single endpoint that would accept the above criteria, and find and update all matching `ProwJobs` to the provided state.

**Why is this needed:**
OpenShift recently experienced an outage in one of the build clusters in our Prow instance. Due to this, we had a large number of jobs stuck in the `triggered` state. We could have made good use of something that allowed us to abort all of these jobs. I think a general purpose API to update the status of jobs would be useful for other purposes as well, but the most salient use case would be mass-abort by criteria.

Prucek:
Original comment: https://github.com/kubernetes/test-infra/issues/30108#issuecomment-1954984056

**Endpoint**
We should achieve this functionality by adding a new endpoint to the [Gangway API](https://docs.prow.k8s.io/docs/components/optional/gangway/). The endpoint, `BulkJobStatusChange`, will accept a `BulkJobStatusChangeRequest` containing:

- currentState
- desiredState
- cluster (optional)
- startedAfter (optional)
- startedBefore (optional)
- jobType (optional)
- org (optional)
- repo (optional)

It will use the provided parameters to search for any `ProwJob` CRs that match. All matching CRs will have their `status.state` field updated to match the `desiredState` provided. Note that we should validate each of these parameters to make sure the values make sense; especially the `desiredState`.

**Updating the State**
Each listed ProwJob will have its `status.state` field updated to the `desiredState` provided by utilizing prowjob client's `UpdateStatus` [function](https://github.com/kubernetes/test-infra/blob/2efbce017d1a6413667320cd8f979fd40b8ac7b5/prow/client/clientset/versioned/typed/prowjobs/v1/prowjob.go#L139-L153).

PR #116, opened 2024-04-18
`gangway`: add status change endpoint

Prucek:
Resolves: https://github.com/kubernetes-sigs/prow/issues/115

Add new `BulkJobStatusChange` endpoint to change the status of multiple PJs.


pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+
+	err, md := getHttpRequestHeaders(ctx)
+	if err != nil {
+		logrus.WithError(err).Debug("could not find request HTTP headers")
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	if err := request.Validate(); err != nil {
+		logrus.WithError(err).Debug("could not validate request fields")
+		return nil, err
```
smg247, line 181:
```suggestion
                return nil, status.Error(codes.InvalidArgument, err.Error())
```

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+	}
+
+	mainConfig := gw.ConfigAgent.Config()
+	allowedApiClient, err := mainConfig.IdentifyAllowedClient(md)
+	if err != nil {
+		logrus.WithError(err).Debug("could not find client in allowlist")
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	pjList, err := gw.ProwJobClient.List(ctx, metav1.ListOptions{})
```
smg247, line 191: We should put a TODO in here explaining why all prowjobs must be listed and how to fix it when it is possible

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+	jobExecutions := &JobExecutions{}
+	var count int32
+	for _, pj := range pjList.Items {
+		if !isMatchingCondition(pj, request) {
+			continue
+		}
+		if allowedApiClient != nil {
+			authorized := ClientAuthorized(allowedApiClient, pj)
+			if !authorized {
+				logrus.Error("client is not authorized to execute the given job")
```
smg247, line 206:
```suggestion
				logrus.Error("client is not authorized to modify the given job")
```

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+				return nil, status.Error(codes.PermissionDenied, "client is not authorized to execute the given job")
```
smg247, line 207:
```suggestion
				return nil, status.Error(codes.PermissionDenied, "client is not authorized to modify the given job")
```

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+			}
+		}
+
+		pj.Status.State = prowcrd.ProwJobState(strings.ToLower(request.GetJobStatusChange().GetDesired().String()))
+		updatedPj, err := gw.ProwJobClient.Update(ctx, &pj, metav1.UpdateOptions{})
+		if err != nil {
+			logrus.WithError(err).Debug("failed to update ProwJob status")
```
smg247, line 214: This seems like a bigger deal than a debug log. It at least needs to be a warning, but probably better off as an error level.

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+			continue
+		}
+		logrus.WithField("name", pj.Name).Info("ProwJob status updated to:", updatedPj.Status.State)
```
smg247, line 217:
```suggestion
		logrus.WithField("name", pj.Name).Infof("ProwJob status updated to: %s", updatedPj.Status.State)
```

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+
+		jobExec := &JobExecution{
+			Id:        updatedPj.Name,
+			JobName:   updatedPj.Spec.Job,
+			JobType:   JobExecutionType(TranslateProwJobType(updatedPj.Spec.Type)),
+			JobStatus: TranslateProwJobStatus(&updatedPj.Status),
+		}
+		jobExecutions.JobExecution = append(jobExecutions.JobExecution, jobExec)
+		count++
```
smg247, line 226: I don't think we need to keep track of the count like this, we can just do a `len(jobExecutions.JobExecution)` at the end, right?

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+	}
+
+	// If more than n jobs were affected, we only return the first n jobs
```
smg247, line 229: May as well use `10` rather than `n` in the comment

pkg/gangway/gangway.go
```diff
@@ -135,13 +146,143 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+	// because we don't want to return a huge list of jobs and wait for a long time
+	n := int32(10)
```
smg247, line 231: this should be defined as a const

pkg/gangway/gangway.go
```diff
@@ -349,6 +490,17 @@ func (cjer *CreateJobExecutionRequest) Validate() error {
 	return nil
 }
 
+func (bjscr *BulkJobStatusChangeRequest) Validate() error {
```
smg247, line 493: Do we need to validate that the startedBefore and startedAfter are parseable as times if they are provided?

petr-muller:
:+1: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_prow/155/pull-prow-integration/1791473589053034496#1:build-log.txt%3A887:

```
 === FAIL: test/integration/test TestGangwayBulkJobStatusChange (12.74s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x1b69d80] 
```

Let's revert

PR #117, opened 2024-04-18
Revive prowjob when node is terminated (enabled by default)

inteon:
Recreated pods when we detect they failed due to its node shutting down.
This saves us from having to rerun jobs that were terminated due to a spot instance shutdown.

## Current behavior

Now, when a node is terminated, the prowjob fails (PJ gets in the FailureState).

## New behavior

The prowjob is now deleted and recreated (revived) when the node is terminated.

If the pod is revived more than what is allowed through the MaxRevivals config value, the prowjob errors (PJ gets in the ErrorState).

Additionally, the ErrorOnTermination option can be used to error a prowjob directly when the pod is terminated (PJ gets in the ErrorState).

## Detecting node termination

The pod status that we see on pods terminated due to a spot instance shutdown look like this:
```yaml
status:
  message: Pod was terminated in response to imminent node shutdown.
  phase: Failed
  reason: Terminated
```
or
```yaml
status:
  conditions:
  - message: 'PodGC: node no longer exists'
    reason: DeletionByPodGC
    status: "True"
    type: DisruptionTarget
  phase: Failed
```
or
```yaml
status:
  conditions:
  - message: 'GCPControllerManager: node no longer exists'
    reason: DeletionByGCPControllerManager
    status: "True"
    type: DisruptionTarget
  phase: Failed
```

The TerminationConditionReasons option allows users to modify what pod condition reason values are used to detect that the node was being terminated (defaults to "DeletionByPodGC" and "DeletionByGCPControllerManager").

### DIFF
--- a/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml
+++ b/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml
@@ -1614,6 +1614,13 @@ spec:
                   If this field is unspecified or false, a new pod will be created to replace
                   the evicted one.
                 type: boolean
+              error_on_termination:
+                description: |-
+                  ErrorOnTermination indicates that the ProwJob should be completed and given
+                  the ErrorState status if the pod that is executing the job is terminated.
+                  If this field is unspecified or false, a new pod will be created to replace
+                  the terminated one.
+                type: boolean
               extra_refs:
                 description: |-
                   ExtraRefs are auxiliary repositories that
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -673,6 +673,13 @@ type Plank struct {
 	// Defaults to 3. A value of 0 means no retries.
 	MaxRevivals *int `json:"max_revivals,omitempty"`
 
+	// TerminationConditionReasons is a set of pod status condition reasons on which the
+	// controller will match to determine if the pod's node is being terminated. If a node is being
+	// terminated the controller will restart the prowjob, unless the ErrorOnTermination option
+	// is set on the prowjob or the MaxRevivals option is reached.
+	// Defaults to ["DeletionByPodGC", "DeletionByGCPControllerManager"].
+	TerminationConditionReasons []string `json:"termination_condition_reasons,omitempty"`
+
 	// DefaultDecorationConfigs holds the default decoration config for specific values.
 	//
 	// Each entry in the slice specifies Repo and Cluster regexp filter fields to
@@ -2526,6 +2533,19 @@ func parseProwConfig(c *Config) error {
 		c.Plank.MaxRevivals = &maxRetries
 	}
 
+	if c.Plank.TerminationConditionReasons == nil {
+		c.Plank.TerminationConditionReasons = []string{
+			// If the node does no longer exist and the pod gets garbage collected,
+			// this condition will be set:
+			// https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-conditions
+			"DeletionByPodGC",
+			// On GCP, before a new spot instance is started, the old pods are garbage
+			// collected (if they have not been already by the Kubernetes PodGC):
+			// https://github.com/kubernetes/cloud-provider-gcp/blob/25e5dcc715781316bc5e39f8b17c0d5b313453f7/cmd/gcp-controller-manager/node_csr_approver.go#L1035-L1058
+			"DeletionByGCPControllerManager",
+		}
+	}
+
 	if err := c.Gerrit.DefaultAndValidate(); err != nil {
 		return fmt.Errorf("validating gerrit config: %w", err)
 	}
@@ -2923,6 +2943,8 @@ func validateAgent(v JobBase, podNamespace string) error {
 		return fmt.Errorf("decoration requires agent: %s (found %q)", k, agent)
 	case v.ErrorOnEviction && agent != k:
 		return fmt.Errorf("error_on_eviction only applies to agent: %s (found %q)", k, agent)
+	case v.ErrorOnTermination && agent != k:
+		return fmt.Errorf("error_on_termination only applies to agent: %s (found %q)", k, agent)
 	case v.Namespace == nil || *v.Namespace == "":
 		return fmt.Errorf("failed to default namespace")
 	case *v.Namespace != podNamespace && agent != p:
--- a/pkg/config/prow-config-documented.yaml
+++ b/pkg/config/prow-config-documented.yaml
@@ -1345,6 +1345,13 @@ plank:
     # Use `org/repo`, `org` or `*` as a key.
     report_templates:
         "": ""
+    # TerminationConditionReasons is a set of pod status condition reasons on which the
+    # controller will match to determine if the pod's node is being terminated. If a node is being
+    # terminated the controller will restart the prowjob, unless the ErrorOnTermination option
+    # is set on the prowjob or the MaxRevivals option is reached.
+    # Defaults to ["DeletionByPodGC", "DeletionByGCPControllerManager"].
+    termination_condition_reasons:
+        - ""
 # PodNamespace is the namespace in the cluster that prow
 # components will use for looking up Pods owned by ProwJobs.
 # The namespace needs to exist and will not be created by prow.
--- a/pkg/plank/reconciler.go
+++ b/pkg/plank/reconciler.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -475,7 +476,7 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
 			pj.Status.PodName = pn
 			r.log.WithFields(pjutil.ProwJobFields(pj)).Info("Pod is missing, starting a new pod")
 		}
-	} else if podUnexpectedStopCause := getPodUnexpectedStopCause(pod); podUnexpectedStopCause != PodUnexpectedStopCauseNone {
+	} else if podUnexpectedStopCause := getPodUnexpectedStopCause(pod, r.config().Plank.TerminationConditionReasons); podUnexpectedStopCause != PodUnexpectedStopCauseNone {
 		switch {
 		case podUnexpectedStopCause == PodUnexpectedStopCauseEvicted && pj.Spec.ErrorOnEviction:
 			// ErrorOnEviction is enabled, complete the PJ and mark it as errored.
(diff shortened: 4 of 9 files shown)


jihoon-seo:
Hello @inteon
AFAIK Plank was deprecated.
Since this PR is adding codes under 'pkg/plank/',
I wonder whether:
1. Adding codes under 'pkg/plank/' is okay regardless of the deprecation of Plank; or
2. There exists another method to achieve the same goal, adding codes to directories other than 'pkg/plank/'.

Please take a look! 😃

inteon:
> Hello @inteon AFAIK Plank was deprecated. Since this PR is adding codes under 'pkg/plank/', I wonder whether:
> 
> 1. Adding codes under 'pkg/plank/' is okay regardless of the deprecation of Plank; or
> 2. There exists another method to achieve the same goal, adding codes to directories other than 'pkg/plank/'.
> 
> Please take a look! 😃

The code still seems to be used in prow-controller-manager: https://github.com/kubernetes-sigs/prow/blob/0a3518154cf7d730ddc5ad7c6b978c480a6e5dc9/cmd/prow-controller-manager/main.go#L202-L204

inteon:
Info: we use this change on our cert-manager prow cluster https://prow.infra.cert-manager.io/ (see https://github.com/cert-manager/testing/tree/0b5dfa456f691e849bb0b3c40f3e00bd8d607127/images/prow-controller-manager-spot).
It works very well, this change has removed a lot of flaky failures due to spot instance shutdowns!

droslean:
> Info: we use this change on our cert-manager prow cluster https://prow.infra.cert-manager.io/ (see https://github.com/cert-manager/testing/tree/0b5dfa456f691e849bb0b3c40f3e00bd8d607127/images/prow-controller-manager-spot). It works very well, this change has removed a lot of flaky failures due to spot instance shutdowns!

This change seems like a nice feature but I am not sure if you are covering all of the generic cases of those pod terminations that are not only related to GCP. Also, this issue can be resolved by adding a pod disruption budget to your cluster. 

In my opinion, we shouldn't enable this feature by default. Instead, we can configure specific reasons for recreating the prowjobs, which will allow users to be more specific about their infrastructure.

inteon:
> I am not confident about enabling this feature by default. The implementation covers only a GCP case, and a pod's termination is deeply dependent on the infrastructure. This means, that a prowjob can run forever in a loop in many cases. Perhaps keeping it disabled as default and allowing the user to enable it in the global config? Also, to avoid the infinite runs, perhaps its better to keep track of the number of the retries and allow the user to control the threshold.
> 
> @petr-muller @stevekuznetsov @cgwalters @BenTheElder Do you guys have any input?

I updated the PR and added a RetryCount which is now incremented every time the pod is re-created (it also counts other retries that were already present in code). There will be a hard failure after 3 retries have been reached (we might want to make this a variable in the future).

inteon:
Blocked by https://github.com/kubernetes-sigs/prow/pull/196 since CRD is too large for my setup.

stevekuznetsov:
@droslean I don't think PDB would help here since you don't get a choice as to when spot instances get taken away from you.

droslean:
> @droslean I don't think PDB would help here since you don't get a choice as to when spot instances get taken away from you.

Yep. My only concern is whether we should allow this feature to be enabled by default.  Since Prow doesn't directly know what the job will do, there can be cases where the job costs will triple if we allow this by default. I would prefer to make this configuration and let the user decide based on the infrastructure.

pkg/plank/reconciler.go
```diff
@@ -672,6 +656,55 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
...
+		// this condition will be set:
+		// https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-conditions
+		if condition.Reason == "DeletionByPodGC" {
+			return PodTransientFailureTerminated
+		}
+
+		// On GCP, before a new spot instance is started, the old pods are garbage
+		// collected (if they have not been already by the Kubernetes PodGC):
+		// https://github.com/kubernetes/cloud-provider-gcp/blob/25e5dcc715781316bc5e39f8b17c0d5b313453f7/cmd/gcp-controller-manager/node_csr_approver.go#L1035-L1058
+		if condition.Reason == "DeletionByGCPControllerManager" {
```
droslean, line 692: Since this is GCP related only, why not make the reasons configurable too, to allow the user to add more cases based on the infrastructure?

inteon:
> > @droslean I don't think PDB would help here since you don't get a choice as to when spot instances get taken away from you.
> 
> Yep. My only concern is whether we should allow this feature to be enabled by default. Since Prow doesn't directly know what the job will do, there can be cases where the job costs will triple if we allow this by default. I would prefer to make this configuration and let the user decide based on the infrastructure.

I discovered there is already a lot of logic to restart pods (and no limit): Eviction, NodeUnreachable, a PodUnknown Phase. So, I added a global limit of 3 and applied that to all restart, including the existing logic. I haven't (yet) made the restarts disabled by default.

droslean:
> > > @droslean I don't think PDB would help here since you don't get a choice as to when spot instances get taken away from you.
> > 
> > 
> > Yep. My only concern is whether we should allow this feature to be enabled by default. Since Prow doesn't directly know what the job will do, there can be cases where the job costs will triple if we allow this by default. I would prefer to make this configuration and let the user decide based on the infrastructure.
> 
> I discovered there is already a lot of logic to restart pods (and no limit): Eviction, NodeUnreachable, a PodUnknown Phase. So, I added a global limit of 3 and applied that to all restart, including the existing logic. I haven't (yet) made the restarts in case of a Spot instance restart disabled by default, because I think the retryCount limit is a better solution.

I would prefer to let the user to decide the list of reasons to restart the prowjob and the retry count limit. For example, in my infrastructure, we run costly jobs, and this feature can potentially increase the cost since its rerunning them by default for those specific reasons. Your solution is good, but I would prefer to make it configurable so the user won't be limited on hardcoded termination reasons and retry limits.  @stevekuznetsov WDYT?

pkg/plank/reconciler.go
inteon: I made the settings configurable: https://github.com/kubernetes-sigs/prow/pull/117/commits/648600a808d9f69de344253e1667fbd13ebad5a8

stevekuznetsov:
Getting the config semantic right might be hard but I'm for it.

BenTheElder:
Yeah, I share some concerns about the restart implications.

For example, with 5k node scale tests, we may prefer to simply take the failure and leave boskos to cleanup rather than attempt to start another run, and yet with the many many CI jobs we have it would be difficult to properly identify and opt out all of the relevant jobs.

cc @aojea @pohly 

Also, even as a GCP employee, I think we should prefer to use portable Kubernetes, but I guess this is at least somewhat more portable now ... do any of the other major vendors with spot instances set a similar status that can be matched or do we need a different mechanism entirely?

What's the use case where the next periodic attempt and/or `/retest` is not sufficient?
Are you using the automatic /retest commenter? It's clunky but it has done the job ~OK.

BenTheElder:
I suspect for _most_ jobs this is better, if bounded, but it's still a surprising behavior change and there's limited bandwidth to go comb through jobs and opt them in/out.

For Kubernetes we could probably opt-out anything reference the boskos scalability pools and be in OK shape.

I don't think we want another best practice field that everyone has to opt in though either .... (like `decorate: true` which should ideally not be redundant eventually)

What if we have a global option to enable this, in addition to per job opt-out? We could wait to turn this on in until we've opted out any jobs with cost concerns?

I'm also not sure about having a single retrycount is the best approach, but I haven't put much thought into it yet. E.G. failure to schedule is pretty distinct from the other modes (Thought I think we already handle that separately?)

BenTheElder:
If we do merge this behavior default-enabled, please help us warn scalability / coordinate opting out sensitive Kubernetes-project CI.

I kinda feel like the instance-wide default should be gated behind a config option.

(if that has changed, please update the title, I'm attempting to keep up with the conversation but I'm not an active code reviewer here)

inteon:
> If we do merge this behavior default-enabled, please help us warn scalability / coordinate opting out sensitive Kubernetes-project CI.
> 
> I kinda feel like the instance-wide default should be gated behind a config option.
> 
> (if that has changed, please update the title, I'm attempting to keep up with the conversation but I'm not an active code reviewer here)

I split this PR into 2 PRs: this PR and https://github.com/kubernetes-sigs/prow/pull/412.

In https://github.com/kubernetes-sigs/prow/pull/412, I improve the existing revival logic (revival = what I call restarting in case the pod is in an error state) and add a limit.
In this PR, I add support for reviving in case of a terminated node.

~~**This PR is now blocked by https://github.com/kubernetes-sigs/prow/pull/412.** Let's first get that one merged.~~ MERGED

inteon:
With https://github.com/kubernetes-sigs/prow/pull/412 merged, this PR is now unblocked and ready to be re-reviewed.

config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml
```diff
@@ -1584,6 +1584,13 @@ spec:
                   If this field is unspecified or false, a new pod will be created to replace
                   the evicted one.
                 type: boolean
+              error_on_termination:
+                description: |-
+                  ErrorOnTermination indicates that the ProwJob should be completed and given
+                  the ErrorState status if the pod that is executing the job is terminated.
+                  If this field is unspecified or false, a new pod will be created to replace
```
BenTheElder, line 1621: I'm still concerned about making this the default behavior. I don't know if Kubernetes has the bandwidth to dig through all of our jobs and identify which should have this enabled safely or not.

I would generally expect a breaking behavior change to be opt-in

petr-muller:
@inteon I was not involved in this PR before but I think to move this forward we'll need to address @BenTheElder 's [concern](https://github.com/kubernetes-sigs/prow/pull/117#discussion_r2126928686) about this being a breaking change and his preference for this to be an opt-in feature

BenTheElder:
cc @upodroid @xmudrii @ameukam @pohly (some sig testing / k8s infra TLs for visibility on possible CI impact)

pkg/plank/reconciler.go
```diff
@@ -667,14 +674,28 @@ const (
...
 
+	// If there was a Graceful node shutdown, the Pod's status will have a
+	// reason set to "Terminated":
+	// https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
+	if pod.Status.Reason == Terminated {
+		return PodUnexpectedStopCauseTerminated
+	}
+
+	for _, condition := range pod.Status.Conditions {
+		if slices.Contains(terminationConditionReasons, condition.Reason) {
```
petr-muller, line 694: Is a condition **type** match check missing here?

petr-muller:
@BenTheElder maybe we could start with having this opt-in with a plan to eventually flip the default to opt-out once it's polished & tested? I believe the behavior is a good default but I understand the concern about default behavior being changed.

pkg/config/config.go
```diff
@@ -2526,6 +2533,19 @@ func parseProwConfig(c *Config) error {
 		c.Plank.MaxRevivals = &maxRetries
 	}
 
+	if c.Plank.TerminationConditionReasons == nil {
+		c.Plank.TerminationConditionReasons = []string{
```
droslean, line 2537: I am against defaulting those reasons in code. It should be better to let the user configure and customize the reason in the Prow config and not include any defaults at all.

pkg/config/prow-config-documented.yaml
```diff
@@ -1345,6 +1345,13 @@ plank:
     # Use `org/repo`, `org` or `*` as a key.
     report_templates:
         "": ""
+    # TerminationConditionReasons is a set of pod status condition reasons on which the
+    # controller will match to determine if the pod's node is being terminated. If a node is being
+    # terminated the controller will restart the prowjob, unless the ErrorOnTermination option
+    # is set on the prowjob or the MaxRevivals option is reached.
+    # Defaults to ["DeletionByPodGC", "DeletionByGCPControllerManager"].
```
droslean, line 1352: I don't think we should default to anything. The feature should be disabled by default to also keep backwards compatibility, and don't assume that Prow will run on specific infrastructure.

pkg/config/config.go
petr-muller: I personally think that termination reasons are infrastructure detail deep enough for Prow to carry a reasonable list of them that represent the abstract idea of "pod died for infra reason and it is worth retrying". I would not expect many admins to be able to bootstrap such config list themselves, so I'd prefer not to require them to.

PR #123, opened 2024-04-21
build images using cloudbuild

upodroid:
Part of https://github.com/kubernetes-sigs/prow/issues/113 
Part of https://github.com/kubernetes/k8s.io/pull/6740

Right now, our postsubmit job [runs](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-push-prow/1781305980542259200) in a pod which breaks our trusted cluster build policies.

I tested this build at https://console.cloud.google.com/cloud-build/builds;region=global/ce637e5c-b605-490f-b104-589b7dca23c6?project=k8s-infra-ii-sandbox

@BenTheElder @ameukam

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,9 @@
 # Prow
+[![Go Reference](https://pkg.go.dev/badge/sigs.k8s.io/prow.svg)](https://pkg.go.dev/sigs.k8s.io/prow)
+[![Go Report Card](https://goreportcard.com/badge/sigs.k8s.io/prow)](https://goreportcard.com/report/sigs.k8s.io/prow)
+[![LICENSE](https://img.shields.io/github/license/kubernetes-sigs/prow.svg)](https://github.com/kubernetes-sigs/prow/blob/main/LICENSE)
+[![Slack Status](https://img.shields.io/badge/slack-join_chat-white.svg?logo=slack&style=social)](https://kubernetes.slack.com/archives/CDECRSC5U)
+
 ![Prow Logo](./site/static/images/logo-horizontal.svg)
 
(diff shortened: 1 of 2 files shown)


.gcloudignore
upodroid, line 1: an empty .gcloudignore allows GCB to upload the .git folder which is required by hack/prowimagebuilder which runs git commands to fetch git status.

ameukam:
To my understanding, #113 states that we should attempt to promote prow components to `registry.k8s.io` ?

upodroid:
#113 requires Prow maintainers to start versioning prow and promote the release images to registry.k8s.io. In the meanwhile, we need to be able to publish and consume unversioned prow images from an AR registry called `us-docker.pkg.dev/k8s-infra-prow/images`

upodroid:
> Also a regional AR would be more sustainable and cost-effective.

Pricing is the same for both storage and bandwidth. https://cloud.google.com/artifact-registry/pricing#storage


> prow and we specially don't need to co-locate a AR registry and a GKE cluster.

I'm doing this for convenience. Prow images are a special case and will need to be stored for more than 60 days(how long is TBD) and AR has a good security boundary, unlike GCR which interacts with GCS so a dedicated staging project makes no sense.

BenTheElder:
> Pricing is the same for both storage and bandwidth. https://cloud.google.com/artifact-registry/pricing#storage

_currently_

> Prow images are a special case and will need to be stored for more than 60 days

They shouldn't be. Which other projects would we special-case?

The only special case I know currently is registry.k8s.io because we would have a circular dependency problem.

upodroid:
/cc @droslean @matthyx 

I would like to get this merged soon, thanks

matthyx:
/approve
/lgtm
Please remove the `hold` if @BenTheElder comment is addressed.

PR #135, opened 2024-04-25
Introduce an option to overwrite GitHub contexts of pending presubmit ProwJobs when merging batches in `tide`

### LINKED ISSUE #209: Branch protection rules prevent first PR of a merge-batch from being merged
**What happened**:
When `tide` is creating a new merge-batch it starts the required tests for the batch, but also trigger those tests for its member with the lowest PR number (PR 9200 in my screenshot below). 

Screenshot 2024-02-28 at 16 14 50

`tide` starts merging the PRs directly when the tests for the `merge-batch` are completed.
If there are branch-protection rules those rules prevent the PR with the lowest number (PR 9200) from being merged until its individual tests succeeded.
This results in a race between both test runs. When the `merge-batch` test run wins, all PRs except the first are merged and the first PR becomes part of the next `merge-batch`.
It is likely that the merge-batch test run wins the race, because it is created first.

Screenshot 2024-02-28 at 16 47 21

As you can see in tide history, poor PR 9200 already lost a couple of races and is still not merged yet.

Screenshot 2024-02-28 at 17 56 15

While I'm writing this issue, 9200 lost another race 😢 

I except that all PRs of a merge-batch are merged by `tide` if the tests succeed.

I could imagine two approaches to solve the issue:
1. `tide` could wait until all tests of the single PR are finished too before it starts merging
2. `tide` could still merge directly when the tests of the merge-batch succeed. In this case it could overwrite the contexts of the tests of the single PR 

You could check the history of PR 9200 in our prow instance:
https://prow.gardener.cloud/tide-history?repo=gardener%2Fgardener&pull=9200

I started implementing approach 2. in PR #135.

oliver-goetz:
Fixes #209 

With this PR `tide` overwrites the context of `pending` ProwJobs when it merges a batch of PRs. It uses the status of the succeeded batch job with the same context. The feature is disabled by default and can be enabled via `overwrite-pending-contexts` context option.
When the context is overwritten, `tide` deletes any test report comments created by crier ([ref](https://github.com/kubernetes-sigs/prow/blob/main/pkg/crier/reporters/github/reporter.go)) of this PR which might still exist. They contain previous failed runs of the contexts which have just been overwritten. It would be confusing to keep them in the PR.
Before, Github branch-protection rules could prevent merging PRs from a batch (see https://github.com/kubernetes/test-infra/issues/32097 for details).

When a PR is merged the ProwJobs which are still running are canceled automatically by trigger plugin of `hook`, so there is no need to cancel them in `tide`.

This PR replaces https://github.com/kubernetes/test-infra/pull/32416.


pkg/tide/github.go
```diff
@@ -583,3 +598,73 @@ func (mc *mergeChecker) prMergeMethod(c config.Tide, crc *CodeReviewCommon) *typ
...
+func getProwJobsForPRs(prs []CodeReviewCommon, pjs []prowapi.ProwJob) map[int]prowJobsByContext {
+	prowJobsForPRs := make(map[int]prowJobsByContext)
+	prNums := make(map[int]CodeReviewCommon)
+	for _, pr := range prs {
+		prNums[pr.Number] = pr
+	}
+
+	for _, pj := range pjs {
+		for _, pull := range pj.Spec.Refs.Pulls {
+			if pr, found := prNums[pull.Number]; found && pr.HeadRefOID == pull.SHA {
```
stevekuznetsov, line 616: Do we need to also consider the ref for the target branch? Could we have some race like:

1. trigger presubmits on target ref A
2. trigger batch for same
3. push to PR
4. trigger presubmits for target ref B
5. presubmits publish pending statuses
6. batch finishes
7. this code runs, overwriting presubmits 
8. batch would have been cancelled by the new push

oliver-goetz: Okay, I see. 
I restructured this function that it checks if all PRs of a batch job are valid before collecting it. The logic is similar to how batch PRs are accumulated ([ref](https://github.com/kubernetes-sigs/prow/blob/4359601d003f3d5caf70210ff41b3768b7cbe165/pkg/tide/tide.go#L887-L916)).
Now, the context of a batch job which contain invalid or missing PRs are not collected anymore.

oliver-goetz:
@stevekuznetsov could you check this PR again please? 🙂

oliver-goetz:
Now, any leftover test report comments will be deleted too, when the context was overwritten.
This avoids situations like shown in the screenshot where the context is marked as SUCCESS but there is still a comment with a link to a failed test.

Screenshot 2024-06-05 at 17 25 55

It might happen when the batch jobs succeed and there is a failing test in the trigger PR, which is restarted fast enough while the batch job is still running.

oliver-goetz:
ping 🔔 
Could someone review this PR please.

pkg/tide/github.go
```diff
@@ -265,6 +270,17 @@ func (gi *GitHubProvider) mergePRs(sp subpool, prs []CodeReviewCommon, dontUpdat
 			continue
 		}
 
+		// Overwrite context of pending presubmit prow jobs if enabled
```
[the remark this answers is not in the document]
oliver-goetz: I don't think it makes sense to set the status to anything else than success in the tide context. I updated the name of the function.

pkg/tide/github.go
```diff
@@ -583,3 +657,65 @@ func (mc *mergeChecker) prMergeMethod(c config.Tide, crc *CodeReviewCommon) *typ
...
+				if _, ok := prowJobsForPRs[pull.Number]; !ok {
+					prowJobsForPRs[pull.Number] = prowJobsByContext{
+						successfulBatchJob:   make(map[string]prowapi.ProwJob),
+						pendingPresubmitJobs: make(map[string]bool),
+					}
+				}
+				prowJobsForPRs[pull.Number].successfulBatchJob[pj.Spec.Context] = pj
+			}
+		case prowapi.PresubmitJob:
+			if jobState != pendingState {
```
krzyzacy, line 701: do we also want to override failed presubmit runs if the batch job finishes with success?

oliver-goetz: If a presubmit run of a batch member fails, the batch jobs continue to run. However, the batch is removed from the merge pool, so it would never reach this point here.

petr-muller:
Let's not block on me, I'll skim post-merge (eventually :D ) Thanks @krzyzacy for reviewing this one, it's great to see you back. And thanks @oliver-goetz for your submission and patience :bow: 

/lgtm

petr-muller (approved):

PR #143, opened 2024-05-07
Feature: add ding talk report support

sawyer523:
image


jihoon-seo:
Since the commit messages of two commits are not distinctive, so let me add this label:
Please consider remove it if not appropriate.

Issue #118, opened 2024-04-19
branchprotector does not honor CTRL-C


tuminoid:
Shortly: branchprotector does not honor CTRL-C and if killed, may leave threads running on the background.

I did a dry-run as [suggested by the docs](https://docs.prow.k8s.io/docs/components/optional/branchprotector/#deploy-cronjob-to-production) and wanted to cancel it as it took so long to run. 

Issue:
1. it doesn't honor CTRL-C, it does nothing
2. if suspended by CTRL-Z, and killed, it leaves worker threads behind, which keep running and spamming until separately killed

How to reproduce:
```
go run ./prow/cmd/branchprotector \
  --config-path=/path/to/config.yaml \
  --github-token-path=/path/to/my-github-token
```
Our [prow config](https://github.com/metal3-io/project-infra/blob/main/prow/manifests/overlays/metal3/config.yaml) for reference.

jihoon-seo:
We might update branchprotector code to handle OS signals.

Ref:
- [Handle CTRL-C in Golang](https://gist.github.com/jnovack/297cee036f3e5a430aa9444c0ae1b06d)
- [Is it possible to capture a Ctrl+C signal (SIGINT) and run a cleanup function, in a "defer" fashion?](https://stackoverflow.com/questions/11268943/is-it-possible-to-capture-a-ctrlc-signal-sigint-and-run-a-cleanup-function-i)
- [Exit Go program on Ctrl+C](https://stackoverflow.com/questions/24608210/exit-go-program-on-ctrlc)

tuminoid:
/assign

Well I need it fixed so I can dry-run branchprotector vs our config where it acts funny, so I'll send a PR soon.

PR #144, opened 2024-05-07
branchprotector: add support for SIGTERM/SIGINT catching

tuminoid:
Terminate branchprotector on TERM/INT signals. Otherwise we will leave threads running in the background or simply ignore the signals. For example, aborting a dry-run when testing it manually goes horribly wrong without this.

Fixes: #118

### DIFF
--- a/cmd/branchprotector/protect.go
+++ b/cmd/branchprotector/protect.go
@@ -21,10 +21,12 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"os/signal"
 	"regexp"
 	"sort"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/sirupsen/logrus"
 
@@ -132,6 +134,14 @@ func main() {
 		enableAppsRestrictions: o.enableAppsRestrictions,
 		enabled:                o.githubEnablement.EnablementChecker(),
 	}
+
+	interrupt := make(chan os.Signal, 1)
+	signal.Notify(interrupt, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		s := <-interrupt
+		logrus.Errorf("Received an interrupt: %s, cancelling...", s)
+		os.Exit(1)
+	}()
 
 	go p.configureBranches()
 	p.protect()


cmd/branchprotector/protect.go
```diff
@@ -133,6 +135,14 @@ func main() {
 		enabled:                o.githubEnablement.EnablementChecker(),
 	}
 
+	interrupt := make(chan os.Signal, 1)
+	signal.Notify(interrupt, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		s := <-interrupt
+		logrus.Errorf("Received an interrupt: %s, cancelling...", s)
+		os.Exit(1)
```
droslean, line 143: I don't think that it's safe to just exit here. Instead, we need to gracefully stop all the other goroutines. IIRC the interrupts library is already handling this case, perhaps you can take a look.

tuminoid: Thanks for the comment @droslean. I'm not a Go expert, but branchprotector is a one-shot command and not a server/service. All of the prow components that use interrupts are running an actual server, and they have context etc. Using Interrupts doesn't seem to fit here, and since branchprotector isn't a server, I'm not seeing the harm of Exiting right away? I don't see any other prow one-shot command use Interrupts either.

Can you point to relevant example to help me out?

droslean: [Here](https://github.com/kubernetes-sigs/prow/pull/144/files#diff-58095a12a640f817a91a9590a6b391dcff66b3e2560ce85713272c932e687a8dR146) we are spawning a go routine to configure the branches, and in this [line](https://github.com/kubernetes-sigs/prow/pull/144/files#diff-58095a12a640f817a91a9590a6b391dcff66b3e2560ce85713272c932e687a8dR147) another operation starts which both share multiple channels to communicate. Instead of force exiting, we should ensure that the channels are closed and all go routines stopped, we should gather the errors and then exit.

petr-muller: TBH @droslean I tend to agree that there's little to gain for a one-shot command implementing graceful shutdown...

droslean: @petr-muller I agree that force-exiting can result in not updating some repositories of a specific org, which is not a big deal, but eating an error in the `configureBranches()` go routine, shouldn't be wise. Ref https://github.com/kubernetes-sigs/prow/blob/main/cmd/branchprotector/protect.go#L185 that is being consumed in https://github.com/kubernetes-sigs/prow/blob/main/cmd/branchprotector/protect.go#L139. In addition, implementing a graceful exiting should be trivial if the context can be utilized appropriately.

petr-muller (approved):
Let's merge this as is. Eating errors is not great but so is not respecting signals. The change is in a good direction, adding proper error handling in the presence of a SIGINT or whatever can be added if/when needed.

PR #146, opened 2024-05-08
fix(pkg): fallback at using git client when github API files limits is reached

FedeDP:
Github API has an hard limit of 3000 files returned for push/pull request events. Workaround this when checking pre/post submit jobs to be triggered, by manually diffing using the git client.
See https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files.

Porting of https://github.com/kubernetes/test-infra/pull/30615 to the new repo.

### DIFF
--- a/pkg/config/jobs.go
+++ b/pkg/config/jobs.go
@@ -541,7 +541,7 @@ func NewGitHubDeferredChangedFilesProvider(gc git.ClientFactory, client githubCl
 			if len(changes) == github.ChangesFilesLimit && gc != nil {
 				repoClient, err := gc.ClientFor(org, repo)
 				if err == nil {
-					// Use git client since Github PushEvent is limited to 3000 keys:
+					// Use git client since github PushEvent is limited to 3000 keys:
 					// https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files
 					files, err := repoClient.Diff(headSHA, baseSHA)
 					if err == nil {
--- a/pkg/plugins/trigger/push.go
+++ b/pkg/plugins/trigger/push.go
@@ -46,7 +46,7 @@ func listPushEventChanges(gc git.ClientFactory, pe github.PushEvent) config.Chan
 			if len(changes) == github.ChangesFilesLimit && gc != nil {
 				repoClient, err := gc.ClientFor(pe.Repo.Owner.Name, pe.Repo.Name)
 				if err == nil {
-					// Use git client since Github PushEvent is limited to 3000 keys:
+					// Use git client since github PushEvent is limited to 3000 keys:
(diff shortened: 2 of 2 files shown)


pkg/config/jobs.go
```diff
@@ -534,8 +536,25 @@ func NewGitHubDeferredChangedFilesProvider(client githubClient, org, repo string
 			if err != nil {
 				return nil, fmt.Errorf("error getting pull request changes: %w", err)
 			}
-			for _, change := range changes {
-				changedFiles = append(changedFiles, change.Filename)
+
+			// Fallback to use gitClient since github API truncated the response
+			if len(changes) == github.ChangesFilesLimit && gc != nil {
```
FedeDP, line 541: As suggested here: https://github.com/kubernetes/test-infra/pull/30615#discussion_r1339994398 this is now a fallback, that is only executed when:
* gitclient is not nil
* length of changes is exactly 3000

Of course, this means that if unluckily number of changes is exactly 3000, we would run the git client flow for nothing.

pkg/plugins/trigger/push.go
```diff
@@ -21,27 +21,44 @@ import (
...
 			}
 			for _, modified := range commit.Modified {
-				changed[modified] = true
+				changes[modified] = true
 			}
 		}
 		var changedFiles []string
-		for file := range changed {
-			changedFiles = append(changedFiles, file)
+		if len(changes) == github.ChangesFilesLimit && gc != nil {
```
FedeDP, line 46: As above for pull requests, only load the new code if exactly 3000 files changes are found.

FedeDP:
Note: i am still trying to understand how to test these changes; i am currently building all images and pushing them under my own dockerhub account; then i might switch our https://github.com/falcosecurity/test-infra repo to use my own image and see if triggering a job that failed before, works fine now.

pkg/plugins/trigger/push.go
```diff
@@ -21,28 +21,46 @@ import (
...
 	prowapi "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	"sigs.k8s.io/prow/pkg/config"
+	"sigs.k8s.io/prow/pkg/git/v2"
 	"sigs.k8s.io/prow/pkg/github"
 	"sigs.k8s.io/prow/pkg/pjutil"
 )
 
-func listPushEventChanges(pe github.PushEvent) config.ChangedFilesProvider {
+func listPushEventChanges(gc git.ClientFactory, pe github.PushEvent) config.ChangedFilesProvider {
+	var changedFiles []string
```
FedeDP, line 30: This should be the only actual "breaking change", ie: cache `changedFiles` just like we do in `pull-request.go`, to avoid a git client `diff` call for each configured post submit when we are in the fallback case.

FedeDP:
/remove-lifecycle stale

Sorry i still hadn't a change to test this.

PR #147, opened 2024-05-13
Implement ConfigTree in approve plugin

### LINKED ISSUE #221: prow config: where possible, all config should behave like branchprotector for overrides
initial issue: https://github.com/kubernetes/test-infra/issues/9757

The branchprotector override mechanism is amazing and we should use it in every config where people can configure orgs or repos, etc, to allow for more specific overrides. More specifically, plugins like approve have a flat list of configs and don't have any override logic, but users expect it and commit configuration that does nothing.

matthyx:
https://github.com/kubernetes/test-infra/pull/12440

Fixes #221


matthyx:
@petr-muller since it is almost you and me, are we interested in improving the config structs?

PR #155, opened 2024-05-17
fix typo in serviceaccount format

upodroid:


upodroid:
/cc @petr-muller 

Error message: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8s-infra-prow-images/1791465441835618304

PR #156, opened 2024-05-17
Revert "`gangway`: add status change endpoint"

petr-muller:
Reverts kubernetes-sigs/prow#116

Causes test flakes and its test panic


upodroid (approved):

petr-muller:
Eek why there are postsubmits here?

petr-muller:
oh... I used GH revert button which actually creates a new branch here, because I have write rights... and our postsubmits are not restricted to main branch, which is probably a mistake for these cases :thinking: would I push Prow if the postsubmits worked? :D

petr-muller:
/override post-k8s-infra-prow-images

I'll fix that in job config by restricting branches

PR #159, opened 2024-05-21
Revert "Revert "`gangway`: add status change endpoint""

Prucek:
Original PR: https://github.com/kubernetes-sigs/prow/pull/116
Reverts kubernetes-sigs/prow#156
Resolves:  [kubernetes-sigs/prow#155](https://github.com/kubernetes-sigs/prow/issues/115)

The problem was in the `WaitForJobExecutionStatus()` method, it was only waiting for the execution to exist and not for the actual status to appear, and then of course, sometimes it was `pending` and sometimes `triggered`, and that caused the flakiness. A simple [true->false](https://github.com/kubernetes-sigs/prow/pull/159/files#diff-6ee23a21e73e43adbaa398d6ff301be035bccd13dff69a112b3fc0574115dabaL48)  solved it.


petr-muller:
Can you drop a sentence or two into the description about what the problem was and what is the fix?

pkg/gangway/gangway.go
```diff
@@ -135,13 +147,145 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	// TODO:
+	// All ProwJob need to be listed, because FieldSelectors are not supported by CRDs yet.
+	// Once FieldSelectors are supported (Kubernetes 1.30 maybe), we can filter the ProwJob list by the desired fields.
+	// Issue link: https://github.com/kubernetes/kubernetes/issues/53459
+	pjList, err := gw.ProwJobClient.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		logrus.WithError(err).Debug("failed to list ProwJobs")
```
smg247, line 198: I think this log should be `error` level, this really shouldn't ever come up, but if it does it is important to address.

pkg/gangway/gangway.go
```diff
@@ -135,13 +147,145 @@ func (gw *Gangway) GetJobExecution(ctx context.Context, gjer *GetJobExecutionReq
...
+	for _, pj := range pjList.Items {
+		if !isMatchingCondition(pj, request) {
+			continue
+		}
+		if allowedApiClient != nil {
+			authorized := ClientAuthorized(allowedApiClient, pj)
+			if !authorized {
+				logrus.Error("client is not authorized to modify the given job")
+				return nil, status.Error(codes.PermissionDenied, "client is not authorized to modify the given job")
+			}
```
smg247, line 212: Ideally, these authorization errors would be collected and returned at the end. We should still modify the jobs that the client is authorized to modify without returning early

smg247:
/test pull-prow-integration

let's try it once more ;)

pkg/gangway/gangway.go
```diff
@@ -195,10 +196,11 @@ func (gw *Gangway) BulkJobStatusChange(ctx context.Context, request *BulkJobStat
 	// Issue link: https://github.com/kubernetes/kubernetes/issues/53459
 	pjList, err := gw.ProwJobClient.List(ctx, metav1.ListOptions{})
 	if err != nil {
-		logrus.WithError(err).Debug("failed to list ProwJobs")
+		logrus.WithError(err).Errorf("failed to list ProwJobs")
 		return nil, err
 	}
 
+	errs := []error{}
```
smg247, line 203: Generally, we use:
var errs []error

petr-muller:
/test integration
/approve

@smg247 seeing you already reviewed the code I'm leaving the lgtm up to you ;)

PR #164, opened 2024-05-27
[Feature]: [assign plugin] Support GitHub Organization Teams as targets for assign/cc commands

### LINKED ISSUE #145: [Feature]: [assign plugin] Support GitHub Organization Teams as targets for assign/cc commands
Would it be possible to support GitHub Organization Teams as targets for the assign/cc commands?

Currently, I am working on the contextual logging feature, which follows a specific review process for PRs:

```
/cc [@]kubernetes/wg-structured-logging-reviews
```

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#what-to-include-in-the-pull-request

However, prow does not support `/cc <teams>`, resulting in an error.

In most cases, the reviewers selected by the blunderbuss plugin based on the OWNERS file are sufficient. However, for PRs like the one below, the hack directory may not necessarily include members of [@]kubernetes/wg-structured-logging-reviews.
https://github.com/kubernetes/kubernetes/pull/124722

If `/cc <teams>` were supported, I think it would simplify the review request process for above case.
(This is because there would be no need to send review requests to specific individuals within the team.)

**Implementation proposal:**
- Allow `/cc <teams>` and `/assign <teams>` commands to randomly select one member from the specified team as a reviewer or assignee. We could discuss whether adjusting the number of selections might be necessary.
- [Implement this using the `layeredsets.PopRandom()` function, already utilized in the blunderbuss plugin, for random selection](https://github.com/kubernetes-sigs/prow/blob/c5e374dc7eb1aa4283a37d4a8f68bd67f52e22ad/pkg/plugins/blunderbuss/blunderbuss.go#L328-L359).
- Ensure that the GitHub teams used for PRs within the Kubernetes org are limited to those within the Kubernetes org, maintaining organizational alignment.

I would appreciate your thoughts on this issue.

/kind feature

Okabe-Junya:
close #145 

## Feature Description

Support `/cc <teams>` `/assign <teams>` command to request review(s) to teams,

**Example**

For example, let’s assume that the members of the `kubernetes/sig-team-foo` team are Alice, Bob, and Carol.
In this case, `/cc kubernetes/sig-team-foo` command will request reviews from all three members: Alice, Bob, and Carol.
Of course,  `/assign` command works in the same way.

### DIFF
--- a/pkg/plugins/assign/assign.go
+++ b/pkg/plugins/assign/assign.go
@@ -71,6 +71,8 @@ type githubClient interface {
 	UnrequestReview(org, repo string, number int, logins []string) error
 
 	CreateComment(owner, repo string, number int, comment string) error
+
+	ListTeamMembersBySlug(org, teamSlug, role string) ([]github.TeamMember, error)
 }
 
 func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error {
@@ -82,6 +84,10 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error 
 		err = combineErrors(err, handle(newReviewHandler(e, pc.GitHubClient, pc.Logger)))
 	}
 	return err
+}
+
+func isTeamLogin(login string) bool {
+	return strings.Contains(login, "/")
 }
 
 func parseLogins(text string) []string {
@@ -125,7 +131,17 @@ func handle(h *handler) error {
 			users[e.User.Login] = add
 		} else {
 			for _, login := range parseLogins(re[2]) {
-				users[login] = add
+				if isTeamLogin(login) {
+					teamMembers, err := h.gc.ListTeamMembersBySlug(org, login, "all")
+					if err != nil {
+						return fmt.Errorf("failed to list team members: %w", err)
+					}
+					for _, member := range teamMembers {
+						users[member.Login] = add
+					}
+				} else {
+					users[login] = add
+				}
 			}
 		}
 	}
--- a/pkg/plugins/assign/assign.go
+++ b/pkg/plugins/assign/assign.go
@@ -87,7 +87,11 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error 
 }
 
 func isTeamLogin(login string) bool {
-	return strings.Contains(login, "/")
+	match, err := regexp.MatchString(`^[a-zA-Z0-9]+/[a-zA-Z0-9]+$`, login)
+	if err != nil {
+		return false
+	}
+	return match
 }
 
 func parseLogins(text string) []string {
--- a/pkg/plugins/assign/assign.go
+++ b/pkg/plugins/assign/assign.go
@@ -87,7 +87,7 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error 
 }
 
 func isTeamLogin(login string) bool {
-	match, err := regexp.MatchString(`^[a-zA-Z0-9]+/[a-zA-Z0-9]+$`, login)
+	match, err := regexp.MatchString(`^[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*/[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*$`, login)
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


Okabe-Junya:
Since you are  the author of #145, let me cc @bells17 

/cc bells17

pkg/plugins/assign/assign_test.go
```diff
@@ -95,6 +95,16 @@ func (c *fakeClient) CreateComment(owner, repo string, number int, comment strin
 	return nil
 }
 
+func (c *fakeClient) ListTeamMembersBySlug(org, teamSlug, role string) ([]github.TeamMember, error) {
+	if teamSlug == "kubernetes/sig-testing-misc" {
```
droslean, line 99: You should add this when creating a new fake client. In this method, you could check if the teamSlug key exists in the corresponding map.

pkg/plugins/assign/assign.go
```diff
@@ -84,6 +86,10 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error
 	return err
 }
 
+func isTeamLogin(login string) bool {
+	return strings.Contains(login, "/")
```
droslean, line 90: This will also include the case `something/` which is non-valid. Let's try to check the length of the slice if you split it by `/` or even better match a regex expression

Okabe-Junya: Thanks!! I fixed it to check if it's empty (https://github.com/kubernetes-sigs/prow/pull/164/commits/7440b238f87c02402755b1390b31154410e0ac87)

pkg/plugins/assign/assign_test.go
Okabe-Junya: I implemented this like the below section, Is there anything else I should do?

https://github.com/kubernetes-sigs/prow/blob/ea10bd8144c3d988528011af024abe47ea8dabb2/pkg/plugins/override/override_test.go#L299-L307

pkg/plugins/assign/assign.go
```diff
@@ -84,6 +86,19 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error
 	return err
 }
 
+func isTeamLogin(login string) bool {
```
droslean, line 89: IMHO the following is a better approach:
```go
func isTeamLogin(login string) bool {
	match, err := regexp.MatchString(`^[a-zA-Z0-9]+/[a-zA-Z0-9]+$`, login)
	if err != nil {
		return false
	}
	return match
}

```

droslean (commented):
/hold

My only concern is that there should be already people that are doing `/cc @org/team`  with the expectation that prow won't assign all people to the PR. This change will change this behavior. I would prefer to make this configurable and disable it by default.

@petr-muller @stevekuznetsov  Any thoughts?

pkg/plugins/assign/assign.go
Okabe-Junya: Thank you!!

fixed it https://github.com/kubernetes-sigs/prow/pull/164/commits/e04afd17ba3b2f5e0bc114d3eaa314b69d80538a

Okabe-Junya:
> My only concern is that there should be already people that are doing `/cc @org/team` with the expectation that prow won't assign all people to the PR.

This is a good point, but I don’t really agree with this. The reason is that executing this command results in an error from prow (as mentioned in the issue #145).

If you want to mention the team simply, wouldn’t it be done like `cc @org/team`? (without the `/` and not as a prow command)

pkg/plugins/assign/assign.go
Okabe-Junya: I made a few more corrections, this is because it didn’t work well with teams or organizations that include a “-”.

https://github.com/kubernetes-sigs/prow/pull/164/commits/4af6ef4f7ec331aa223947220fc83732419df45c

pkg/plugins/assign/assign_test.go
droslean: It's a nit. I guess this can be fixed in a follow-up for all fakeclients.

Okabe-Junya:
It looks like I need to fix some CI errors...

Okabe-Junya:
Great!! Now, I fixed all CI errors!!

PTAL again @droslean, to move forward this PR :) 

(plz execute `/lgtm` again), Thank you!!

Okabe-Junya:
Thank you for your kind support droslean!!

PR #183, opened 2024-06-07
Remove-CRD-size-removing-PipelineRunSpec

cizara:
the previous CRD failed to install as it's too big and the Kubernetes API  server rejects it with:

one or more objects failed to apply, reason: rpc error: code = ResourceExhausted desc = trying to send message larger than max (2171854 vs. 2097152) (retried 2 times).

By removing PipelineRunSpec and using TektonPipelineRunSpec (accepts V1 and V1beta1) the size is acceptable again


cizara:
sorry, don't want to create this pr here

PR #184, opened 2024-06-07
Use Cluster and not Manager for build clusters

alvaroaleman:
At the time controller-runtime was introduced to this project, it didn't have a clean abstraction for "cache + client of a cluster". The only thing it had was the manager, which is that plus some additional things like metrics, webhooks, leader election and code related to run things. As a result, we ended up using the manager and disabling everything it doesn't need.

By now, controller-runtime has the `cluster` abstraction for precisely this use-case. This change makes us use the `cluster` where appropriate. It should not result in any outwards-visible behavior changing.

### DIFF
--- a/cmd/prow-controller-manager/main.go
+++ b/cmd/prow-controller-manager/main.go
@@ -167,20 +167,20 @@ func main() {
 		interrupts.Terminate()
 	}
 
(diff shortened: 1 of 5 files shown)


droslean:
/lgtm
/approve
/hold cancel

Thanks @alvaroaleman !!!!

PR #188, opened 2024-06-11
feat(pkg/plugins/dco): ignore merged type commits

wuhuizuo:
Keep same with DCO GitHub app

DCO app will ignore the merge type commits:
https://github.com/dcoapp/app/blob/main/lib/dco.js#L15-L19
image
So the branch updating triggered by contributor will not make the DCO check failed. 

But the prow's "dco" plugin will fail when the branch has new merge type commits introduced by branch updating.


droslean:
/ok-to-test

@wuhuizuo More details about this change would be much appreciated.

wuhuizuo:
> /ok-to-test
> 
> @wuhuizuo More details about this change would be much appreciated.

@droslean  I added some details in description. PTAL

droslean (approved):

Issue #181, opened 2024-06-03
ProwJob CRD is too large


maboehm:
Hi,

the prowJob CR just recently crossed the threshold of being too large, for example running against a kind cluster:

```sh
$ kind create cluster
...
$ kubectl version
Client Version: v1.30.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.0

# this is the last commit this still works on
$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/prow/e67659d368fd013492a9ce038d801ba8998b7d10/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml --server-side
customresourcedefinition.apiextensions.k8s.io/prowjobs.prow.k8s.io serverside-applied

# latest version fails
$ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/prow/main/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml --server-side
Error from server: etcdserver: request is too large
```

Am I doing something wrong when applying?

In terms of solving this issue, the main problem is that there is a lot of repetition, e.g. the `taskPodTemplate` appears several times. It would be possible to have some fields as e.g. `apiextension.JSON` or `runtime.RawExtension` and defer validation to a validatingwebhook. Which makes the output of e.g. `kubectl explain` less helpful.

matthyx:
> ```shell
> kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/prow/main/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml --server-side
> ```

(apparently it doesn't, sorry)

maboehm:
Either way, since tools like Helm or Flux use `apply` the size should work for that as well.

lentzi90:
I have been using `apply --server-side=true` to avoid the annotation. It looks like the CRD itself is not finally too large. With server side apply it worked up until https://github.com/kubernetes-sigs/prow/commit/e67659d368fd013492a9ce038d801ba8998b7d10 (including). From https://github.com/kubernetes-sigs/prow/commit/64052956637fb92301edded100fcf18e49930ed5 it is no longer possible to apply.

matthyx:
ok well I think we have to shrink it... @maboehm wanna have a try?

PR #196, opened 2024-06-17
Don't embed tekton schema in ProwJob CRD

maboehm:
The CRD has become too large for standard kubernetes / etcd installations. With this change, the CRD generation is instructed to treat all `tektoncd` types as schema-less (while instructing the apiserver to keep unknown fields). This reduces the size down to around 500KB

Seems like upstream, tekton is doing something similar, their CRD does not have a schema: https://github.com/tektoncd/pipeline/blob/main/config/300-crds/300-pipelinerun.yaml#L62-L72

I also updated the `admission` component to use strict validation and to be able to handle CREATE requests as well. With this component deployed, you would get at least some level of safety back, even if the response is a bit lacking right now:

```
# non-existing field
Error from server: error when creating "test/integration/test/testdata/valid_bare_prowjob.yaml": admission webhook "prowjob-validator.prow.k8s.io" denied the request: decode new: strict decoding error: unknown field "spec.pipeline_run_spec.foo"

# wrong type on field
Error from server: error when creating "test/integration/test/testdata/valid_bare_prowjob.yaml": admission webhook "prowjob-validator.prow.k8s.io" denied the request: decode new: json: cannot unmarshal string into Go struct field PipelineRunSpec.spec.pipeline_run_spec.pipelineRef of type v1beta1.PipelineRef
```

(I also did not see the admission component installed in https://github.com/kubernetes/test-infra/tree/master/config/prow so not sure who uses it. We dont have it in our env.)

Fixes https://github.com/kubernetes-sigs/prow/issues/181


matthyx:
awesome @maboehm can you check the CLA?

maboehm:
Just an additional comment: we dont use the tekton stuff in our setup, so might be worth double checking (but I did check that the fields are in fact preserved, so confidence is high)

maboehm:
It's been two weeks, should we just move forward?

matthyx:
> It's been two weeks, should we just move forward?

I'd say yes.
/approve
/hold

PR #200, opened 2024-06-27
Update github.com/prometheus/common module

danilo-gemoli:
As per title, update the `prometheus/common` module:
```sh
go get github.com/prometheus/common@v0.54.0
```

What follow is the sole breaking change I have hit.
The field:
```golang
expfmt.FmtProtoDelim
```
is no longer public and it has been replaced by this function:
```golang
expfmt.NewFormat(expfmt.TypeProtoDelim)
```
They both result in the same string:
```golang
fmtProtoDelim        Format = protoFmt + ` encoding=delimited`
```
so we should be safe.

Check https://github.com/prometheus/common/pull/576 for more information.


danilo-gemoli:
We are running out of resources:
```
There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (0/46 nodes are available: 3 node(s) had untolerated taint {node-group: stable}, 43 Insufficient cpu. preemption: 0/46 nodes are available: 3 Preemption is not helpful for scheduling, 43 No preemption victims found for incoming pod..)
```

PR #203, opened 2024-07-02
gangway: BulkJobStatusChange: return early and finish later

Prucek:
This change will accept the `BulkJobStatusChange` request and return immediately and it will finish the processing in a separate goroutine. On bigger prow instances, with a lot of Prowjobs, it was observed, that it takes a lot of time to only to list all Prowjobs, and calls to this endpoint caused: `context timeout`


smg247:
This takes a bit of a different direction than what I was expecting. My impression is that this service will nearly always be used when there are more than 10 jobs to update. If we have 10 or less, we will likely find it just as easy to just manually update each one. With that context, I think it makes sense to always return the list of jobs to be updated, prior to actually updating any of them. Then spawn a new thread to update all the jobs. What do you think about that approach?

pkg/gangway/gangway.go
```diff
@@ -170,87 +173,96 @@ func TranslateProwJobType(prowJobType prowcrd.ProwJobType) JobExecutionType {
...
+		return &emptypb.Empty{}, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	var errs []error
-	jobExecutions := []*JobExecution{}
-	for _, pj := range pjList.Items {
-		if !isMatchingCondition(pj, request) {
-			continue
+	// creating a context that does not get cancelled and finish the task in the background
+	context, _ := context.WithTimeout(context.WithoutCancel(ctx), CONTEXT_TIMEOUT)
```
smg247, line 197: Can this be initialized closer to where it is used?

pkg/gangway/gangway.go
```diff
@@ -170,87 +173,96 @@ func TranslateProwJobType(prowJobType prowcrd.ProwJobType) JobExecutionType {
...
 
-		pj.Status.State = prowcrd.ProwJobState(strings.ToLower(request.GetJobStatusChange().GetDesired().String()))
-		updatedPj, err := gw.ProwJobClient.Update(ctx, &pj, metav1.UpdateOptions{})
-		if err != nil {
-			logrus.WithError(err).Errorf("failed to update ProwJob status")
-			errs = append(errs, fmt.Errorf("%s, failed to update ProwJob status: %s", err, pj.Name))
-			continue
-		}
-		logrus.WithField("name", pj.Name).Infof("ProwJob status updated to: %s", updatedPj.Status.State)
+	return &emptypb.Empty{}, nil
```
smg247, line 238: Do you know what status code this ends up returning?

pkg/gangway/gangway.go
```diff
@@ -170,87 +173,96 @@ func TranslateProwJobType(prowJobType prowcrd.ProwJobType) JobExecutionType {
...
+	}
+	if request.JobType == JobExecutionType_PERIODIC {
+		labelSelector.MatchLabels["prow.k8s.io/type"] = string(prowcrd.PeriodicJob)
 	}
-	jobsAffected := &JobsAffected{
-		Count:         int32(jobLen),
-		JobExecutions: jobExecutions,
+	if request.Refs != nil {
+		if request.Refs.GetOrg() != "" {
+			labelSelector.MatchLabels["prow.k8s.io/refs.org"] = request.Refs.GetOrg()
```
smg247, line 254:
```suggestion
		if org := request.Refs.GetOrg(); org != "" {
			labelSelector.MatchLabels["prow.k8s.io/refs.org"] = org
```

pkg/gangway/gangway.go
```diff
@@ -170,87 +173,96 @@ func TranslateProwJobType(prowJobType prowcrd.ProwJobType) JobExecutionType {
...
+		}
+		if request.Refs.GetRepo() != "" {
```
smg247, line 256: same as above

pkg/gangway/gangway.go
```diff
@@ -170,87 +173,96 @@ func TranslateProwJobType(prowJobType prowcrd.ProwJobType) JobExecutionType {
...
-		jobExec := &JobExecution{
-			Id:        updatedPj.Name,
-			JobName:   updatedPj.Spec.Job,
-			JobType:   JobExecutionType(TranslateProwJobType(updatedPj.Spec.Type)),
-			JobStatus: TranslateProwJobStatus(&updatedPj.Status),
-		}
-		jobExecutions = append(jobExecutions, jobExec)
+func getRequestLabelSelector(request *BulkJobStatusChangeRequest) *metav1.LabelSelector {
+	labelSelector := &metav1.LabelSelector{MatchLabels: make(map[string]string)}
+	if request.JobType == JobExecutionType_PRESUBMIT {
```
smg247, line 243: this could be a switch statement

test/integration/test/gangway_test.go
```diff
@@ -515,24 +516,22 @@ func TestGangwayBulkJobStatusChange(t *testing.T) {
...
 				}
 			} else {
-				if jobsAffected.Count != int32(tt.expectedCount) {
-					t.Fatalf("Expected %d jobs to be affected, got %d", tt.expectedCount, jobsAffected.Count)
-				}
-				for _, job := range jobsAffected.JobExecutions {
-					if job.JobStatus != gangway.JobExecutionStatus_ABORTED {
-						t.Fatalf("Expected job status to be %q, got %q", gangway.JobExecutionStatus_ABORTED, job.JobStatus)
+				for _, jobExecution := range jobExecutions {
+					timeout := 120 * time.Second
```
smg247, line 526: how did you decide on the timeout? that seems like a long time to wait during an integration test, does it actually take that long?

Prucek: No, it doesn't take long, I copied it from the test above

pkg/gangway/gangway.go
Prucek: I have not tried it, because I had to change the proto file, I will check that, but it should be `OK 0`

Prucek: yes, it is so and return this: 
`{}`

pkg/gangway/gangway.go
```diff
@@ -170,87 +173,97 @@ func TranslateProwJobType(prowJobType prowcrd.ProwJobType) JobExecutionType {
...
+			}
+			pj.Status.State = prowcrd.ProwJobState(strings.ToLower(request.GetJobStatusChange().GetDesired().String()))
+			updatedPj, err := gw.ProwJobClient.Update(context, &pj, metav1.UpdateOptions{})
+			if err != nil {
+				logrus.WithError(err).Errorf("failed to update ProwJob status")
+				continue
+			}
+			logrus.WithField("name", pj.Name).Infof("ProwJob status updated to: %s", updatedPj.Status.State)
 		}
+		cancel()
```
smg247, line 237: Do we need this? Is it possible to just use `_` instead of creating the `cancel` var?

Prucek: yes, unfortunately, otherwise the linter job is failing

droslean (approved):

Issue #204, opened 2024-07-04
[BUG] CC multiple people does not work anymore


gregpavl:
**What happened**:

We upgraded the prow components from `v20231227` to `v20240515`. 

example: 

```
-        image: gcr.io/k8s-prow/prow-controller-manager:v20231227-1f5d684b52
+        image: gcr.io/k8s-prow/prow-controller-manager:v20240515-dd5d0eeaa
```

Before the upgrade, 
the command `/cc @person_1 @person_2 @person_3 @person_4` was CC the GH usernames correctly (all at once) in the PRs. 

After the upgrade, 
the CC functionality seems broken. 
If you try to CC multiple people, the bot will catch only the first CCed user.

**What you expected to happen**:

to continue CC multiple people with the same command inline at once.

**How to reproduce it (as minimally and precisely as possible)**:

try to CC people on the same line. `/cc @person_1 @person_2 @person_3 @person_4` 
you can even try to `/cc @person_X` multiple people in the same comment. It has the same behavior.

Right now, you need to do CC people in separate comments. 

Eg. 
Comment 1: 
`/cc @person_1` 
Comment 2: 
`/cc @person_2` 
Comment 3: 
`/cc @person_3` 
Comment 4: 
`/cc @person_4` 

**Please provide links to example occurrences, if any**:

n/a

**Anything else we need to know?**:

- k8s cluster: 1.28.10-do.0 (v. 1.28.10 - managed by DO)
- github org | free

---

original ref: https://github.com/kubernetes/test-infra/issues/32905

droslean:
My first thought is maybe https://github.com/kubernetes-sigs/prow/pull/164/ is related. @Okabe-Junya Any ideas?

Okabe-Junya:
Upon checking the git history, it seems that the only change to the assign plugin after the migration to k-sigs/prow was my PR(#164). 

I apologize if this caused any inconvenience. I have created a PR to revert #164, which might serve as a useful workaround for now.

Okabe-Junya:
What I don't understand is that while my PR was merged on **May 28**, the tag where the bug happened is `v20240515`. (Is this tag based on the date the image was built?)

Moreover, my PR only changed the behavior for teams, so it shouldn't have affected the `/cc Alice Bob ...` format command at all. All elements should result in the `isTeamLogin` function returning false, meaning that simply `users[login] = add` would be executed.

[...]

PR #206, opened 2024-07-04
Fix assign plugins

Okabe-Junya:
this PR is following up to #164 

## Bug Description

[`ListTeamMembersBySlug` function](https://github.com/kubernetes-sigs/prow/blob/79d27b6e3be35974fbe103d3f574d70dfea6f03c/pkg/github/client.go#L3758-L3786) requires `org` and `teamSlug` as arguments, but I mistakenly passed org and login as arguments.

```go
teamMembers, err := h.gc.ListTeamMembersBySlug(org, login, "all")
```

Meaning, I was passing arguments like `kubernetes, kubernetes/sig-misc-approvers` - The correct arguments are `kubernetes, sig-misc-approvers`

## Change List 

- Split the `login` into org and teamSlug before passing them to the function

### DIFF
--- a/pkg/plugins/assign/assign.go
+++ b/pkg/plugins/assign/assign.go
@@ -136,7 +136,11 @@ func handle(h *handler) error {
 		} else {
 			for _, login := range parseLogins(re[2]) {
 				if isTeamLogin(login) {
-					teamMembers, err := h.gc.ListTeamMembersBySlug(org, login, "all")
+					org, teamSlug := strings.Split(login, "/")[0], strings.Split(login, "/")[1]
+					if org == "" || teamSlug == "" {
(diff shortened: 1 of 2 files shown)


droslean:
@Okabe-Junya Is this the fix for https://github.com/kubernetes-sigs/prow/issues/204 ?

Okabe-Junya:
@droslean 

It's not certain, but since this code clearly contains a bug, I think it's a good to move this PR forward.

ref. https://github.com/kubernetes-sigs/prow/issues/204#issuecomment-2209429428

droslean:
@Okabe-Junya It would be better to revert the whole feature and you can open another PR to re-introduce it with the fix as well.

PR #211, opened 2024-07-26
Use k8s-staging-test-infra images for base images.

michelle192837:
Switch the image bases to use those built in k8s-staging-test-infra instead. Ref https://github.com/kubernetes/test-infra/issues/32432.

Atm: Just doing this for the `alpine` and `git-custom-k8s-auth` images.

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,50 +1,50 @@
 # Distroless images:
 # defaultBaseImage: gcr.io/distroless/static:nonroot
 baseImageOverrides:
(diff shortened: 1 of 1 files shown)


michelle192837:
/hold

Actually, I think the autobump needs an update first before this.

michelle192837:
/hold cancel

Wait no, these aren't autobumped, that migration will need to happen once we switch the actual k8s-prow/<component> images over to the ones built in the new location. This should be fine.

Need to look into the test failure though.

ameukam:
/lgtm
cc @droslean @matthyx

PR #217, opened 2024-08-02
Specify the image file for image build/push.

michelle192837:
The default is `prow/.prow-images.yaml`, which made sense when Prow was in k8s/test-infra since it was nested in a subdirectory, but doesn't make sense anymore now that the files are unnested in their own repo. We could update the default instead, but I think having the argument here explicitly is clearer in this crops up again in future.

I believe this should fix the image push job for Prow images (`post-k8s-infra-prow-images`), which doesn't have any published images in k8s-staging-test-infra yet.

Ref https://github.com/kubernetes/test-infra/issues/32432

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,22 +106,22 @@ PROW_IMAGE ?=
 
 .PHONY: push-images
 push-images:
-	hack/make-rules/go-run/arbitrary.sh run ./hack/prowimagebuilder --ko-docker-repo="${REGISTRY}" --push=true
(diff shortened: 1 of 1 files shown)


michelle192837:
Update: I am actually just bad at finding images, and confused the default in k8s/test-infra with the default in this repo's version of hack/prowimagebuilder (which correctly points to the same file).

In any case, this job has been working and is still working after the change (which should be a no-op). Sorry for the confusion!

PR #222, opened 2024-08-07
Fix panic in gerrit client ChangeExist

nathanperkins:
The old code assumed that when `err != nil`, the `resp` is still populated, which isn't true in all branches. This causes a panic which can occur every 6 hours

We add a `resp != nil` check to protect against this panic.

Also, add a unit test and fake gerritChange with errors which would catch this panic. I confirmed that removing the change triggers the test failure.


cjwagner (approved):
/ok-to-test

cjwagner (approved):

PR #229, opened 2024-08-13
fakejira: use key instead of name for issue creation

AlexNPavel:
Jira uses the project key to construct the issue key.


bradmwilliams (approved):
/lgtm

petr-muller (approved):

PR #230, opened 2024-08-14
New scheduling strategy: external

jmguzik:
`External` sends `Job` field from prowjob spec to URL and expects cluster assignment in return

### DIFF
--- a/pkg/config/prow-config-documented.yaml
+++ b/pkg/config/prow-config-documented.yaml
@@ -1348,6 +1348,15 @@ push_gateway:
 # It has to be explicitly enabled.
 scheduler:
     enabled: true
+    external:
+        # Cache is the cache configuration for the external scheduling strategy
+        cache:
+            # CacheCleanupInterval is the interval to clean up the cache
+            cleanup_interval: 0s
+            # EntryTimeoutInterval is the interval to consider an entry in the cache
+            entry_timeout_interval: 0s
+        # URL is the URL of the REST API to query for the cluster assigned to prowjob
+        url: ' '
     # Scheduling strategies
     failover:
         # ClusterMappings maps a cluster to another one. It is used when we
--- a/pkg/plugins/plugin-config-documented.yaml
+++ b/pkg/plugins/plugin-config-documented.yaml
@@ -295,7 +295,11 @@ cat:
     # Path to file containing an api key for thecatapi.com
     key_path: ' '
 cherry_pick_approved:
-    - # Approvers is the list of GitHub logins allowed to approve a cherry-pick.
+    - # AllowMissingApprovedLabel allows approving cherry-pick without the approved label.
+      allow_missing_approved_label: true
+      # AllowMissingLGTMLabel allows approving cherry-pick without the lgtm label.
+      allow_missing_lgtm_label: true
+      # Approvers is the list of GitHub logins allowed to approve a cherry-pick.
       approvers:
         - ""
       # BranchRegexp is the regular expression for branch names such that
--- a/pkg/config/scheduler.go
+++ b/pkg/config/scheduler.go
@@ -16,11 +16,14 @@ limitations under the License.
 
 package config
 
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 type Scheduler struct {
 	Enabled bool `json:"enabled,omitempty"`
 
 	// Scheduling strategies
 	Failover *FailoverScheduling `json:"failover,omitempty"`
+	External *ExternalScheduling `json:"external,omitempty"`
 }
 
 // FailoverScheduling is a configuration for the Failover scheduling strategy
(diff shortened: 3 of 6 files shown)


pkg/plugins/plugin-config-documented.yaml
```diff
@@ -295,7 +295,11 @@ cat:
     # Path to file containing an api key for thecatapi.com
     key_path: ' '
 cherry_pick_approved:
-    - # Approvers is the list of GitHub logins allowed to approve a cherry-pick.
+    - # AllowMissingApprovedLabel allows approving cherry-pick without the approved label.
```
jmguzik, line 298: generated by code-gen, seemed missed by somebody

pkg/config/prow-config-documented.yaml
```diff
@@ -1355,6 +1355,13 @@ scheduler:
         # configured to in the first place.
         mappings:
             "": ""
+    rest_by_job:
+        # Interval is the interval to clean up the cache
+        interval: 0s
```
droslean, line 1360: Perhaps we can use a better name for this. Maybe `cache_cleanup_interval` ?

pkg/config/prow-config-documented.yaml
```diff
@@ -1355,6 +1355,13 @@ scheduler:
...
+        # Timeout is the timeout for individual cache entry
+        timeout: 0s
```
droslean, line 1362: And for this as well. Since both are configurations for the cache only, perhaps you can consider adding them under a `cache` stanza.

pkg/scheduler/strategy/rest_by_job.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// query sends a POST request to the specified URL with the provided request body
+// and returns the response as a SchedulingResponse object
+func query(url string, reqBody SchedulingRequest) (SchedulingResponse, error) {
+	jsonData, err := json.Marshal(reqBody)
+	if err != nil {
+		return SchedulingResponse{}, fmt.Errorf("error marshaling JSON: %w", err)
+	}
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
```
droslean, line 70: This is a requirement that the users need to understand to build their REST API for this purpose. I would prefer to have a library for the user to import to the server. In addition, it would make more sense to implement this feature using gRPC and allow users to just import a protobuf library and just start their server accepting specifying messages that will be pre-configured by the library.

pkg/scheduler/strategy/rest_by_job.go
```diff
@@ -0,0 +1,128 @@
...
+
+	rbj.cache[pj.Spec.Job] = &cacheEntry{
+		r:         resp,
+		timestamp: time.Now(),
+	}
+	return Result(resp), nil
+}
+
+// cleanupCache removes stale entries from the cache
+func (rbj *RESTByJob) cleanupCache() {
```
droslean, line 117: This is running when a job is scheduled. Why not have another mechanism where the cache is being cleaned after every N time?

pkg/config/prow-config-documented.yaml
jmguzik: will do

pkg/config/prow-config-documented.yaml
jmguzik: good idea

pkg/scheduler/strategy/rest_by_job.go
jmguzik: I would stick with rest for the time being as req/resp is simple enough. We could add grpc if needed as additional strategy if needed in the future.

pkg/scheduler/strategy/rest_by_job.go
jmguzik: It's invoking this function every time, yes, but if you look at the implementation of the fcn, in most of the cases function quits immediately as interval is not met. So, effectively, it will clean map every `n` hour. I think this is better then after every n time, more understandable for the user.

jmguzik: I'll change it to be more visible (move to Schedule fcn).

pkg/config/prow-config-documented.yaml
```diff
@@ -1355,6 +1355,15 @@ scheduler:
         # configured to in the first place.
         mappings:
             "": ""
+    rest_by_job:
+        # Cache is the cache configuration for the REST by job scheduling strategy
+        cache:
+            # CacheCleanupInterval is the interval to clean up the cache
+            cache_cleanup_interval: 0s
```
droslean, line 1362: Since its already under the `cache` stanza, we can just name it as  `cleanup_interval`

danilo-gemoli (changes requested):

pkg/scheduler/strategy/rest_by_job.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type cacheEntry struct {
+	r         SchedulingResponse
+	timestamp time.Time
+}
+
+// RestByJob is a strategy that schedules a ProwJob on a specific cluster
+// based on the response of the ProwJob's Job name to given URL
+type RESTByJob struct {
```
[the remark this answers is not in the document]
jmguzik: Plugin does 2 things. Takes `Job` field and sends it via `REST`. Changing to grpc means new plugin. How outsourcing is a better name? It tells nothing about what is happening.

pkg/scheduler/strategy/rest_by_job.go
```diff
@@ -0,0 +1,126 @@
...
+	cfg       config.RESTByJobScheduling
+	cache     map[string]*cacheEntry
+	timestamp time.Time
+}
+
+// NewRESTByJob creates a new RestByJob instance with caching
+func NewRESTByJob(cfg config.RESTByJobScheduling) *RESTByJob {
```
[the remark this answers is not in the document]
jmguzik: where is this config.Getter as other strategies (failover) are doing the same what is done here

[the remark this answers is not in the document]
danilo-gemoli: My bad, it seems I have done it already https://github.com/kubernetes-sigs/prow/blob/16c13cde5644f1c40747c358ad3deb77899f45ce/pkg/scheduler/reconciler.go#L85

PR #232, opened 2024-08-16
cherrypickapproved: implement org-level configuration

xmudrii:
This PR implements the org-level configuration for the `cherrypickapproved` plugin. Right now, it's possible to configure this plugin only on the repo-level. If you have multiple repos, you have to repeat the config for each repo, which is inefficient and clutters the configuration file, especially if the config is the same for all repos.

With this PR, such a configuration is allowed:

```yaml
cherry_pick_approved:
  - org: org-name
    branchregexp: '^release/v.*$'
    approvers:
      - ...
```

The repo-level config has the priority over the org level configuration. This also means that this is not a breaking change, i.e. the existing configurations will continue working as before.

/assign @saschagrunert 
cc @kubernetes-sigs/release-engineering

### DIFF
--- a/pkg/plugins/cherrypickapproved/cherrypickapproved.go
+++ b/pkg/plugins/cherrypickapproved/cherrypickapproved.go
@@ -135,10 +135,18 @@ func (h *handler) handle(log *logrus.Entry, gc plugins.PluginGitHubClient, e git
 	)
 
 	// Filter configurations
-	foundRepoOrg := false
+	foundConfig := false
 	for _, cfg := range cfgs {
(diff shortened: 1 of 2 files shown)


matthyx:
@xmudrii you might want to take a look at https://github.com/kubernetes-sigs/prow/pull/147 where I'm trying to propose a generic solution that could be applied to all plugins

xmudrii:
In addition to https://github.com/kubernetes-sigs/prow/pull/147, I'd like that we take a different route here, so I'm going to close this PR and start over again if needed.
/close

PR #237, opened 2024-08-16
Add Orgs sharding for Blunderbuss

dhaiducek:
Using a pointer lets sharding determine whether a config had been provided so they can be properly compared. If one is not provided, this also instantiates one in `setDefaults()`, which is called following the sharding logic.

Migrated from:
- https://github.com/kubernetes/test-infra/pull/29405

### DIFF
--- a/pkg/genyaml/populate_struct.go
+++ b/pkg/genyaml/populate_struct.go
@@ -127,6 +127,10 @@ func PopulateStruct(in any) any {
 			if strings.Contains(typeOf.Elem().Field(i).Tag.Get("json"), ",omitempty") {
 				valueOf.Elem().Field(i).SetBool(true)
 			}
+		case reflect.Int:
+			if strings.Contains(typeOf.Elem().Field(i).Tag.Get("json"), ",omitempty") {
+				valueOf.Elem().Field(i).SetInt(1)
+			}
 		}
 
 	}
--- a/pkg/plugins/blunderbuss/blunderbuss.go
+++ b/pkg/plugins/blunderbuss/blunderbuss.go
@@ -59,25 +59,47 @@ func configString(reviewCount int) string {
 }
 
 func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
-	var reviewCount int
+	reviewCount := 2
 	if config.Blunderbuss.ReviewerCount != nil {
 		reviewCount = *config.Blunderbuss.ReviewerCount
 	}
-	two := 2
 	yamlSnippet, err := plugins.CommentMap.GenYaml(&plugins.Configuration{
 		Blunderbuss: plugins.Blunderbuss{
-			ReviewerCount:         &two,
-			MaxReviewerCount:      3,
-			ExcludeApprovers:      true,
-			UseStatusAvailability: true,
-			IgnoreAuthors:         []string{},
+			BlunderbussConfig: &plugins.BlunderbussConfig{
+				ReviewerCount:         &reviewCount,
+				MaxReviewerCount:      3,
+				ExcludeApprovers:      true,
+				UseStatusAvailability: true,
+				IgnoreAuthors:         []string{},
+			},
+			Orgs: map[string]plugins.BlunderbussOrgConfig{
+				"": {
+					BlunderbussConfig: &plugins.BlunderbussConfig{
+						ReviewerCount:         &reviewCount,
+						MaxReviewerCount:      3,
+						ExcludeApprovers:      true,
+						UseStatusAvailability: true,
+						IgnoreAuthors:         []string{},
+					},
+					Repos: map[string]plugins.BlunderbussConfig{
+						"": {
+							ReviewerCount:         &reviewCount,
+							MaxReviewerCount:      3,
+							ExcludeApprovers:      true,
+							UseStatusAvailability: true,
+							IgnoreAuthors:         []string{},
+						},
+					},
+				},
+			},
 		},
 	})
 	if err != nil {
 		logrus.WithError(err).Warnf("cannot generate comments for %s plugin", PluginName)
 	}
 	pluginHelp := &pluginhelp.PluginHelp{
-		Description: "The blunderbuss plugin automatically requests reviews from reviewers when a new PR is created. The reviewers are selected based on the reviewers specified in the OWNERS files that apply to the files modified by the PR.",
+		Description: "The blunderbuss plugin automatically requests reviews from reviewers when a new PR is created. " +
+			"The reviewers are selected based on the reviewers specified in the OWNERS files that apply to the files modified by the PR.",
 		Config: map[string]string{
 			"": configString(reviewCount),
 		},
@@ -140,14 +162,14 @@ func handlePullRequestEvent(pc plugins.Agent, pre github.PullRequestEvent) error
 		pc.GitHubClient,
 		pc.OwnersClient,
 		pc.Logger,
-		pc.PluginConfig.Blunderbuss,
+		pc.PluginConfig.BlunderbussFor(pre.Repo.Owner.Login, pre.Repo.Name),
 		pre.Action,
 		&pre.PullRequest,
 		&pre.Repo,
 	)
 }
 
-func handlePullRequest(ghc githubClient, roc repoownersClient, log *logrus.Entry, config plugins.Blunderbuss, action github.PullRequestEventAction, pr *github.PullRequest, repo *github.Repo) error {
+func handlePullRequest(ghc githubClient, roc repoownersClient, log *logrus.Entry, config plugins.BlunderbussConfig, action github.PullRequestEventAction, pr *github.PullRequest, repo *github.Repo) error {
 	// Ignore pull request event if:
 	// - not an opened or ready for review event
 	// - if they have a /cc command in the description (will be handled elsewhere)
@@ -181,7 +203,7 @@ func handleGenericCommentEvent(pc plugins.Agent, ce github.GenericCommentEvent) 
 		pc.GitHubClient,
 		pc.OwnersClient,
 		pc.Logger,
-		pc.PluginConfig.Blunderbuss,
+		pc.PluginConfig.BlunderbussFor(ce.Repo.Owner.Login, ce.Repo.Name),
 		ce.Action,
 		ce.IsPR,
 		ce.Number,
@@ -191,7 +213,7 @@ func handleGenericCommentEvent(pc plugins.Agent, ce github.GenericCommentEvent) 
 	)
 }
 
-func handleGenericComment(ghc githubClient, roc repoownersClient, log *logrus.Entry, config plugins.Blunderbuss, action github.GenericCommentEventAction, isPR bool, prNumber int, issueState string, repo *github.Repo, body string) error {
+func handleGenericComment(ghc githubClient, roc repoownersClient, log *logrus.Entry, config plugins.BlunderbussConfig, action github.GenericCommentEventAction, isPR bool, prNumber int, issueState string, repo *github.Repo, body string) error {
 	if action != github.GenericCommentActionCreated || !isPR || issueState == "closed" {
 		return nil
 	}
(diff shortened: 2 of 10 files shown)


jmguzik (commented):
@dhaiducek Are you still interested and willing to push it further? 
I feel really bad about this one as I was involved here https://github.com/kubernetes/test-infra/pull/28169 and I think we even pushed it, but missed detail about pointer. Correct me if I am wrong.

This code was reviewed so generally from my POV there is nothing major to comment. This change is huge though and it can affect again prow installations, including openshift one. I am just loudly thinking how we should approach the testing regarding the topic. 

It would be ideal to test it in some kind of puppy environment, but it's not straightforward to setup, but maybe worth a try.

dhaiducek:
> @dhaiducek Are you still interested and willing to push it further?

Hey @jmguzik! Yes, I'm willing to dive back into it--it's still relevant for us. No hard feelings--I got pretty gun-shy myself when it merged and broke things and had to be reverted, and I know it's a massive change. But yes, as I recall it broke last time because the implementation didn't use a pointer for the global config and didn't correctly identify when another global config wasn't specified.

I wouldn't know the first thing about setting up a sandbox Prow env, but I can try or let me know if there's anything I can do to help! Before, if I'm looking at the thread correctly, it failed during the rehearsal CI and never made it into production: https://redhat-internal.slack.com/archives/CBN38N3MW/p1683036380568929

jmguzik:
I'll test that on puppy env during next shift week to see if the current impl behaves :wink:

pkg/plugins/config.go
```diff
@@ -901,6 +923,27 @@ func (c *Configuration) ApproveFor(org, repo string) *Approve {
...
+		}
+
+		// Return org configuration if defined
+		if orgConfig.BlunderbussConfig != nil {
+			return *orgConfig.BlunderbussConfig
+		}
+	}
+
+	// Return base config
+	return *c.Blunderbuss.BlunderbussConfig
```
jmguzik, line 944:
```suggestion
	if c.Blunderbuss.BlunderbussConfig != nil {
		return *c.Blunderbuss.BlunderbussConfig
	}
	return BlunderbussConfig{}
```

makes sense?

pkg/plugins/config.go
```diff
@@ -2193,6 +2245,69 @@ func (l *Label) mergeFrom(other *Label) error {
...
+
+	// Merge Orgs configs, skipping conflicts and adding a message to the aggregated error message
+	for org, otherOrgConfig := range other.Orgs {
+		if orgConfig, ok := b.Orgs[org]; ok {
+			// Use incoming org if current org config is empty, otherwise verify they're the same
+			if otherOrgConfig.BlunderbussConfig != nil {
+				if orgConfig.BlunderbussConfig == nil {
+					orgConfig.BlunderbussConfig = otherOrgConfig.BlunderbussConfig
+					b.Orgs[org] = orgConfig
+				} else if !reflect.DeepEqual(&orgConfig.BlunderbussConfig, &otherOrgConfig.BlunderbussConfig) {
```
jmguzik, line 2280: Are we comparing here values or pointers?

pkg/plugins/config.go
```diff
@@ -2193,6 +2245,69 @@ func (l *Label) mergeFrom(other *Label) error {
...
+			// Initialize Repos map if it's empty to accept incoming Repos configs
+			if otherOrgConfig.Repos != nil && orgConfig.Repos == nil {
+				orgConfig.Repos = map[string]BlunderbussRepoConfig{}
+				b.Orgs[org] = orgConfig
+			}
+			// Merge Repos configs, skipping conflicts and adding a message to the aggregated error message
+			for repo, otherRepoConfig := range otherOrgConfig.Repos {
+				if repoConfig, ok := orgConfig.Repos[repo]; ok {
+					// Verify the repo configurations are the same
+					if !reflect.DeepEqual(&repoConfig.BlunderbussConfig, &otherRepoConfig.BlunderbussConfig) {
```
jmguzik, line 2294: Are we comparing here values or pointers?

pkg/plugins/config.go
```diff
@@ -2193,6 +2245,69 @@ func (l *Label) mergeFrom(other *Label) error {
...
+
+	var errs []error
+
+	// Handle global configs
+	if other.BlunderbussConfig != nil {
+		if b.BlunderbussConfig == nil {
+			b.BlunderbussConfig = other.BlunderbussConfig
+		} else if !reflect.DeepEqual(b.BlunderbussConfig, other.BlunderbussConfig) {
+			// Add error when both configs declare different global defaults
+			errs = append(errs, fmt.Errorf("global configurations for blunderbuss do not match"))
```
jmguzik, line 2263: should that not be stopping error?

jmguzik (commented):
I would like to get this closer to merging this week. Left some comments/suggestions.

pkg/plugins/config.go
dhaiducek: Not necessarily. By continuing on, it can aggregate all of the errors, not just the mismatch of the default values.

pkg/plugins/config.go
dhaiducek: This compares values.

pkg/plugins/config.go
dhaiducek: This compares values

pkg/plugins/config.go
dhaiducek: It could make sense, but I'm not sure it's necessary since the `nil` case is handled here in `setDefaults()`:
https://github.com/kubernetes-sigs/prow/pull/237/files#diff-19905b632cb1213462d0e01d336d4f49504923b1365fceca80506f7f9a755a61R1129-R1132

dhaiducek (commented):
/remove-lifecycle stale

Sorry, I let this one flounder and get stale--I've left some responses to the review comments.

pkg/plugins/config.go
dhaiducek: Oops--no it doesn't! Let me fix these. Good catch.

petr-muller:
We recently got a very good reason to get this merged, thanks for keeping it fresh! I started to review the PR and it looks good skimmed, I'll do some testing and look a bit closer and I hope we'll get this done soon.

dhaiducek:
> We recently got a very good reason to get this merged, thanks for keeping it fresh! I started to review the PR and it looks good skimmed, I'll do some testing and look a bit closer and I hope we'll get this done soon.

Sounds great, thanks @petr-muller! I'll keep an eye out for reviews/followups (and might have to look over the code to refresh myself on it too! 😅).

pkg/plugins/blunderbuss/blunderbuss.go
petr-muller, line 266: I think `handleStatus` and `handleStatusEvent` need to weave the `BlunderbussFor(...)` result instead of the full config from `PluginConfig`, right?

pkg/plugins/config.go
petr-muller, line 1273: I dont thick the check is _that_ useful (among other things, other fields are not checked...) but this should descend into `b.Orgs` and check the layers below.

petr-muller (commented):
wow, it was criminal of us to not merge this sooner, it's pretty polished

my only question would be whether the `handleStatus` path should get the fragment config instead of the full one like the other handlers (inline)

I also noticed the validation is not done on fragments but that one is not that important as there's barely any validation there (although, in sharded configs validation becomes more important, so adding a shard validation that would do more checks would be a great followup)

dhaiducek:
@petr-muller Thanks for the review. Somehow I missed it and it got completely buried! 🤦‍♂️ I'll take a look at the comments today.

pkg/plugins/blunderbuss/blunderbuss.go
dhaiducek: Great catch--yes, it should!

pkg/plugins/config.go
dhaiducek: Indeed. I've added also added another check that ReviewerCount isn't > MaxReviewerCount in the process. I couldn't think of any other validation needed offhand.

dhaiducek:
I've made a couple of changes that might be considered outside the scope of this PR, so let me know if I should split them out (or revert them):

- ~Made the `ReviewerCount` an `int` instead of `*int`. Since there's validation that it's greater than zero, I think the new logic makes sense but I'd be okay reverting it to if there are concerns about it being a breaking change, or moving it to a separate PR.~
- For the generated plugin config YAML, set any `omitempty` integer values to `1` so that they'd show up. This is similar to setting the empty space for the string.

pkg/plugins/config.go
dhaiducek: Not sure what I was thinking. Might as well protect from panics! I've added the `nil` check. 🙂

pkg/plugins/config.go
```diff
@@ -1393,11 +1422,76 @@ func validateExternalPlugins(pluginMap map[string][]ExternalPlugin) error {
...
+		if b.WaitForStatus.Description == "" {
+			b.WaitForStatus.Description = "Not mergeable. (PullRequest is missing sufficient approving GitHub review\\(s\\)|Needs (lgtm|approved|approved, lgtm) labels?)\\.?"
+		}
 	}
-	return nil
+}
+
+// descendShards recurses the Blunderbuss config into the global config, org
+// configs, and repo configs, calling the provided function for each config.
+func (b *Blunderbuss) descendShards(f func(*BlunderbussConfig) error) error {
```
dhaiducek, line 1457: Created this convenience function to run a function against each level of the sharded configuration.

pkg/plugins/config.go
```diff
@@ -1561,6 +1655,23 @@ func validateRepoMilestone(milestones map[string]Milestone) {
 	}
 }
 
+func (b *Blunderbuss) compileRegexpsAndDurations() error {
```
dhaiducek, line 1668: I realized I missed populating the new `DescriptionRe` field for the `WaitForStatus` field.

pkg/plugins/config.go
```diff
@@ -1393,11 +1422,76 @@ func validateExternalPlugins(pluginMap map[string][]ExternalPlugin) error {
...
+	// Instantiate a global Blunderbuss config if it wasn't set in the config
+	if b.BlunderbussConfig == nil {
+		b.BlunderbussConfig = &BlunderbussConfig{}
+	}
+	// Set default reviewer count if it's zero or unset
+	if b.ReviewerCount == 0 {
+		b.ReviewerCount = defaultBlunderbussReviewerCount
+	}
+	// Set default "wait for status" fields if they're unset
+	if b.WaitForStatus != nil {
```
dhaiducek: Wanted to make sure since this was relatively new: was the expectation that this be unpopulated by default? So it'd get populated with defaults if a user specified `waitForStatus: {}`, or should it have gotten populated by default if it's _not_ provided, i.e. `== nil`? I assume the former to avoid a change in default behavior, but just wanted to verify.

pkg/genyaml/populate_struct.go
```diff
@@ -127,6 +127,10 @@ func PopulateStruct(in any) any {
 			if strings.Contains(typeOf.Elem().Field(i).Tag.Get("json"), ",omitempty") {
 				valueOf.Elem().Field(i).SetBool(true)
 			}
+		case reflect.Int:
+			if strings.Contains(typeOf.Elem().Field(i).Tag.Get("json"), ",omitempty") {
+				valueOf.Elem().Field(i).SetInt(1)
```
dhaiducek, line 132: This sets `omitempty` `int` fields for the YAML doc so they're not omitted when marshalling (tangentially related to this PR since `MaxReviewerCount` has an `omitempty` tag.

pkg/plugins/config.go
```diff
@@ -1393,11 +1422,86 @@ func validateExternalPlugins(pluginMap map[string][]ExternalPlugin) error {
 	return nil
 }
 
-func validateBlunderbuss(b *Blunderbuss) error {
-	if b.ReviewerCount != nil && *b.ReviewerCount < 1 {
-		return fmt.Errorf("invalid request_count: %v (needs to be positive)", *b.ReviewerCount)
+func (b *Blunderbuss) SetDefaults() {
```
dhaiducek, line 1425: Added a `SetDefaults()` function that descends into the config to populate `ReviewerCount` and `WaitForStatus` to maintain consistent default setting behaviors.

dhaiducek (commented):
All right, sorry for the major churn here. I think I've got the changes settled out now if someone could take another look. Preserved the ReviewerCount pointer but added consistent default setting behaviors across the shards so that if it's not provided it'll be set to 2 (and if it's set to zero, it'll throw a validation error). Also added some other missing defaulting behaviors for sharding.

PR #248, opened 2024-08-20
Update repoowners.go

ijun111:
/sig appsaliases:
  sig-foo:
    - david
    - erin
  sig-bar:
    - bob
    - frankemeritus_approvers:
- david    # 2018-05-02
- emily    # 2019-01-05ijun111:patch-1release-engineering

### DIFF
--- a/pkg/repoowners/repoowners.go
+++ b/pkg/repoowners/repoowners.go
@@ -1,4 +1,4 @@
-/*
+/sig apps/*
 Copyright 2017 The Kubernetes Authors.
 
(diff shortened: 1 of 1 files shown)


cjwagner (changes requested):
Hi @ijun111
I don't understand the motivation of this PR and I don't think this will pass our boilerplate validation.
I see your GitHub account is only a few days old and has submitted similar changes to many other repositories which seems a bit suspicious so I'm going to go ahead and close this PR. 
Please provide more context about this change if I am mistaken.

PR #252, opened 2024-08-20
Allow Trigger To Watch Any State - pjutil

midnightconman:
Sometimes you only want to wait for pending states


matthyx:
/lgtm
I have nothing against this, just letting other approvers have a look.

pkg/pjutil/trigger.go
```diff
@@ -81,10 +87,10 @@ func TriggerAndWatchProwJob(o prowflagutil.KubernetesOptions, prowjob *pjapi.Pro
 		}
 	}
 
-	if result.State != pjapi.SuccessState {
+	if result.State != state {
```
droslean, line 90: The state is already hardcoded in [here](https://github.com/kubernetes-sigs/prow/pull/252/files#diff-e7923d41ca0936576991fa85b88304b28d35473e13f79d39f3153a2aaac39fc6R55) to a SuccessState. I don't see any reason of passing it as an argument.

midnightconman: I don't understand what you are referring to? The old signature of `TriggerAndWatchProwJob` has a hard coded success state, but the new signature of `TriggerAndWatchProwJobState` takes a `pjapi.ProwJobState` as a param.

Do you see differently?

Issue #269, opened 2024-09-05
PR with "Change requested" leads to Tide repeatedly attempting MERGE


sdowell:
**What happened**:

For a Github PR that has passed all checks but has the `Change requested` state, Tide repeatedly attempts to MERGE but fails (presumably rejected by Github due to the `Change requested` state).

**What you expected to happen**:

I would expect Prow/Tide to be aware of the `Change requested` state and that it is blocking - and therefore not try to perform MERGE.

**How to reproduce it (as minimally and precisely as possible)**:

- Create a pull request
- Request changes on the PR
- Pass all other checks (e.g. presubmits, lgtm, approve, etc)
- Observe that Tide repeatedly attempts to perform MERGE and fails

**Please provide links to example occurrences, if any**:

Here is an example PR and tide history:
- https://github.com/GoogleContainerTools/kpt-config-sync/pull/210
- https://oss.gprow.dev/tide-history?repo=GoogleContainerTools%2Fkpt-config-sync&branch=main&pull=210

**Anything else we need to know?**:
N/A

---

Copied over from https://github.com/kubernetes/test-infra/issues/27750

petr-muller:
Copying some relevant information here from a [previous attempt to fix](https://github.com/kubernetes-sigs/prow/pull/270) this via approve plugin:

Many advanced branch protection features enforcing PR merge eligibility are younger than Tide and the big Prow orgs (kube, openshift) are mostly not using them. Generally Tide expects a PR to be mergeable once it passes Tide's criteria (query match + required jobs pass) and considers inability to merge to be "[user errors](https://github.com/kubernetes-sigs/prow/blob/7c1e4e90b5121d1c4a5547f069d70053707e45e5/pkg/tide/github.go#L274)".

I think that in today's world this does not stand. I'd prefer the solution to be

1. Tide to properly surface any occurrences when it thinks it may merge but is disallowed to do so
2. Tide to check as many mergeability properties as possible in its [isAllowedToMerge](https://github.com/kubernetes-sigs/prow/blame/7c1e4e90b5121d1c4a5547f069d70053707e45e5/pkg/tide/github.go#L520) method to avoid thinking it may merge a PR when GH would disallow it

[...]

PR #270, opened 2024-09-05
fix: do not apply approved label if change requested

sdowell:
An outstanding CHANGES_REQUESTED state prevents a PR from being merged even if another reviewer approves the PR. This change makes it so that the approved label cannot be added until all change requested states have been resolved.

This prevents a scenario where tide will repeatedly try to merge a PR and be rejected by GitHub due to the changes requested state.

Fixes: https://github.com/kubernetes-sigs/prow/issues/269

### DIFF
--- a/pkg/plugins/approve/approve.go
+++ b/pkg/plugins/approve/approve.go
@@ -590,15 +590,21 @@ func addApprovers(approversHandler *approvers.Approvers, approveComments []*comm
 			continue
 		}
 
-		if reviewActsAsApprove && c.ReviewState == github.ReviewStateApproved {
-			approversHandler.AddApprover(
-				c.Author,
-				c.HTMLURL,
-				false,
-			)
+		if c.ReviewState == github.ReviewStateApproved {
+			approversHandler.RemoveChangeRequested(c.Author)
+			if reviewActsAsApprove {
+				approversHandler.AddApprover(
+					c.Author,
+					c.HTMLURL,
+					false,
+				)
+			}
 		}
-		if reviewActsAsApprove && c.ReviewState == github.ReviewStateChangesRequested {
-			approversHandler.RemoveApprover(c.Author)
+		if c.ReviewState == github.ReviewStateChangesRequested {
+			approversHandler.AddChangeRequested(c.Author)
+			if reviewActsAsApprove {
+				approversHandler.RemoveApprover(c.Author)
+			}
 		}
 
 		for _, match := range commandRegex.FindAllStringSubmatch(c.Body, -1) {
--- a/pkg/plugins/approve/approve_test.go
+++ b/pkg/plugins/approve/approve_test.go
@@ -1054,11 +1054,14 @@ Approvers can cancel approval by writing ` + "`/approve cancel`" + ` in a commen
 <!-- META={"approvers":["alice","cjwagner"]} -->`,
 		},
 		{
-			name:                "approve command supersedes simultaneous changes requested review",
-			hasLabel:            false,
-			files:               []string{"a/a.go"},
-			comments:            []github.IssueComment{},
-			reviews:             []github.Review{newTestReview("Alice", "/approve", github.ReviewStateChangesRequested)},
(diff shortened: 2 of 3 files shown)


petr-muller:
/hold

> An outstanding CHANGES_REQUESTED state prevents a PR from being merged even if another reviewer approves the PR.

I may be wrong (will have to look at GH docs) but I don't think the above is universally true, IIRC this was a branch protection configuration option.

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-pull-request-reviews-before-merging

Says:

> If a person with admin permissions chooses the Request changes option in a review, then that person must approve the pull request before the pull request can be merged. If a reviewer who requests changes on a pull request isn't available, anyone with write permissions for the repository can dismiss the blocking review.

I'm not convinced the PR is the right fix. When `reviewActsAsApprove` is not enabled, I don't think `approve` should take it into account at all. If the repo is configured via BP to require reviews for merging (leading Tide to loop merge attempts) then it is basically misconfigured and Tide should surface this problem better (I believe we had an issue for this class of problems - it was not just request changes, but other cases where BP prevents PRs from merging but Tide does not know about it).

sdowell:
> I may be wrong (will have to look at GH docs) but I don't think the above is universally true, IIRC this was a branch protection configuration option.

I think you're right, I see also the wording of the BP setting:

> When enabled, pull requests targeting a matching branch require a number of approvals and no changes requested before they can be merged.

Is there a doc/guide anywhere on how to properly set up GitHub, and branch protection in particular, for Prow/Tide?

> If the repo is configured via BP to require reviews for merging (leading Tide to loop merge attempts) then it is basically misconfigured and Tide should surface this problem better (I believe we had an issue for this class of problems - it was not just request changes, but other cases where BP prevents PRs from merging but Tide does not know about it).

Yeah this is really my main gripe, the behavior came as a bit of a surprise and it took some digging for us to understand what was going wrong. I see it as more of an implementation detail how it's addressed, but would like for the error to be more transparent. Do you have any suggestions?

petr-muller:
> Is there a doc/guide anywhere on how to properly set up GitHub, and branch protection in particular, for Prow/Tide?

Unfortunately I'm not aware of any. Many advanced branch protection features enforcing PR merge eligibility are younger than Tide and the big Prow orgs (kube, openshift) are mostly not using them. Generally Tide expects a PR to be mergeable once it passes Tide's criteria (query match + required jobs pass) and considers inability to merge to be "[user errors](https://github.com/kubernetes-sigs/prow/blob/7c1e4e90b5121d1c4a5547f069d70053707e45e5/pkg/tide/github.go#L274)".

I think that in today's world this does not stand. I'd prefer the solution to be 

1. Tide to properly surface any occurrences when it *thinks* it may merge but is disallowed to do so
2. Tide to check as many mergeability properties as possible in its [isAllowedToMerge](https://github.com/kubernetes-sigs/prow/blame/7c1e4e90b5121d1c4a5547f069d70053707e45e5/pkg/tide/github.go#L520) method to avoid thinking it may merge a PR when GH would disallow it

It is not entirely straightforward. 1) is a problem because the status context (which right now is the only PR-facing Tide user interface, maybe with the exception of Tide PR dashboard which people don't really use that much) is controlled by a separate controller than the one that actually tries to merge. 2) is also not easy because checking mergeability for BP would require Tide asking about more properties of the PR (and possibly the repo, to check BP requirements).

sdowell:
> I think that in today's world this does not stand. I'd prefer the solution to be
> 1. Tide to properly surface any occurrences when it thinks it may merge but is disallowed to do so
> 2. Tide to check as many mergeability properties as possible in its [isAllowedToMerge](https://github.com/kubernetes-sigs/prow/blame/7c1e4e90b5121d1c4a5547f069d70053707e45e5/pkg/tide/github.go#L520) method to avoid thinking it may merge a PR when GH would disallow it

Ack - closing this PR as the solution should take a different approach

PR #272, opened 2024-09-11
Fallback to config in external strategy

jmguzik:
Fallback to cluster configured in the config file when unsuccessfull with scheduling for external strategy

### DIFF
--- a/pkg/scheduler/strategy/external.go
+++ b/pkg/scheduler/strategy/external.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	prowv1 "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	"sigs.k8s.io/prow/pkg/config"
 )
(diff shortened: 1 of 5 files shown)


pkg/scheduler/strategy/external.go
```diff
@@ -99,7 +102,8 @@ func (e *External) Schedule(_ context.Context, pj *prowv1.ProwJob) (Result, erro
 
 	resp, err := query(e.cfg.URL, SchedulingRequest{Job: pj.Spec.Job})
 	if err != nil {
-		return Result{}, fmt.Errorf("error querying URL: %w", err)
+		logrus.WithField("job", pj.Spec.Job).WithField("cluster", pj.Spec.Cluster).Warn("scheduling failed, using Spec.Cluster entry")
```
smg247, line 105: you aren't using the logger that you added to `External`

pkg/scheduler/strategy/__debug_bin1472196796
smg247, line 1: I assume this is an accident?

jmguzik: yes :(

pkg/scheduler/strategy/external_test.go
```diff
@@ -124,6 +125,7 @@ func TestExternalSchedule(t *testing.T) {
 				},
 				cache:     tt.cache,
 				timestamp: time.Now().Add(-3 * time.Hour), // To trigger cache cleanup if mockCleanup is true
+				log:       logrus.NewEntry(logrus.New()),
```
droslean, line 128:
```suggestion
				log:       logrus.WithField("name", tt.name)
```

PR #273, opened 2024-09-11
`blunderbuss`: allow `request_count` to be set to '0' effectively disabling functionality

smg247:
There is a situation where a repository doesn't want `blunderbuss`, but it is configured at the org level. Since there is no way to exclude the plugin from functioning on the repo, we must have some way to disable the functionality assigning reviewers. We can do this by allowing `request_count` to be `0`

For: https://issues.redhat.com/browse/DPTP-4164

### DIFF
--- a/pkg/plugins/blunderbuss/blunderbuss.go
+++ b/pkg/plugins/blunderbuss/blunderbuss.go
@@ -148,6 +148,9 @@ func handlePullRequest(ghc githubClient, roc repoownersClient, log *logrus.Entry
 	if !(action == github.PullRequestActionOpened || action == github.PullRequestActionReadyForReview) || assign.CCRegexp.MatchString(pr.Body) {
 		return nil
 	}
+	if config.ReviewerCount != nil && *config.ReviewerCount == 0 {
+		return nil
+	}
 	if pr.Draft && config.IgnoreDrafts {
 		// ignore Draft PR when IgnoreDrafts is true
 		return nil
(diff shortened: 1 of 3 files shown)


cjwagner:
> Since there is no way to exclude the plugin from functioning on the repo

This isn't true since February 2021. Check out the `excluded_repos` field in the plugin configuration.
https://github.com/kubernetes/test-infra/issues/20631, https://github.com/kubernetes/test-infra/pull/20707

smg247:
> > Since there is no way to exclude the plugin from functioning on the repo
> 
> This isn't true since February 2021. Check out the [`excluded_repos`](https://github.com/search?q=repo%3Akubernetes-sigs%2Fprow+excluded_repos&type=code) field in the plugin configuration. [kubernetes/test-infra#20631](https://github.com/kubernetes/test-infra/issues/20631), [kubernetes/test-infra#20707](https://github.com/kubernetes/test-infra/pull/20707)

Ah, thanks for pointing that out. I even went searching for something like that and couldn't find anything.

We won't need this after all then

/close

smg247:
@cjwagner, I am attempting to [use](https://github.com/openshift/release/pull/56596) the `excluded_repos` config, but I am getting a validation error. Maybe I am missing how this is supposed to be used, but I don't see how this [logic](https://github.com/kubernetes-sigs/prow/blob/96e88c5c121cfd7722cbbf01944e5d17b85a9913/pkg/plugins/config.go#L1104-L1118) allows us to exclude a single plugin from an org's configuration for a repo, while including the rest of them.

cjwagner:
From what I can tell the validation is incomplete and needs to be improved to account for excluded_repos. Maybe we just haven't encountered this case before.

smg247:
That is my conclusion as well. I do think that fixing that validation is a better solution than what I have done with this PR, so I will take that approach instead.

PR #275, opened 2024-09-13
fix `validatePluginsDupes` to account for `excluded_repos`

smg247:
Excluded repos are still validated for duplicate plugin configuration against their org. This prevents the possibility to exclude a repo from an org's plugin configuration, and then select some of the plugins that the org has enabled in the repo's configuration.

Discovered this issue from trying to use the `excluded_repos` configuration after abandoning the functionality in https://github.com/kubernetes-sigs/prow/pull/273.

### DIFF
--- a/pkg/plugins/config.go
+++ b/pkg/plugins/config.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"reflect"
 	"regexp"
+	"slices"
 	"sort"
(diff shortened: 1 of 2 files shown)


smg247:
@cjwagner, are you able to take a look at this?

cjwagner (approved):

PR #280, opened 2024-09-20
Fix cache mode determination

netomi:
This PR fixes 2 things:

- cache mode determination in coalesce.go (https://github.com/kubernetes-sigs/prow/blob/main/pkg/ghcache/coalesce.go#L242)
- add additional paths for repository API calls that we use and triggered a warning in ghproxy and asks to raise an issue with the project

in coalescer RoundTrip method, the cache Mode is determined at the end by inspecting the final response object. However, at this point the response object is not the original 304 response from the GitHub server anymore, but the cache response which has usually a status code of 200. So the method `cacheResponseMode` never returns `ModeRevalidated` but `Changed` which is also reflected in the cache.

This does not seem to have another negative effect afaict apart from being confusing and triggered me to inspect whether ghproxy really correctly cached and revalidates requests. Instead of looking at the status code (which is not reliable anymore at this step of the processing), the fix looks at the `X-From-Cache` header that is set by the `httpcache` package in case a cached version is returned.

A test case had to be adapted to take that change into account correctly.

This is my first go PR so please bear with me. With these fixes ghproxy runs pretty nice for our use-case (https://github.com/eclipse-csi/otterdog)

Thanks a a lot for this very helpful piece of software.

### DIFF
--- a/pkg/ghcache/coalesce_test.go
+++ b/pkg/ghcache/coalesce_test.go
@@ -28,6 +28,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/cjwagner/httpcache"
+
 	"sigs.k8s.io/prow/pkg/github/ghmetrics"
 
 	"k8s.io/apimachinery/pkg/util/diff"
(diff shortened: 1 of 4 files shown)


petr-muller:
Wow, thanks for looking into this :+1: 

/ok-to-test

PR #294, opened 2024-10-08
(spyglass/lenses/html) add allow-forms to sandbox

norrs:
This adds allow-forms to the iframe sandbox.

> Blocked form submission to '' because the form's frame is sandboxed
and the 'allow-forms' permission is not set.

This allow's us to click the `<a href=link-to-buildbuddy-invocation target=_blank>` which links to streaming build results on buildbuddy, as we use bazel with remote builders on buildbuddy.io and the buil-logs uploaded to S3 only contains this hyperlink.

With this change we allow chrome to handle the login process at buildbuddy.io which requires a form-post to handle login over SSO.

https://web.dev/articles/sandboxed-iframes

Original lens PR: https://github.com/kubernetes/test-infra/pull/10208

Later changes that are similar to this one:

allow same-origin: https://github.com/kubernetes-sigs/prow/commit/b9a016753cd4bb426a5e809659941fffa1d2ad60
allow popups: https://github.com/kubernetes/test-infra/pull/23069

### DIFF
--- a/pkg/spyglass/lenses/html/template.html
+++ b/pkg/spyglass/lenses/html/template.html
@@ -13,7 +13,7 @@
     {{/* Do _not_ hide this by default, that will break inner javascript that dynamically resizes. Hiding post-render is ok, so we hide on first resize request */}}
     <tr class="initial" id="{{.ID}}-tr">
       <td colspan="2" style="border: 0px; padding: 0px;">
-        <iframe srcdoc="{{.Content}}" title="{{.Filename}}" sandbox="allow-scripts allow-popups allow-same-origin" id="{{.ID}}" width="100%" scrolling="no"></iframe>
+        <iframe srcdoc="{{.Content}}" title="{{.Filename}}" sandbox="allow-scripts allow-popups allow-same-origin allow-forms" id="{{.ID}}" width="100%" scrolling="no"></iframe>
       </td>
     </tr>
     {{end}}
--- a/pkg/spyglass/lenses/html/testdata/TestRenderBody_Simple-0.yaml
+++ b/pkg/spyglass/lenses/html/testdata/TestRenderBody_Simple-0.yaml
@@ -21,7 +21,7 @@ window.addEventListener(&quot;load&quot;, function(){
     var config = { attributes: true, childList: true, characterData: true, subtree:true}; // PT2
     observer.observe(window.document, config);                                            // PT3
 });
-</script>" title="file.html" sandbox="allow-scripts allow-popups allow-same-origin" id="file.html-0" width="100%" scrolling="no"></iframe>
+</script>" title="file.html" sandbox="allow-scripts allow-popups allow-same-origin allow-forms" id="file.html-0" width="100%" scrolling="no"></iframe>
       </td>
     </tr>
     
--- a/pkg/spyglass/lenses/html/testdata/TestRenderBody_With_description-0.yaml
+++ b/pkg/spyglass/lenses/html/testdata/TestRenderBody_With_description-0.yaml
@@ -21,7 +21,7 @@ window.addEventListener(&quot;load&quot;, function(){
     var config = { attributes: true, childList: true, characterData: true, subtree:true}; // PT2
     observer.observe(window.document, config);                                            // PT3
 });
-</script>" title="file.html" sandbox="allow-scripts allow-popups allow-same-origin" id="file.html-0" width="100%" scrolling="no"></iframe>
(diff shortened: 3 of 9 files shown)


petr-muller:
/ok-to-test

I don't know browsers & web apps enough to judge if this has risks, I hope one of the assigned reviewers has a chance to look

michelle192837:
I'm not super familiar with the risks of this either, but I can take a look into it.

michelle192837:
Thanks, that's more detailed than what I was about to post 😅 

While we're discussing it, do you have an example link to Spyglass for what you're trying to enable? (I looked at the linked discussion in buildbuddy but didn't see a Spyglass example there or here).

Relevant as well: the discussion on the original PR for allow-same-origin: https://github.com/kubernetes/test-infra/pull/22921

michelle192837:
My first lean is that if this is already opening a new tab/window and it's clear this is an external link (so go at your own risk), `allow-popups-to-escape-sandbox` is less broad and less subject to misuse than `allow-forms`. That said given we have `allow-scripts` already, I think either shouldn't be much more serious to add.

(I do think separately, we should probably review `allow-scripts` and `allow-same-origin` in Spyglass, but that's outside the scope of this PR).

norrs:
> While we're discussing it, do you have an example link to Spyglass for what you're trying to enable? (I looked at the linked discussion in buildbuddy but didn't see a Spyglass example there or here).

So we are running prow inside our business unit at Cisco, so the only thing I can provide is a screenshot where I have hidden a bit things, but it should give you the gist of my problem in a visual presentation:

image


> 
> Relevant as well: the discussion on the original PR for allow-same-origin: [kubernetes/test-infra#22921](https://github.com/kubernetes/test-infra/pull/22921)

Interesting, it seems to have been discussed before as well. 


> https://github.com/kubernetes/test-infra/pull/22921#pullrequestreview-714948262
> `We control what we put into the iframe so this should be fine. Not a frontend specialist though.`

I have to admit that I would agree with this one, and the threat model should be operator of the prow installation that should be responsible for what it trust and define what to add spyglasses to.

> https://github.com/kubernetes/test-infra/pull/22921#issuecomment-1078440770
> image

and also this statement about mixing the `allow-scripts` with `allow-same-origin`. 

Comes back to : https://github.com/kubernetes/test-infra/pull/10208 and Rendering implementation .
where my mind is 🤷  - I do not know because I do not understand how this remote spy glass and the security model is intended to work. 

So I kinda come around to: is it better to document that the threat model is up to the operator of the prow instance to only spyglass sources it trusts?

> https://github.com/kubernetes-sigs/prow/pull/294#issuecomment-2402959743

But from the twitter example on https://web.dev/articles/sandboxed-iframes : 

image

image

image

When reading this over again, I belive this support my statement about the threat model, and a proper enhancement would be to be able for the operator of the prow installation to define this pr spyglass. This way it can sandbox each spyglass properly according to it needs. But hopefully we could argue that this probably is out-of-scope for this PR? 

So then the question remains, do we simply add `allow-forms` ? But I could live without this PR tho, since I did learn about the "middle-mouse-click" and 🙏 this is not a browser bug 😆 



Another thing about the `allow-popups-to-escape-sandbox` , this is my guess what it might protect us from:  a bad-iframe-source who has a symlink that links to another internal website, and you have managed to escape sandbox and you could maybe inject some javascript as well somehow with some clever XSS or something? Im not a pen-tester or by any means any frontend expert, but it is a hunch if I let my fantasy run wild and try to think about bad things. 

My mind wandered over this thought as well: 
if you embed something on your site, via an iframe, you should kinda trust it. I guess this is why advertisers started running ads via iframes with different domains etc to avoid cookiestealing I believe.

BenTheElder:
I don't think we should allow this by default, if we're going to weaken the sandbox in any way this should be opt-in and prow.k8s.io will not opt-in.

Why not write the log link out to a file that renders the URL directly, instead of inside the streaming logs?

BenTheElder:
> (I do think separately, we should probably review allow-scripts and allow-same-origin in Spyglass, but that's outside the scope of this PR).

agreed

norrs:
> I don't think we should allow this by default, if we're going to weaken the sandbox in any way this should be opt-in and prow.k8s.io will not opt-in.

I agree,  I believe this is achievable with https://github.com/kubernetes-sigs/prow/pull/296/files , and also turns this PR void/not needed. 

> 
> Why not write the log link out to a file that renders the URL directly, instead of inside the streaming logs?

🤔 Good question, my hunch is that this allows it to show several logs from artifacts uploaded in the object bucket without having to update the configuration from prow.

norrs:
Closed, please see  https://github.com/kubernetes-sigs/prow/pull/296#issuecomment-2482774632 .

PR #296, opened 2024-10-10
(spyglass/lenses) allow configuration sandbox permissions

norrs:
This provides the ability to configure iframe sandbox permissions pr lense. This allows the operator of the prow installation to define which permissions it trust to each lense.

PR comes from the ideas and discussions in
https://github.com/kubernetes-sigs/prow/pull/294

### DIFF
--- a/cmd/checkconfig/testdata/combined.yaml
+++ b/cmd/checkconfig/testdata/combined.yaml
@@ -23,6 +23,10 @@ deck:
           - (FAIL|Failure \[)\b
           - panic\b
           - ^E\d{4} \d\d:\d\d:\d\d\.\d\d\d]
+          iframe_sandbox_permissions:
+          - allow-scripts
+          - allow-popups
+          - allow-popups-to-escape-sandbox
       required_files:
       - build-log.txt
     - lens:
--- a/pkg/spyglass/lenses/buildlog/lens.go
+++ b/pkg/spyglass/lenses/buildlog/lens.go
@@ -50,10 +50,11 @@ const (
 var defaultHighlightLineLengthMax = 10000 // Default maximum length of a line worth highlighting
 
 type config struct {
-	HighlightRegexes   []string         `json:"highlight_regexes"`
-	HideRawLog         bool             `json:"hide_raw_log,omitempty"`
-	Highlighter        *highlightConfig `json:"highlighter,omitempty"`
-	HighlightLengthMax *int             `json:"highlight_line_length_max,omitempty"`
+	HighlightRegexes         []string         `json:"highlight_regexes"`
+	HideRawLog               bool             `json:"hide_raw_log,omitempty"`
+	Highlighter              *highlightConfig `json:"highlighter,omitempty"`
+	HighlightLengthMax       *int             `json:"highlight_line_length_max,omitempty"`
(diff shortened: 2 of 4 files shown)


cmd/checkconfig/testdata/combined.yaml
```diff
@@ -23,6 +23,10 @@ deck:
           - (FAIL|Failure \[)\b
           - panic\b
           - ^E\d{4} \d\d:\d\d:\d\d\.\d\d\d]
+          iframe_sandbox_permissions:
+          - allow-scripts
+          - allow-popups
+          - allow-popups-to-escape-sandbox
```
norrs, line 29: Updated this as it looked like a sane idea to do if this is supposed to demostrate a full configuration file example.

pkg/spyglass/lenses/buildlog/lens.go
```diff
@@ -97,6 +99,17 @@ func (lens Lens) Header(artifacts []api.Artifact, resourceDir string, config jso
...
+// defaultSandboxPermissions is the default value for iframe_sandbox_permissions lense config if it is not specified.
+var defaultSandboxPermissions = strings.Join(
+	[]string{
+		"allow-scripts",
+		"allow-top-navigation",
+		"allow-popups",
+		"allow-same-origin",
+	},
+	" ",
+)
```
norrs, line 111: This is defaultSandboxPermissions for the parsedConfig, tho I personally like to work with arrays instead of strings with `foo bar zar`. 

Not sure if you prefer this or the `allow-scripts allow-top-navigation allow-popups allow-same-origin` instead. 

Please let me know, I'll update if you want to change it.

norrs: Suggestion is to do another PR after this to adjust the defaults to not combine `allow-same-origin` together with `allow-scripts`, but I suppose this needs to be communicated to the users of the prow installation, so they are aware of the ability to configure this pr lense.

michelle192837:
Thanks for the ping and apologies for the delay, this dropped off my radar >< 

I'll review today!

pkg/spyglass/lenses/buildlog/lens_test.go
```diff
@@ -61,6 +62,14 @@ func TestGetConfig(t *testing.T) {
 				}
 				return d
 			}(),
+		}, {
```
michelle192837, line 65: nit: Add test case verifying what the iframe permissions are if none are specified.

michelle192837 (commented):
LGTM aside from a minor nit.

pkg/spyglass/lenses/buildlog/lens_test.go
norrs: Added test @michelle192837 , good nit 👍

norrs:
Seems like we need an approval from [cmd/checkconfig/OWNERS](https://github.com/kubernetes-sigs/prow/blob/main/cmd/checkconfig/OWNERS): @chases2 , @stevekuznetsov or @cjwagner

michelle192837:
Those folks are no longer working on Prow unfortunately. I think this should get approval from someone in https://github.com/kubernetes-sigs/prow/blob/main/OWNERS until that OWNERS file is fixed.

(Pulling randomly) @petr-muller  or @matthyx ?

matthyx (approved):
/lgtm

PR #303, opened 2024-10-17
`tide`: allow for optional contexts to be defined via regex

smg247:
OpenShift is developing a new external prow plugin that will create GitHub `check_runs`. These will have a consistent naming scheme, and we would like `tide` to ignore them. This new field will allow for this.

### DIFF
--- a/pkg/config/tide.go
+++ b/pkg/config/tide.go
@@ -840,7 +840,7 @@ func (cp *TideContextPolicy) Validate() error {
 	for _, contextRegex := range cp.OptionalRegexContexts {
 		re, err := regexp.Compile(contextRegex)
 		if err != nil {
-			return fmt.Errorf("optional context regex %q doesn't comipile: %w", contextRegex, err)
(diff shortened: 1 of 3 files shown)


droslean (approved):

pkg/config/tide.go
```diff
@@ -940,6 +959,12 @@ func (cp *TideContextPolicy) IsOptional(c string) bool {
 	if sets.New[string](cp.OptionalContexts...).Has(c) {
 		return true
 	}
+	for _, regexContext := range cp.OptionalRegexContexts {
+		re := regexp.MustCompile(regexContext)
```
petr-muller, line 963: @smg247 can you follow up and change this so that we do not recompile all regexes all the time? ideally they'd be compiled once at startup and we'd store and use that

smg247: Yeah, that's a good idea. I'll take care of that

smg247: https://github.com/kubernetes-sigs/prow/pull/304

PR #304, opened 2024-10-18
`tide`: follow up to optional context regex field

smg247:
modify optional context regex logic to only compile the regexp on load, and fix it so it is properly loaded and merged

This is a follow up to: https://github.com/kubernetes-sigs/prow/pull/303


droslean:
/lgtm
/approve

/hold

Holding for @petr-muller to take a look.

petr-muller (approved):
Thanks!

/hold cancel

PR #317, opened 2024-11-05
`deck`: move readiness serve to where we actually start the server

smg247:
This is a follow up to the failed attempt from https://github.com/kubernetes-sigs/prow/pull/308. It makes it more clear that the CSRF logic wraps the server, and starts the readiness endpoint immediately before starting the server.

### DIFF
--- a/cmd/deck/main.go
+++ b/cmd/deck/main.go
@@ -460,9 +460,6 @@ func main() {
 		mux = prodOnlyMain(cfg, pluginAgent, authCfgGetter, githubClient, o, mux)
 	}
 
-	// signal to the world that we're ready
-	health.ServeReady()
-
 	// cookie secret will be used for CSRF protection and should be exactly 32 bytes
 	// we sometimes accept different lengths to stay backwards compatible
 	var csrfToken []byte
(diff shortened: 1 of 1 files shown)


cmd/deck/main.go
```diff
@@ -500,13 +497,15 @@ func main() {
 		return
 	}
 
+	var server *http.Server
 	if csrfToken != nil {
 		CSRF := csrf.Protect(csrfToken, csrf.Path("/"), csrf.Secure(!o.allowInsecure))
-		logrus.WithError(http.ListenAndServe(":8080", CSRF(traceHandler(mux)))).Fatal("ListenAndServe returned.")
```
petr-muller, line 505: lol sneaky

petr-muller (approved):

petr-muller:
/hold

Can we remove the `--allow-insecure` from the tests again? Or at least from one, to integeration-test both paths?

smg247:
> Can we remove the `--allow-insecure` from the tests again? Or at least from one, to integeration-test both paths?

Since the other PR was reverted, this has required no changes to the integration tests at all. It is not present in the `deck_tenant_deployment` now.

cmd/deck/main.go
smg247: Really, it's like it was written to deceive

petr-muller:
> Since the other PR was reverted, this has required no changes to the integration tests at all. It is not present in the deck_tenant_deployment now.

Ugh, I forgot that because we reverted the test manifests are now in the state I wanted. Code reviews on Friday :grimacing:

PR #330, opened 2024-11-23
Export providers.S3Credentials

midnightconman:
I would like to export this struct for use in another package, any objections?


petr-muller (approved):

midnightconman:
Thank you! 😸

Issue #334, opened 2024-12-02
pull-prow-integration is broken


BenTheElder:
from: https://kubernetes.slack.com/archives/C09QZ4DQB/p1733130987871239

https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-prow-integration

/assign

PR #335, opened 2024-12-02
install kind for integration tests

BenTheElder:
this is pretty cheap/fast, kind is a very small go binary, and upgrades can now be tested with PRs to this repo

fixes https://github.com/kubernetes-sigs/prow/issues/334

### DIFF
--- a/hack/tools/ensure-kind.sh
+++ b/hack/tools/ensure-kind.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# Copyright 2024 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


hack/tools/ensure-kind.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# usage: source this script to ensure kind is setup
+
+# ensure kind is available for setup-kind-cluster.sh etc
+# TODO: consider factoring out the kind version used for CI or
+# else ensure all CI jobs use this script
+# TODO: consider enforcing a particular version locally as well to avoid confusion
+if ! command -v kind; then
+    local repo_gobin="${REPO_ROOT}/_bin"
+    GOBIN="${repo_gobin}" go install sigs.k8s.io/kind@v0.25.0
```
BenTheElder, line 24: note: I didn't put this in the tools module because there's no reason to align the dependencies.

I thought about fetching a binary instead but matching the host platform is : effort : and we know we have go in this context

BenTheElder:
ah, there's a verify check looking for executable bit.

the new script is only meant to be sourced, not directly executed (because it modifies PATH env) but we can make it +x to satisfy the lint :-)

BenTheElder:
(added executable bit to setup-kind.sh)

petr-muller (approved):
Thanks!

PR #338, opened 2024-12-03
`deck`: "/log" endpoint should not show `hidden` jobs if deck isn't configured to allow `hidden`

smg247:
In OpenShift's prow installation, we give many jobs a tenant id of `gangway` in order to facilitate triggering them via gangway. Due to this, we must include the `gangway` tenant id in our `deck` instance in order for these jobs to show.

This results in deck listing the `hidden` jobs because of this logic: https://github.com/kubernetes-sigs/prow/blob/db89760fea406dd2813e331c3d52b53b5bcbd140/pkg/deck/jobs/jobs.go#L145-L149
That isn't so bad, and does not have a simple solution.

However, we recently discovered that the `/log` endpoint (which is accessible via a button beside the listing) actually shows the logs. This PR fixes that part so we don't leak sensitive data.

Note: I tested this out by using a locally built image, and it functions as desired in the standard, and "hidden" `deck` instances.

### DIFF
--- a/pkg/deck/jobs/jobs.go
+++ b/pkg/deck/jobs/jobs.go
@@ -73,7 +73,7 @@ type Job struct {
 }
 
 type serviceClusterClient interface {
-	ListProwJobs(selector string) ([]prowapi.ProwJob, error)
+	ListProwJobs(selector string, hiddenRepos func() sets.Set[string]) ([]prowapi.ProwJob, error)
 }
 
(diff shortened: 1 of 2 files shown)


smg247:
Most of the changes here are just moving stuff around so we can use the `pjHasHiddenRefs` function in the `JobsAgent` itself.

Issue #309, opened 2024-10-25
Prow integration test is flaky


smg247:
[pull-prow-integration](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-prow-integration) fails frequently with the following flake:
```
=== FAIL: test/integration/test TestDeletePod/running-pod (60.04s)
    sinker_test.go:246: Creating client for cluster: kind-kind-prow-integration
    sinker_test.go:263: Creating prowjob: running-pod-f7470c65da881743dda945debe3495a7
    sinker_test.go:267: Finished creating prowjob: running-pod-f7470c65da881743dda945debe3495a7
    sinker_test.go:271: Creating pod: running-pod-f7470c65da881743dda945debe3495a7
    sinker_test.go:275: Finished creating pod: running-pod-f7470c65da881743dda945debe3495a7
    sinker_test.go:286: Wait for sinker deleting pod or timeout in 1 minute: running-pod-f7470c65da881743dda945debe3495a7
    sinker_test.go:306: client rate limiter Wait returned an error: rate: Wait(n=1) would exceed context deadline
=== FAIL: test/integration/test TestDeletePod (0.00s)
```

Example failing run: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_prow/308/pull-prow-integration/1853470929506537472

petr-muller:
Interesting, this looks new :thinking:

smg247:
The original description was a real failure, but it is flaky. The actual flake is now in the description.

VannTen:
Not that the original error (in the edit history) is till present in the build log :
` error: .status.conditions accessor error: <nil> is of the type <nil>, expected []interface{} ` https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_prow/346/pull-prow-integration/#1:1868943776085446656/build-log.txt%3A698

smg247:
I believe this is caused by https://github.com/kubernetes-sigs/prow/pull/262/commits/6c290be6d3b18042be29e7285372f98a70a40026.

@xrstf, do you happen to remember why you needed to do this?

What I am seeing is that this now frequently flakes with the following error:
```
client rate limiter Wait returned an error: rate: Wait(n=1) would exceed context deadline
```
[...]

PR #346, opened 2024-12-17
Respect newline in command examples in command-help

VannTen:
Currently, the command-help page renders multiline command examples (the
only example I've seen is for /release-note-edit) on one line, which is
confusing for users.

Change the style to respect newline in the source.

Link: https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

Before: 
image

After:
image


VannTen:
Failing test is an instance of #309

PR #349, opened 2024-12-18
Add option for blunderbuss to wait for tide status 

dobesv:
As a code reviewer, I don't want to review code which has not passed the test suites yet.  

If I am asked to review something and then the tests fail after my review, it feels like I wasted my time as the person submitting it has to address the feedback from the tools anyway.  However, if I do not review it immediately then I have to manually check back later to see whether the tests have passed - there is no way currently to set up a notification for tests passing under the current system.

This change allows configuring blunderbuss to request reviews when a matching commit status is reported instead of immediately when the pull request is opened or marked as ready for review.  By setting the condition to one which matches the status that `tide` posts when all tests have passed but it is waiting for a label or for GitHub reviews to be posted, I can defer the automated review requests until the tests have passed.

This way someone can open a pull request, and when they eventually get all the tests passing, I can be prompted to review the pull request.  I won't waste time reviewing a pull request that has failing tests or lint errors, and I also will not have to manually "check back" on all the pull requests to see which ones have the tests passing.

For example in this project, you could in theory set up blunderbuss to wait with requesting reviews until tide shows this status:

image


dobesv:
I probably would need some guidance how to resolve that failure in `pull-prow-integration`, I'm having trouble relating that failure to my change here.

pkg/plugins/config.go
```diff
@@ -172,6 +184,9 @@ type Blunderbuss struct {
 	// This is useful when a bot user or admin opens a PR that will be
 	// merged regardless of approvals.
 	IgnoreAuthors []string `json:"ignore_authors,omitempty"`
+	// WaitForTideStatus specifies whether to request reviews if the tide status indicates that
```
petr-muller, line 187:
```suggestion
	// WaitForStatus specifies whether to request reviews if the tide status indicates that
```

petr-muller (commented):
This looks nice, I like how the expensive GH operations are only performed on status events after all checks are done and we are certain we found the context we want to act on.

Thanks!

petr-muller (approved):
I actually wanted to approve this :grimacing: 
/meow

PR #350, opened 2024-12-20
Upgrade golang v1.23

danilo-gemoli:
Bump Golang `v1.23` in:
- `go.mod`
- `hack/tools/go.mod`
- `site/go.mod`

`golangci-lint` has been upgraded as well to `v1.62.2`, but this time there is more than fixing the lint errors that usually show up when doing such maintenance work:

The `ast.Package` has been deprecated, therefore I fixed the following in `pkg/genyaml/genyaml.go`:
```diff
import "go/doc"

- astDoc := doc.New(apkg, "", 0)
+ astDoc, err := doc.NewFromFiles(fset, astFiles, importPath)
```

and I have changed some `func` signatures as well in order to pass `importPath` down the stack, and also introduce a callback:
```golang
// ImportPathResolver resolves an import path out of physical directory path. Example:
//
//	/usr/src/prow/pkg/genyaml -> sigs.k8s.io/prow/pkg/genyaml
type ImportPathResolver func(dir string) (string, error)
```
to let the caller decide how to properly resolve a package name out of a physical directory.

As far as I was able to observer, there is no change in behavior, so we should be safe hopefully.


site/go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/kubernetes-sigs/prow
 
-go 1.20
+go 1.23.4
```
matthyx, line 3: 10:08:22 AM: hugo --enableGitInfo --buildFuture -b https://deploy-preview-350--k8s-prow.netlify.app/
10:08:22 AM: go: errors parsing go.mod:
10:08:22 AM: /opt/build/repo/site/go.mod:3: invalid go version '1.23.4': must match format 1.23
10:08:22 AM: Error: failed to download modules: failed to execute 'go [mod download]': failed to execute binary "go" with args [mod download]: go: errors parsing go.mod:
10:08:22 AM: /opt/build/repo/site/go.mod:3: invalid go version '1.23.4': must match format 1.23
10:08:22 AM:  *errors.errorString

matthyx:
thanks @danilo-gemoli can you check the deploy steps?

danilo-gemoli:
Sure, on it

matthyx (approved):
looking good, I'm just letting others chime in...
/lgtm
/hold

matthyx:
I think we can go
/unhold

PR #354, opened 2025-01-07
refactor(pkg/jenkins): change the state update logic

wuhuizuo:
This can improved Jenkins type CI status display for better user experience.

- Keep the prow job as `triggered` when the Jenkins build is still in queue.
- Only set the prow job to `pending` state when the Jenkins build is running.

## Previous Situation:
When a CI job was triggered, the status would show "Pending - Jenkins job enqueued" for running jobs, which could be misleading as it suggested the job hadn't started yet, and there are no links to view during the run.
image

## This refactoring effect:
Running CI jobs now display "Pending - Jenkins job running" with detail url provided.
image

### DIFF
--- a/pkg/jenkins/controller.go
+++ b/pkg/jenkins/controller.go
@@ -448,27 +448,28 @@ func (c *Controller) syncTriggeredJob(pj prowapi.ProwJob, reports chan<- prowapi
 		} else {
 			pj.Status.Description = "Jenkins job enqueued."
 		}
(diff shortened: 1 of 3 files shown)


jmguzik (approved):
I think it's reasonable change.

jmguzik:
Probably approver will ask you to squash commits so you may want to do that.

petr-muller (approved):

PR #358, opened 2025-01-14
Config to automatically Re-trigger failed periodics

jmguzik:
This PR introduces a new configuration field (along with the functionality in horologium component) for ProwJobs to support automatically re-triggering periodic jobs based on a specified policy. The new `retrigger-failed-run` field allows retrying a failed periodic job with customizable settings for the number of attempts and the interval between retries.

Example usage:
```
retrigger-failed-run:
  until_success: true
  attempts: 3
  interval: 6h
```

`until_success: true` alows to stop if success state is reached. Set to `false` it triggers until `attempts` reached.
sloves https://github.com/kubernetes-sigs/prow/issues/268

### DIFF
--- a/cmd/horologium/main.go
+++ b/cmd/horologium/main.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -30,12 +31,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
 
 	prowapi "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
+	v1 "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	"sigs.k8s.io/prow/pkg/config"
 	"sigs.k8s.io/prow/pkg/cron"
 	pkgflagutil "sigs.k8s.io/prow/pkg/flagutil"
 	prowflagutil "sigs.k8s.io/prow/pkg/flagutil"
 	configflagutil "sigs.k8s.io/prow/pkg/flagutil/config"
 	"sigs.k8s.io/prow/pkg/interrupts"
+	"sigs.k8s.io/prow/pkg/kube"
 	"sigs.k8s.io/prow/pkg/logrusutil"
 	"sigs.k8s.io/prow/pkg/metrics"
 	"sigs.k8s.io/prow/pkg/pjutil"
@@ -44,6 +47,7 @@ import (
 
 const (
 	defaultTickInterval = time.Minute
+	maxRetries          = 10
 )
 
 type options struct {
@@ -206,8 +210,17 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
 				"job":            p.JobBase.Name,
 			}).Debug("Trigger time has not yet been reached.")
 		}
-		if !previousFound || shouldTrigger {
-			prowJob := pjutil.NewProwJob(pjutil.PeriodicSpec(p), p.Labels, p.Annotations,
+
+		var labels map[string]string
+		if p.Labels != nil {
+			labels = make(map[string]string)
+			for k, v := range p.Labels {
+				labels[k] = v
+			}
+		}
+
+		if !previousFound || shouldTrigger || shouldTriggerFailedRun(j, p, now, logger, &labels) {
+			prowJob := pjutil.NewProwJob(pjutil.PeriodicSpec(p), labels, p.Annotations,
 				pjutil.RequireScheduling(cfg.Scheduler.Enabled))
 			prowJob.Namespace = cfg.ProwJobNamespace
 			logger.WithFields(logrus.Fields{
@@ -226,4 +239,55 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
 		return fmt.Errorf("failed to create %d prowjobs: %v", len(errs), errs)
 	}
 	return nil
+}
+
+func shouldTriggerFailedRun(j v1.ProwJob, p config.Periodic, now time.Time, logger *logrus.Entry, labels *map[string]string) bool {
+	if p.Retry == nil {
+		return false
+	}
+	if !j.Complete() {
+		return false
+	}
+	runCount := 1
+
+	countLabel, exists := j.Labels[kube.ReRunLabel]
+	if exists {
+		count, err := strconv.Atoi(countLabel)
+		if err != nil {
+			logger.WithFields(logrus.Fields{}).WithFields(
+				pjutil.ProwJobFields(&j),
+			).Warn("Failed to convert label value.")
+			return false
+		}
+		runCount = count + 1
+	}
+
+	maxForJob := p.Retry.Attempts
+	if maxForJob > maxRetries {
+		maxForJob = maxRetries
+	}
+	if runCount > maxForJob {
+		return false
+	}
+
+	lastRunTime := j.Status.StartTime.Time
+	if now.Sub(lastRunTime) <= p.Retry.GetInterval() {
+		return false
+	}
+
+	if !p.Retry.RunAll && j.Status.State == v1.SuccessState {
+		return false
+	}
+
+	if *labels == nil {
+		*labels = make(map[string]string)
+	}
+	(*labels)[kube.ReRunLabel] = strconv.Itoa(runCount)
+
+	logger.WithFields(logrus.Fields{
+		"attempt": strconv.Itoa(runCount),
+		"run_all": p.Retry.RunAll,
+	}).Debug("Job marked to be retried")
+
+	return true
 }
--- a/cmd/horologium/main_test.go
+++ b/cmd/horologium/main_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"reflect"
+	"strconv"
 	"testing"
 	"time"
 
@@ -30,9 +31,11 @@ import (
 	fakectrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	prowapi "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
+	v1 "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	"sigs.k8s.io/prow/pkg/config"
 	"sigs.k8s.io/prow/pkg/flagutil"
 	configflagutil "sigs.k8s.io/prow/pkg/flagutil/config"
+	"sigs.k8s.io/prow/pkg/kube"
 )
 
 type fakeCron struct {
@@ -60,9 +63,14 @@ func TestSync(t *testing.T) {
 	testcases := []struct {
 		testName string
 
-		jobName         string
-		jobComplete     bool
-		jobStartTimeAgo time.Duration
+		jobName                  string
+		jobComplete              bool
+		jobStartTimeAgo          time.Duration
+		retry                    *config.Retry
+		state                    prowapi.ProwJobState
+		interval                 time.Duration
+		retryLabelNumber         int
+		expectedRetryLabelNumber int
 
 		shouldStart bool
 	}{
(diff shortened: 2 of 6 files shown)


cmd/horologium/main.go
```diff
@@ -199,14 +202,27 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
 			}).Debug("Trigger time has not yet been reached.")
 		}
+		run := false
+
+		// Proceed only if the run deviates from the normal flow.
+		if !shouldTrigger {
+			run = shouldTriggerFailedRun(j, p, now, logger, previousFound)
+		}
 		if !previousFound || shouldTrigger {
+			run = true
```
Prucek, line 219: can't this be simplified into this:
```suggestion
		if !previousFound || shouldTrigger || shouldTriggerFailedRun(j, p, now, logger, previousFound){
```

petr-muller:
/hold

Please provide a PR description of what the PR actually implements - the link to the issue is helpful but not sufficient, because even the issue contains some discussion about options for a vague idea, not a clear implementation direction.

jmguzik:
@petr-muller done

jmguzik:
/hold cancel
then?

petr-muller:
I'll review tomorrow :pray:

cmd/horologium/main.go
```diff
@@ -227,3 +243,48 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
+		if !j.Complete() {
+			return false
+		}
+		countLabel, exists := j.Labels[kube.ReRunLabel]
+		if exists {
+			if count, err := strconv.Atoi(countLabel); err == nil {
+				runCount = count + 1
+			}
+		}
+	}
```
petr-muller, line 263: I do not understand what does the logic try to achieve when `previousFound==false`. In that case, previous logic resulted in a simple trigger so that you could establish the baseline for interval-based triggers. No "rerun" is necessary in that context, that sounds like something that only has defined meaning value when there _is_ a previous job?

I'd expect the high-level to be something like the following:

1. If there's no prowjob then simply trigger one, it is not a retry
2. If previous job is a failure then
  a. If it was not a retry, spawn a retry (and label it to recognize its position in the series)
  b. If it was a retry per its label, spawn a retry if there's a retry budget per the job config
3. If previous job is a success _and_ its a retry _and_ all-retries is set _and_ there's a retry budget per job config then spawn a retry
4. Otherwise, run the next normally triggered job (not a retry)

cmd/horologium/main.go
```diff
@@ -199,14 +202,27 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
+		if !shouldTrigger {
+			run = shouldTriggerFailedRun(j, p, now, logger, previousFound)
+		}
 		if !previousFound || shouldTrigger {
+			run = true
+		}
+		if run {
+			if p.RetriggerFailedRun != nil && p.Labels == nil {
+				p.Labels = make(map[string]string)
+				p.Labels[kube.ReRunLabel] = strconv.Itoa(1)
```
petr-muller, line 224: L224 looks like something that needs to be done even when `p.Labels` was not nil

petr-muller (commented):
One thing we need to be intentional about is the interaction between the next legit trigger and a retry, when both could occur, and also whether it is acceptable for retries to "push back" interval triggers.

As done here, next legit trigger takes precedence, basically interrupting the retry sequences, which may impact especially the `stop on success = false` case where we expect to deterministically get the configured amount of retries but we wont because the legit-triggerred job interrupts that.

That sounds acceptable to me but needs to be intentional, documented behavior.

pkg/config/jobs.go
```diff
@@ -266,6 +266,17 @@ type Postsubmit struct {
...
 }
 
+// RetriggerFailedRun defines the configuration for retrying failed prowjobs.
+type RetriggerFailedRun struct {
+	// UntilSuccess stops retries once a successful run is achieved.
+	// The Attempts field still limits the total number of retries.
+	UntilSuccess bool
+	// Attempts specifies the maximum number of retry attempts allowed.
+	Attempts int
+	// Interval defines the wait duration between consecutive retry attempts.
```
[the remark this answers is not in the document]
petr-muller: Number of attempts also sounds like something that should be bounded somehow. Maybe a horologium config that would limit this, no matter what the job configured?

pkg/kube/prowjob.go
```diff
@@ -69,6 +69,9 @@ const (
 	// IsOptionalLabel is added in resources created by prow and
 	// carries the Optional from a Presubmit job.
 	IsOptionalLabel = "prow.k8s.io/is-optional"
+	// ReRunLabel is added in periodics that are configured to be
+	// re-runned several times
```
petr-muller, line 73: "configured to be-rerunned" does not seem to be correct. IIUC the jobs which _are retries_ will have this label?

Ideally also document the expected value (position in the series of retries) or something like this.

pkg/config/jobs.go
```diff
@@ -266,6 +266,17 @@ type Postsubmit struct {
 	JenkinsSpec *JenkinsSpec `json:"jenkins_spec,omitempty"`
 }
 
+// RetriggerFailedRun defines the configuration for retrying failed prowjobs.
+type RetriggerFailedRun struct {
+	// UntilSuccess stops retries once a successful run is achieved.
+	// The Attempts field still limits the total number of retries.
+	UntilSuccess bool
+	// Attempts specifies the maximum number of retry attempts allowed.
+	Attempts int
```
petr-muller, line 275: Number of attempts also sounds like something that should be bounded somehow. Maybe a horologium config that would limit this, no matter what the job configured?

cmd/horologium/main.go
jmguzik: In this code, label is given to any job that is configured to be re-runned. So if there is no `previousFound`,  it will also receive label, but by default with number 1. So, `not a retry` is always marked 1 in the case of this implementation (for configured jobs ofc )and if you take this under consideration and analyze your 4 points, I think they are all met. 

I agree that this should be documented in some form, that if you configure these new fields for prowjob, they will always receive this label.

pkg/config/jobs.go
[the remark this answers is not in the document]
jmguzik: We could, but do we really want to complicate that with additional master config field?

jmguzik:
@petr-muller thanks for the review. I agree with most of your comments but I will clear them when we reach common understanding.

As noted in the comments, every job marked with the new fields (in the configuration) will have associated label. First run will have value 1. This is because I wanted to keep other periodics unaffected and I make distinction on this field weather to process periodic as normal or repeatable. 
> One thing we need to be intentional about is the interaction between the next legit trigger and a retry, when both could occur, and also whether it is acceptable for retries to "push back" interval triggers.

I think the way it is implemented, normal trigger (in the case of this implementation first run value 1) will always have precedence because as soon as `shouldTrigger` is detected to be true, the code will not follow new section.

> As done here, next legit trigger takes precedence, basically interrupting the retry sequences, which may impact especially the stop on success = false case where we expect to deterministically get the configured amount of retries but we wont because the legit-triggerred job interrupts that.

That's true

>That sounds acceptable to me but needs to be intentional, documented behavior.

Makes sense.

pkg/config/jobs.go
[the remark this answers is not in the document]
petr-muller: I guess not, at least not right now, but then let's hardcode a bound into horologium, say, 10?

cmd/horologium/main.go
petr-muller: OK, I get it now. I got confused because I thought that from the two vanilla triggered cases (no previous & time-trigger), only the 'no previous' was somehow handled in a special way. But the time-triggered jobs will actually get the default label value `1` on L224 - which would be applicable even for the no-previous case and this default logic would not need to be present in two places. I'd suggest to do something like this at the callsite

```
// shouldTrigger and previousFound already set correctly

var retryLabelValue int // zero
if p.RetriggerFailed != nil && p.RetriggerFailed.Attempts > 0 {
  retryLabelValue = 1 // Set default for vanilla triggers when job has retries configured
  if  !shouldTrigger && previousFound { // only retrigger when no vanilla trigger is due
    shouldRetriggerFailed, retryLabelValue = shouldTriggerFailedRun(j, p, now, logger)
  }
}
```

then you could condition the run as 

```
if shouldTrigger || shouldRetriggerFailed || !previousFound
```

and later just
```
if retryLabelValue > 0 {
  p.Labels[retryLabel] = Itoa(retryLabelValue)
}
```

That way the `shouldTriggerFailedRun` logic is really simple and deals just with "is the previous run something that needs to be retriggered?", and the top-level clearly communicates different reasons why something got triggered, and the new reason can be easily included as a label for "Triggering new run." log.

cmd/horologium/main.go
```diff
@@ -227,3 +243,48 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
+			return false
+		}
+		countLabel, exists := j.Labels[kube.ReRunLabel]
+		if exists {
+			if count, err := strconv.Atoi(countLabel); err == nil {
+				runCount = count + 1
+			}
+		}
+	}
+	if runCount > p.RetriggerFailedRun.Attempts {
```
petr-muller, line 264: So this means given a vanilla-triggered job has a label with `1`, then when I configure `attempts: 1`, I **do not** actually get a retry behavior? That is a bit confusing to me, I think I would expect to get a single retry after failure for such config.

cmd/horologium/main.go
petr-muller: Actually, one thing I noticed - `p.Labels` is the config structure. Let's not modify the config structure, but create  a new map from it (=populate defaults from the config) and add the controller's labels in there.

cmd/horologium/main.go
petr-muller: I am not super-strong on exactly ^^^ but both vanilla trigger paths should not be handled separately (both just get the default value) and the retry decision logic should be doing just the one thing.

cmd/horologium/main.go
jmguzik: You are right. I'll start counting from 0.

pkg/config/jobs.go
```diff
@@ -282,6 +295,8 @@ type Periodic struct {
 	// Tags for config entries
 	Tags []string `json:"tags,omitempty"`
 
+	RetriggerFailed *RetriggerFailed `json:"retrigger_failed_run,omitempty"`
```
droslean, line 298: nit: perhaps something like `retry_on_failure` will be a better name

pkg/config/jobs.go
```diff
@@ -266,6 +266,19 @@ type Postsubmit struct {
 	JenkinsSpec *JenkinsSpec `json:"jenkins_spec,omitempty"`
 }
 
+// RetriggerFailed defines the configuration for retrying failed prowjobs.
+type RetriggerFailed struct {
+	// UntilSuccess stops retries once a successful run is achieved.
+	// The Attempts field still limits the total number of retries.
+	RunAll bool `json:"run_all,omitempty"`
```
droslean, line 273: I can't think of a case where the user wants to initiate all retries and keep the last try's results. How about keeping this behavior as the default? Will it also make sense to make it configurable?

pkg/config/jobs.go
```diff
@@ -266,6 +266,19 @@ type Postsubmit struct {
...
+	// Attempts specifies the maximum number of retry attempts allowed.
+	Attempts int `json:"attempts,omitempty"`
+	// Interval defines the wait duration between consecutive retry attempts.
+	Interval string `json:"interval,omitempty"`
```
droslean, line 277: Interval is good but we could add a `backoff` configuration as well. I am not sure that the interval time will always be accurate between the retries. Shall we also consider how horologium reconciles the prowjobs?

droslean (commented):
lgtm in overall. I left some nits for @jmguzik to consider

pkg/config/jobs.go
jmguzik: I would like to keep it as it was already discussed during previous review and changed upon request. Also it's closer to what was requested in the issue.

pkg/config/jobs.go
jmguzik: Requestor wanted to have 2 strategies. Default one is until_success so if you achieve success next retry won't happen. Of course it's respecting the attempts field so it won't go over them. The second requested strategy was run_all. If this bool is true, no matter what will be the result, it will run defined number of times. This is to have more sophisticated statistical analisys of the case.

pkg/config/jobs.go
jmguzik: I do not think backoff is needed in this case. Let's take somebody who has test that runs 1h10m but defined 1h interval. In the current implementation, this periodic will be just scheduled right after previous one stops. 
The only case that this not happen is when we have repetition of "big interval" and the next queue is about to start, but the old one is not finished. So:

1st run 2nd run ... should be 3rd but is taking so long so instead 1st run occurs.

I think that does not require changes in implementation but a proper comment explanation to document the chosen approach. If that would be the only nit, I will update comment as a followup.

jmguzik: specifically speaking about second part of https://github.com/kubernetes-sigs/prow/pull/358#pullrequestreview-2559293998

jmguzik: documented behavior not happened and my proposal is to do it as followup (if that ends to be the only ask)

cmd/horologium/main.go
petr-muller: /hold 

I do not think this was addressed?

cmd/horologium/main.go
```diff
@@ -199,15 +203,23 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
 			}
 			continue
 		}
-		if !shouldTrigger {
+		if !shouldTrigger && p.RetriggerFailed == nil {
```
petr-muller, line 206: This will remove the logging even when no retry is in order (last execution was a success and we wait for next vanilla trigger)

cmd/horologium/main.go
```diff
@@ -199,15 +203,23 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
 				"previous-found": previousFound,
 				"name":           p.Name,
 				"job":            p.JobBase.Name,
 			}).Debug("Trigger time has not yet been reached.")
 		}
-		if !previousFound || shouldTrigger {
-			prowJob := pjutil.NewProwJob(pjutil.PeriodicSpec(p), p.Labels, p.Annotations,
+
+		if !previousFound || shouldTrigger || shouldTriggerFailedRun(j, p, now, logger, previousFound) {
+			labels := p.Labels
```
petr-muller, line 215: Well this still modifies the config, `labels` is not a copy, just a reference to the original map.

cmd/horologium/main.go
```diff
@@ -199,15 +203,23 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
+			if p.RetriggerFailed != nil {
+				if labels == nil {
+					labels = make(map[string]string)
+				}
+				labels[kube.ReRunLabel] = strconv.Itoa(0)
```
petr-muller, line 220: Here we reset the label always to the default of 0

cmd/horologium/main_test.go
```diff
@@ -60,9 +63,13 @@ func TestSync(t *testing.T) {
...
-		jobName         string
-		jobComplete     bool
-		jobStartTimeAgo time.Duration
+		jobName          string
+		jobComplete      bool
+		jobStartTimeAgo  time.Duration
+		retriggerFailed  *config.RetriggerFailed
+		state            prowapi.ProwJobState
+		interval         time.Duration
+		retryLabelNumber int
```
petr-muller, line 72: Can we add an expected label number too and check it?

cmd/horologium/main.go
```diff
@@ -227,3 +239,54 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
+			if count, err := strconv.Atoi(countLabel); err == nil {
+				runCount = count + 1
+			}
+		}
+	}
+	attempts := p.RetriggerFailed.Attempts
+	if attempts > maxRetryCount {
+		attempts = maxRetryCount
+	}
+	if runCount >= attempts {
```
petr-muller, line 263: This does not seem correct. 

A vanilla-triggered job will have a label of `0`. So `runCount` will be `1` when we consider to retry it (L255). So when I configure `Attempts = 1`, I will expect a single retry but I will get none?

cmd/horologium/main.go
```diff
@@ -227,3 +239,54 @@ func sync(prowJobClient ctrlruntimeclient.Client, cfg *config.Config, cr cronCli
...
+	}
+
+	if !j.Complete() {
+		return false
+	}
+
+	if p.Labels == nil {
+		p.Labels = make(map[string]string)
+	}
+	p.Labels[kube.ReRunLabel] = strconv.Itoa(runCount)
```
petr-muller, line 284: Let's not modify the config even here

pkg/config/jobs.go
petr-muller, line 298:
```suggestion
	RetriggerFailed *RetriggerFailed `json:"retrigger_failed,omitempty"`
```

For consistency and to avoid verbose configuration 

TBH, I would actually prefer to call this a `Retry` instead of `Retrigger` (pretty much everywhere where you use `Retrigger`) The word "retry" even implies the `Failed` and it would read rather nicely in the config:

```
retry:
  attempts: 3
  run_all: true
```

pkg/kube/prowjob.go
```diff
@@ -69,6 +69,9 @@ const (
 	// IsOptionalLabel is added in resources created by prow and
 	// carries the Optional from a Presubmit job.
 	IsOptionalLabel = "prow.k8s.io/is-optional"
+	// ReRunLabel is added in periodics that are configured to be
+	// re-runned several times, value starts from 0 and it's increased on re-run
+	ReRunLabel = "prow.k8s.io/re-run"
```
petr-muller, line 74: This is more a note on the semantics of the label than something that needs to be addressed, but I am a bit twitchy about some inconsistencies of how this label is treated:

1. Even jobs _not_ labelled can be retried, because the decision is driven by the job configuration, not the label on the last execution. If my last execution is a failed job executed before the `retrigger_failed` stanza was added to the config (= so it has no retry label), it will still start to be retried after the config is added. The presence of the label plays no role in the decision, if its absent, it is assumed to be zero.
2. Jobs with `re-run: 0` label will only be retried if they failed
3. Jobs with `re-run: >0` can be retried if they failed or passed (depending on the `run_all` config)

I would personally prefer if we labelled just the actual retries, but I am not too strong on it (I am mostly interested in whether labeling the non-retries has some intended use).

cmd/horologium/main.go
jmguzik: The intention was not to use this comment which is a debug level if the feature is enabled but use for old jobs. Not sure what to do with it to avoid complicated if.

petr-muller:
We have looked at the PR together, proposed a little naming change + getting rid of the processing of the case where previous job does not exist in `retriggerFailedRun`, then this one is good to go.

jmguzik:
I think all comments are now addressed :)

petr-muller (approved):
/lgtm
/approve
/ship-it
/meow

Thanks!

PR #367, opened 2025-01-31
fix(status-reconciler): correctly run run_if_changed presubmits

### LINKED ISSUE #314: Newly added ProwJobs don't honor `skip_if_changed`, etc.. configurations
When a new check is added to prow, it is added to (and triggered on) all currently opened PRs in the repo regardless of the `skip_if_changed`, and `only_run_if_changed` configs. This causes massive disruption in some cases where there are many opened PRs in the repo. When a new PR is opened after the ProwJob has been added, the configurations are honored.

hown3d:
Fixes #314

### DIFF
--- a/pkg/statusreconciler/controller.go
+++ b/pkg/statusreconciler/controller.go
@@ -245,7 +245,7 @@ func (c *Controller) triggerNewPresubmits(addedPresubmits map[string][]config.Pr
 			// added. we know all of the presubmits we are filtering need to be forced to run, so we can
 			// enforce that with a custom filter
 			filter := pjutil.NewArbitraryFilter(func(p config.Presubmit) (shouldRun bool, forcedToRun bool, defaultBehavior bool) {
-				return true, false, true
+				return true, false, false
 			}, "inline-filter")
 			org, repo, number, branch := pr.Base.Repo.Owner.Login, pr.Base.Repo.Name, pr.Number, pr.Base.Ref
 			changes := config.NewGitHubDeferredChangedFilesProvider(c.githubClient, org, repo, number)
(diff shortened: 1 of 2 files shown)


petr-muller:
/cc

I'll need to take a better look at this. I am a bit suspicious that while this indeed probably fixes #314 I think this actually makes a tradeoff where status-reconciler will not trigger jobs that should be triggered? As in "we can't trigger wrong jobs if we trigger no jobs?".

Not triggering new, potentially blocking, jobs is also a failure case with non-trivial impact: when a job is added that is required for a PR to merge, and status-reconciler does not trigger it, nothing ever will and the PR is stuck waiting for the required result. Unless a human notices and triggers them manually (which is one of the cases status-reconciler is supposed to solve).

petr-muller (approved):
Looked more into the code, and the change does exactly what we need (my mental model of status-reconciler was partially wrong).

Perfect, thanks a lot!

pkg/statusreconciler/controller.go
```diff
@@ -245,7 +245,7 @@ func (c *Controller) triggerNewPresubmits(addedPresubmits map[string][]config.Pr
 			// added. we know all of the presubmits we are filtering need to be forced to run, so we can
 			// enforce that with a custom filter
 			filter := pjutil.NewArbitraryFilter(func(p config.Presubmit) (shouldRun bool, forcedToRun bool, defaultBehavior bool) {
```
wuhuizuo, line 247: can we update the comment also, like this:
```go
			// we want to appropriately trigger and skip from the set of identified presubmits that were
			// added. if a run_if_changed job has its regex pattern changed we should check PRs against the new regex pattern and
			// trigger if needed instead of forcing the job to run on all PRs. so the default behavior of the filter need to be false.
```

Issue #344, opened 2024-12-12
needs-triage label not removed when /triage unresolved


BenTheElder:
https://github.com/kubernetes/kubernetes/issues/128235#issuecomment-2540047850 (note I later manually removed it for now)

I would expect the `needs-` label to be removed when adding the triage label, it added the label but didn't remove needs-triage

/kind bug
/sig testing

cc @petr-muller do you think this is a flake, missing feature, or a bug?

[I would like to mark this as a help wanted if not good first issue depending on that]

petr-muller:
I'm not that familiar with that plugin but not removing a `needs-triage` while adding a correct label looks like a bug to me.

BenTheElder:
This also seems to happen with
`/triage needs-information`
in kubernetes/kubernetes, example: https://github.com/kubernetes/kubernetes/pull/129850#issuecomment-2625210222

I could've sworn this worked before, it seems like only `/triage accepted` is working.

/help
I'm going to signal boost this as help-wanted in SIG Testing.

elizabeth-dev:
this is being handled in kubernetes/test-infra#34255, not sure if it being a different repo means a new issue should be created there

BenTheElder:
Let's close this one, give you've found that it's clearly a misconfiguration and not a prow implementation bug, thank you!

BenTheElder:
(I would move it but IIRC you can't move issues across organizations)

PR #370, opened 2025-02-04
Remove a category label `needs-*` when the user is defining one

danilo-gemoli:
These are the available `needs-*` categories that can be applied to an issue (see [here](https://github.com/kubernetes-sigs/prow/blob/aae1f05957ef15eb73500e0d1d3c7f01db6bdabd/pkg/plugins/label/label.go#L40)):
```
needs-kind, needs-priority, needs-sig, needs-triage
```
When an issue has one of those, as `needs-triage`, and the user is defining that category like so:
```
/triage unresolved
```
then the `needs-triage` will be removed.

### DIFF
--- a/pkg/plugins/label/label.go
+++ b/pkg/plugins/label/label.go
@@ -143,6 +143,23 @@ func getLabelsFromGenericMatches(matches [][]string, labelFilter func(string) bo
 	return labels
 }
 
+// getCategoriesFromLabels extracts the categories from `labels` that are also in `needsLabels`.
+// Example:
+//
+//	labels 		:= []string{"triage/unresolved", "area/testing"}
+//	needsLabels := []string{"triage"}
+//	result 		:= []string{"triage"}
+func getCategoriesFromLabels(labels, needsLabels []string) sets.Set[string] {
+	needsLabelsSet := sets.New(needsLabels...)
+	categories := sets.New[string]()
+	for _, label := range labels {
+		if split := strings.Split(label, "/"); len(split) == 2 && needsLabelsSet.Has(split[0]) {
+			categories.Insert(split[0])
+		}
+	}
+	return categories
+}
+
 func handleComment(gc githubClient, log *logrus.Entry, config plugins.Label, e *github.GenericCommentEvent) error {
 	if e.Action != github.GenericCommentActionCreated {
 		return nil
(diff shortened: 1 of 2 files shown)


pkg/plugins/label/label.go
```diff
@@ -143,6 +143,23 @@ func getLabelsFromGenericMatches(matches [][]string, labelFilter func(string) bo
 	return labels
 }
 
+// getCategoriesFromLabels extracts the categories from `labels` that are also in `needsLabels`.
+// Example:
+//
+//	labels 		:= []string{"triage/unresolved", "area/testing"}
+//	needsLabels := []string{"triage"}
+//	result 		:= []string{"triage"}
+func getCategoriesFromLabels(labels, needsLabels []string) sets.Set[string] {
```
jmguzik, line 152:
```suggestion
func getCategoriesFromLabels(labels, needsLabels []string) sets.Set[string] {
	if len(labels) == 0 || len(needsLabels) == 0 {
		return sets.New[string]() 
	}
	needsLabelsSet := sets.New(needsLabels...)
	categories := sets.New[string]()
	for _, label := range labels {
		if prefix, _, found := strings.Cut(label, "/"); found && needsLabelsSet.Has(prefix) {
			categories.Insert(prefix)
		}
	}
	return categories
}
```
Just for you to consider. Returns early and more visible split.

jmguzik:
Is it in any way related to https://github.com/kubernetes-sigs/prow/issues/344?

pkg/plugins/label/label.go
petr-muller: Weak -1 from me, I'm a fan of private-unless-proven-to-be-necessary-public :sweat:

pkg/plugins/label/label.go
```diff
@@ -143,6 +143,23 @@ func getLabelsFromGenericMatches(matches [][]string, labelFilter func(string) bo
 	return labels
 }
 
+// getCategoriesFromLabels extracts the categories from `labels` that are also in `needsLabels`.
+// Example:
+//
+//	labels 		:= []string{"triage/unresolved", "area/testing"}
+//	needsLabels := []string{"triage"}
```
petr-muller, line 150: nit : weird indent

pkg/plugins/label/label.go
jmguzik: Its my bad. I meant insights of the function, my intention was not to propose to make it public. Let me correct that.

jmguzik: Corrected, again comment is about internals not about publicity, sorry for the confusion.

danilo-gemoli:
Yes, this PR is supposed to handle https://github.com/kubernetes-sigs/prow/issues/344. Thanks both for the reviews but I'll put this on hold and eventually close it if https://github.com/kubernetes/test-infra/pull/34255 turns out to be enough.
/hold

Issue #208, opened 2024-07-11
Configured Tests Dashboard


petr-muller:
Migrated https://github.com/kubernetes/test-infra/issues/28979 

**What would you like to be added:**
A new dashboard that lists all configured jobs in the Prow instance. It should be filterable by org and repo. It should show a prettified view of the yaml for the job. It should also link to the job-history page for each respective job.

**Why is this needed:**
Test maintainers need an easy way to see all of the jobs configured for a specific org or repo.
There are a few issues with existing methods for seeing jobs:

- Looking through the config files isn't a good user experience
- Job-History is not easily navigable, and doesn't provide a list view
- Deck only shows jobs that have executed very recently

[msugakov](https://github.com/msugakov) commented [Oct 4, 2023](https://github.com/kubernetes/test-infra/issues/28979#issuecomment-1746789397):

> I needed something like this to compare jobs and configs set up for different branches of the project. I ended up creating a [script](https://github.com/openshift/release/blob/27eb3f902442bb4a8871658c40b4c98e6ba99108/ci-operator/config/stackrox/stackrox/summarize.py) that spits out an HTML page that's [checked in](https://github.com/openshift/release/blob/27eb3f902442bb4a8871658c40b4c98e6ba99108/ci-operator/config/stackrox/stackrox/summary.html) along the configs and can be [viewed online](https://refined-github-html-preview.kidonng.workers.dev/openshift/release/raw/27eb3f902442bb4a8871658c40b4c98e6ba99108/ci-operator/config/stackrox/stackrox/summary.html) courtesy to [Refined GitHub](https://github.com/refined-github/refined-github#file-management). Not sure how much are my needs specific to ci-operator, though.

BenTheElder:
Worth noting: some prow instances (like prow.k8s.io) have a TON of jobs and this might be expensive to render.
FWIW users can find presubmits by commenting `/test` and the robot will reply with a full list.

Maybe a page that fetches the config endpoint and then allows you to filter by job type and repo and renders that subset?

> Deck only shows jobs that have executed very recently

Dependent on your sinker config. We keep ~24h or most recent execution in prow.k8s.io, though I've been thinking about if we might need to reduce this because it's a ton.

PR #372, opened 2025-02-07
Add `Configured Jobs Dashboard` to `deck`

smg247:
This adds a basic, read-only, dashboard that will display all configured jobs for an org (assuming there are 10 or fewer repos in the org) or an individual repo. There is also an index that lists all the orgs and their repos with links to the individual pages. Constituent `periodics` are determined by their `extra_refs`.

For: https://github.com/kubernetes-sigs/prow/issues/208


petr-muller:
:open_mouth: NICE

Do you have a testing deployment somewhere?

cmd/deck/configured_jobs.go
```diff
@@ -0,0 +1,190 @@
...
+	} else {
+		// for undecorated jobs assume the default
+		def := cfg().Plank.GuessDefaultDecorationConfig(fmt.Sprintf("%s/%s", org, repo), job.Cluster)
+		if def == nil || def.GCSConfiguration == nil {
+			return "", "", fmt.Errorf("failed to guess gcs config based on default decoration config")
+		}
+		gcsConfig = def.GCSConfiguration
+	}
+
+	b := gcsConfig.Bucket
```
jmguzik, line 179: Would it make sense to check if the bucket is empty and communicate such error?

cmd/deck/main.go
```diff
@@ -823,6 +827,34 @@ func handleProwJobs(ja *jobs.JobAgent, log *logrus.Entry) http.HandlerFunc {
...
 
+func handleConfiguredJobs(o options, cfg config.Getter, log *logrus.Entry) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		jobConfig := cfg().JobConfig
+		setHeadersNoCaching(w)
+		p := r.URL.Path
+		if strings.HasSuffix(p, "configured-jobs/") { // If there is no further path, fetch the index page
+			index := GetIndex(jobConfig)
+			handleSimpleTemplate(o, cfg, "configured-jobs-index.html", index)(w, r)
+		} else {
```
jmguzik, line 838: Instead of else I would return. Also I think this function needs to be a little bit more cautious around splitting and print errors if what is expected is not really there.

cmd/deck/static/configured-jobs/configured-jobs.ts
```diff
@@ -0,0 +1,166 @@
...
+    const cardDesc = document.createElement("div");
+    cardDesc.innerHTML = job.type;
+    cardDesc.classList.add("mdl-card__supporting-text");
+
+    const cardAction = document.createElement("div");
+    const actionButton = document.createElement("a");
+    actionButton.innerHTML = "Details";
+    actionButton.classList.add(...["mdl-button", "mdl-button--colored", "mdl-js-button", "mdl-js-ripple-effect"]);
+    actionButton.addEventListener("click", () => {
+        const dialogElement = document.querySelector("dialog");
```
jmguzik, line 118: I am not the master of ts, but should such statements be checked and error-ed if not found?

cmd/deck/configured_jobs.go
```diff
@@ -0,0 +1,190 @@
...
+			o := strings.Split(r, "/")[0]
+			if o == org {
+				orgRepos = append(orgRepos, r)
+			}
+		}
+	}
+
+	// If there are more than 10 repos in the org, the page will be slow and not particularly useful,
+	// Instead, just list the repos so they can be drilled down into
+	includeJobs := len(orgRepos) <= 10
```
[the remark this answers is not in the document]
smg247: That may be worth an enhancement later on, but currently there are **no** configurations necessary for this feature, and I don't think this is worth adding an entire config setup for.

cmd/deck/static/configured-jobs/configured-jobs.ts
smg247: I am also pretty green when it comes to ts, but this behavior was adapted from the following logic, and the "dialog" will never be not found unless someone introduces a bug with the HTML page. The bug would be painfully obvious, so I don't think we need additional error checking here.

https://github.com/kubernetes-sigs/prow/blob/739cbaba0899562e02585dc71587724e9b30c762/cmd/deck/static/plugin-help/plugin-help.ts#L30

cmd/deck/static/configured-jobs/configured-jobs.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+function genJobDetails(job: Job): string {
+    return `
+        <div>
+            <summary>Type: ${job.type}</summary>
+            <pre class="prettyprint"><code class="language-yaml job-definition">${job.yamlDefinition}</code></pre>
+            <a href="${job.jobHistoryLink}">Job History</a>
+        </div>
+    `;
+}
```
hector-vido, line 166: This is bigger than the original but is safer and performs better:

```suggestion
function genJobDetails<T extends HTMLElement>(job: Job): T {
    const div = document.createElement('div');
    const summary = document.createElement('summary');
    const pre = document.createElement('pre');
    const code = document.createElement('code');
    const a = document.createElement('a');
    
     a.href = job.jobHistoryLink;
     a.appendChild(document.createTextNode('Job History'));
     pre.className = 'prettyprint';
     code.className = 'language-yaml job-definition';
     
    code.appendChild(document.createTextNode(job.yamlDefinition));
    pre.appendChild(code);
    summary.appendChild(document.createTextNode(`Type: ${job.type}`));
    div.appendChild(summary);
    div.appendChild(pre);
    div.appendChild(a);
    
    return div as T;
}
```

> *PS*: I am not sure if the declaration and return type are right, also probably there is a "typescript way" to do that, but this can be done in the future.

cmd/deck/main.go
```diff
@@ -823,6 +827,43 @@ func handleProwJobs(ja *jobs.JobAgent, log *logrus.Entry) http.HandlerFunc {
 	}
 }
 
+func handleConfiguredJobs(o options, cfg config.Getter, log *logrus.Entry) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
```
Prucek, line 831: Don't we maybe want to filter only GET requests?
```
	if r.Method != "GET" {
	...
	}
```

Prucek:
We probably want to 
right?

jmguzik (approved):
Review from 3 people was done, we can always get back with the followup on possible ideas, fixes, improvements. 
LGTM from me.
/lgtm

PR #378, opened 2025-02-14
Do not suggest author when require_self_approval is true

### LINKED ISSUE #175: `require_self_approval: true` should prevent suggesting PR author for approval
[This is re-filed from https://github.com/kubernetes/test-infra/issues/29827]

**What happened**:

TLDR in a repo with implicit_self_approval off, prow will sometimes suggest the author as the approver, it should suggest someone else.

https://kubernetes.slack.com/archives/CDECRSC5U/p1669738130214399

suggest non-pr-author for finding an approver

.... you'll need implicit_self_approval disabled, yourself in OWNERS, and then file a PR until this happens.

_lots of sample links in the slack thread linked above_

/sig testing

nikhilnishad-goog:
When require_self_approval is set to true, HasSelfApproval is false and hence we must must not add author as an approver assignee. Fixed tests accordingly because there were many tests where require_self_approval was set to true and the author has been statically set in test logic to "cjwagner".

There was one last test which does not make sense at all so removed it. The tests wants to test the behaviour where we do not suggest cjwagner as approver but actually has alice assigned as reviewer. Not only that the comment clearly asks for approval from cjwagner.

Fixes #175


smg247:
Are 
https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/approve/approvers/owners.go#L684-L685
and 
https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/approve/approvers/owners.go#L690-L691
dead code paths now, or are they still relevant for some other reason?

nikhilnishad-goog:
> Are https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/approve/approvers/owners.go#L684-L685 and https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/approve/approvers/owners.go#L690-L691 dead code paths now, or are they still relevant for some other reason?

Not really, the coverage has been lost because the [permanent assignee](https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/approve/approve_test.go#L1279) was never an owner of anything.  Hence `AssignedCCs` returns empty list because "spxtr" does not own anything.  I think it makes sense to add a test case where even an assigned person could be suggested because that person owns something unapproved so far. Does not seem small+simple change though. I will give it a try in this PR so that test coverage is not reduced.


List of owners in the test ->

```
fr := fakeRepo{
		approvers: map[string]layeredsets.String{
			"a":   layeredsets.NewString("alice"),
			"a/b": layeredsets.NewString("alice", "bob"),
			"c":   layeredsets.NewString("cblecker", "cjwagner"),
		},
		leafApprovers: map[string]sets.Set[string]{
			"a":   sets.New[string]("alice"),
			"a/b": sets.New[string]("bob"),
			"c":   sets.New[string]("cblecker", "cjwagner"),
		},
		approverOwners: map[string]string{
			"a/a.go":                   "a",
			"a/aa.go":                  "a",
			"a/b/b.go":                 "a/b",
			"c/c.go":                   "c",
			"d/new-folder/new_file.go": "d",
		},
		autoApproveUnownedSubfolders: map[string]bool{
			"d": true,
		},
	}
```

PR #368, opened 2025-02-03
Handle empty alias group case

nikhilnishad-goog:
For this [issue](https://github.com/kubernetes-sigs/prow/issues/360). Modifies repoowners.go to handle cases where the file has an alias group but does not have any user name listed in it. 

There are multiple acceptable yamls that are being not parsed explictly and edge cases not handled. Examples are -

```
aliases:
  alias-group:
```

```
aliases:
  alias-group: {}
```

```
aliases:
  alias-group: {alias1, alias2}
```

```
aliases:
  alias-group: ""
```

All of these are valid yamls, that can cause silent failures like the one mentioned in #360 .

Fixes #360

PR #381, opened 2025-02-18
Revert "Handle empty alias group case"

petr-muller:
Reverts kubernetes-sigs/prow#368

The change unfortunately broke `approve` plugin in repositories with existing `OWNERS_ALIASES` file with at least one empty group: while looking for approvers it completely fails to resolve such `OWNERS_ALIASES` file and ignores it entirely:

https://github.com/kubernetes-sigs/prow/blob/15dd4c20845d1d483b526a611205516c4f7f4cac/pkg/repoowners/repoowners.go#L474-L479

The change actually also broke the `verify-owners` plugin that is supposed to guard against adding an invalid ownership files. This is because the plugin actually does not validate `OWNERS_ALIASES` that well:

https://github.com/kubernetes-sigs/prow/blob/15dd4c20845d1d483b526a611205516c4f7f4cac/pkg/plugins/verify-owners/verify-owners.go#L272-L277

It validates its content _provided it was correctly parsed_ :grimacing: If it fails to parse then the plugin just errors out instead of placing the label on the PR that adds such file.

Thanks @sbueringer for reporting!


sbueringer:
Makes sense to me, thank you!

I'll leave lgtm to someone familiar with the code

matthyx:
/lgtm
/hold
(holding so you can control when it will merge)

PR #382, opened 2025-02-19
Do not suggest author when require_self_approve is true.

### LINKED ISSUE #175: `require_self_approval: true` should prevent suggesting PR author for approval

nikhilnishad-goog:
Fixes #175

It seems that we were simply adding the author as an assignee when opts.HasSelfApproval()  (returns true if require_self_approval is false) was false. But we should not assign the author when require_self_approve is true. (i.e. when opts.HasSelfApproval() returns false).

Additionally, added test cases to cover for some unvcovered cases - when assignee and suggested approvers, both are needed. 

Also, removed a test case which didn't make any sense, it said do not suggest  to /assign already assigned "cjwagner" but "cjwagner" being the author should not be assigned when selfApprove is false (i.e. require_self_approval is true).

### DIFF
--- a/pkg/plugins/approve/approve.go
+++ b/pkg/plugins/approve/approve.go
@@ -439,10 +439,8 @@ func handle(log *logrus.Entry, ghc githubClient, repo approvers.Repo, githubConf
 	// Author implicitly approves their own PR if config allows it
 	if opts.HasSelfApproval() {
 		approversHandler.AddAuthorSelfApprover(pr.author, pr.htmlURL+"#", false)
-	} else {
-		// Treat the author as an assignee, and suggest them if possible
-		approversHandler.AddAssignees(pr.author)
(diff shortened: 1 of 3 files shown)


nikhilnishad-goog:
@petr-muller Can you please also review this PR and approve if good to go?

PR #384, opened 2025-02-19
Update create-build-cluster.sh to have mandatory package auto install

sjswerdlow:
Currently create-build-cluster.sh will fail after the cluster is made, due to not being able to authenticate with the cluster correctly.


nathanperkins:
We should check with @cjwagner whether this conflicts or needs adjustment for the changes that he just made over in https://github.com/GoogleCloudPlatform/oss-test-infra

PR #385, opened 2025-02-24
fix error shadowing in `sinker` integration test

smg247:
Follows up on https://github.com/kubernetes-sigs/prow/pull/383 to not shadow the `err` var so that the test will properly fail when it should


petr-muller (approved):

petr-muller:
I wonder if the test starts flaking again now :rofl:

smg247:
> I wonder if the test starts flaking again now :rofl: 

It shouldn't, but time will tell 🤷

PR #386, opened 2025-02-24
Allow Getting Closed PullRequests

midnightconman:
Functions to get closed, open, and all pull requests. This is possible because github takes a `state` parameter which can be toggled to change results.

### DIFF
--- a/pkg/github/client.go
+++ b/pkg/github/client.go
@@ -29,6 +29,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"slices"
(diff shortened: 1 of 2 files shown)


matthyx:
/ok-to-test
hi @midnightconman what's the use case?

midnightconman:
I had a use-case for metrics but the v3 calls are too expensive from a token standpoint and we have too many orgs / repos to cache these responses properly... so I'll go ahead and close this out. Thank you for taking a look! 👍

PR #387, opened 2025-02-25
`sinker` integration test: always wait 1 minute to check for pod deletion

smg247:
Follows up on: https://github.com/kubernetes-sigs/prow/pull/385 as it seems the integration test is still failing with a `wait` error.

The polling logic lead to flaky results due to potential race conditions. Simply waiting 1 minute to check should only introduce a slight delay (at most) to the overall process because these tests are already run in parallel.

Fixes: https://github.com/kubernetes-sigs/prow/issues/309

### DIFF
--- a/test/integration/test/sinker_test.go
+++ b/test/integration/test/sinker_test.go
@@ -282,8 +282,9 @@ func TestDeletePod(t *testing.T) {
 			}
 
 			// Make sure pod is deleted.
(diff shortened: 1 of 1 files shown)


smg247:
The failure in the last run was a different test entirely, unrelated to this change.

petr-muller:
I am not exactly thrilled about this (I feel there has to be a "proper" fix there somewhere) but I'm not the one who spent an immense time looking at this code and it at least makes the tests green so :shrug: 

/lgtm

PR #393, opened 2025-03-05
Additional func to retrieve default branch of an org/repo

deepsm007:
Changes:

- Added a new getDefaultBranch function to fetch a repo's default branch via Github client
- Introduced a simple GithubClient interface for easier testing
- Added unit tests to cover both success and error cases.

### DIFF
--- a/cmd/generic-autobumper/bumper/bumper.go
+++ b/cmd/generic-autobumper/bumper/bumper.go
@@ -681,6 +681,23 @@ func getLastBumpCommit(gerritAuthor, commitTag string) (string, error) {
 	return outBuf.String(), nil
 }
 
+type GithubClient interface {
+	GetRepo(ctx context.Context, org, repo string) (*github.Repo, error)
+}
+
+// getDefaultBranch retrieves the default branch name of a given repository
+// using the provided GitHub client. It fetches the repository details and
+// extracts the default branch. It returns a string representing the default
+// branch name and error if the repository details cannot be retrieved.
+func getDefaultBranch(gc GithubClient, org, repo string) (string, error) {
+	ctx := context.Background()
+	repository, err := gc.GetRepo(ctx, org, repo)
+	if err != nil {
+		return "", fmt.Errorf("failed to get repository details for %s/%s: %w", org, repo, err)
+	}
+	return repository.DefaultBranch, nil
+}
+
 // getChangeId generates a change ID for the gerrit PR that is deterministic
 // rather than being random as is normally preferable.
 // In particular this chooses a change ID by hashing the last commit by the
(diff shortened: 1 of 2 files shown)


cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -681,6 +681,28 @@ func getLastBumpCommit(gerritAuthor, commitTag string) (string, error) {
...
+// using the provided GitHub client. It fetches the repository details and
+// extracts the default branch. It returns a string representing the default
+// branch name and error if the repository details cannot be retrieved.
+func getDefaultBranch(gc github.Client, org, repo string) (string, error) {
+	repository, err := gc.GetRepo(org, repo)
+	if err != nil {
+		return "", fmt.Errorf("failed to get repository details for %s/%s: %w", org, repo, err)
+	}
+	return repository.DefaultBranch, nil
+}
```
petr-muller, line 694: TBH adding this wrapper does not simplify the logic much, and the indirection just to extract a single member from the `repository` seems like an overkill.

The callsite in `createOrUpdatePROnDefaultBranch` will look basically identical if it uses `GetRepo` directly, and then just `repository.DefaultBranch` where it uses `defaultBranch` now. I'd prefer that.

go.mod
```diff
@@ -195,6 +195,13 @@ require (
 )
 
 require (
+	github.com/alecthomas/colour v0.1.0 // indirect
+	github.com/alecthomas/repr v0.0.0-20210801044451-80ca428c5142 // indirect
+	github.com/mattn/go-isatty v0.0.20 // indirect
+)
+
+require (
+	github.com/alecthomas/assert v1.0.0
```
smg247, line 204: Were you not able to find any similar unit testing pattern here without including a new assertion lib?

cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -681,6 +681,16 @@ func getLastBumpCommit(gerritAuthor, commitTag string) (string, error) {
 	return outBuf.String(), nil
 }
 
+// createOrUpdatePROnDefaultBranch retrieves the default branch for the repository
+// and creates/updates a PR on that branch without requiring the caller to specify the branch.
+func createOrUpdatePROnDefaultBranch(gc github.Client, org, repo string, opts *Options, prTitle, prBody string) error {
```
smg247, line 686: This function will need to be exported to be useful

smg247:
I would stop trying to use the `fakegithubclient` for the purpose of this unit test. If you follow [this example](https://github.com/kubernetes-sigs/prow/blob/2cddb48099687e374516c0188e833994bbaef9fa/cmd/branchprotector/protect.go#L158-L168) to define a new interface in `bumper` for only the ghc methods it actually needs then you can just mock those methods with a fake client of your own. They have done this exact thing in [branchprotector](https://github.com/kubernetes-sigs/prow/blob/16c13cde5644f1c40747c358ad3deb77899f45ce/cmd/branchprotector/protect_test.go#L83-L105).

PR #95, opened 2024-03-25
Various fixes for presubmit tests.

cjwagner:
/hold

PR #403, opened 2025-03-22
ci: remove python dep installation

petr-muller:
This was added in #95 (specifically 6947186ebd7926acd0cf6bfdf64c85f31c5f5ca1) but it seems that packages installed were never actually used, the verify_boilerplate.py script only uses standard library imports.


petr-muller:
Incorrect boilerplate detection works:

```
1 files have incorrect boilerplate headers:
cmd/exporter/collector.go
ERROR: boilerplate failed
```

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_prow/403/pull-prow-verify-lint/1903241026328334336

krzyzacy (approved):
lgtm as the boilerplate check still passes

Issue #406, opened 2025-03-23
Proposal: Add an include_details option to retrieve more job information


LiangquanLi930:
Currently, when querying a job via the Gangway API, the response includes a limited set of information, like this:
```
{
  "id": "xxxx",
  "job_name": "xxxx",
  "job_type": "xxxx",
  "job_status": "xxxx",
  "gcs_path": ""
}
```
However, in some cases, we need more detailed information about the job, such as the `url`, `startTime`, and `completionTime`. Therefore, I propose adding an include_details parameter to the API query:
+ When include_details=true, the API should return additional information, such as:
    + url: A link to the job details
    + startTime: The time when the job started
    + completionTime: The time when the job completed

This enhancement would make it easier to gather more comprehensive information about jobs, particularly useful for debugging or monitoring purposes.

PR #407, opened 2025-03-23
Add more job information to GetJobExecution

LiangquanLi930:
### Changes
information
- Added `job_url` field to `JobExecution` message to return the job's URL
- Modified `GetJobExecution` method to return job URL, creation time, and completion time

### Purpose
This change allows clients to optionally retrieve more detailed job information when fetching job details, including the job's URL, creation time, and completion time. This is particularly useful for scenarios where detailed job information needs to be displayed.

### Impact
- API changes: Added new fields `job_url`
- Performance impact: Returns more data when requesting detailed information, but the impact is minimal

### Which issue(s) this PR fixes:
Fixes https://github.com/kubernetes-sigs/prow/issues/406


petr-muller:
I am not that knowledgeable about Gangway. Providing more details in a response does not sound bad to me, but I do not like driving this through a boolean modal, it increases complexity. The added details do not seem that heavy, should we just return them by default, without introducing modality?

LiangquanLi930:
> I am not that knowledgeable about Gangway. Providing more details in a response does not sound bad to me, but I do not like driving this through a boolean modal, it increases complexity. The added details do not seem that heavy, should we just return them by default, without introducing modality?

Thanks for your review! I chose to add include_details because this approach ensures compatibility with the existing code and doesn’t break systems that rely on the current behavior. That’s why I opted for this safer choice.

Prucek:
I am just wondering, since you are adding fields to the response, why are the fields not used?
```
jobExec := &JobExecution{
  Id:        prowJobCR.Name,
  JobStatus: TranslateProwJobStatus(&prowJobCR.Status),
}
```
instead of:
```
jobExec := &JobExecution{
	Id:        prowJobCR.Name,
	JobName:   prowJobCR.Spec.Job,
	JobType:   TranslateProwJobType(prowJobCR.Spec.Type),
	JobStatus: TranslateProwJobStatus(&prowJobCR.Status),
}
```

LiangquanLi930:
Thanks all, updated

LiangquanLi930:
> I wonder though if that could somehow impact transcoding? 🤔

That's why I want to add include_details. Maybe @petr-muller and @Prucek can help answer this? Thanks!

petr-muller:
Adding fields should be safe in Protobuf, but I admit I know very little about how Gangway is used from the client-side.

Prucek (approved):

LiangquanLi930:
@petr-muller When you have time, could you help to approve this PR? Thanks

jmguzik:
@Prucek protobuf I agree, I was more interested in our transcoding case here. Probably after merge we need to update definition.

petr-muller (approved):

PR #412, opened 2025-03-28
Consolidate revival logic in case pod is in an unknown, evicted or unreachable state, limit the number of revivals to 3

inteon:
Consolidates the revival logic in `pkg/plank/reconciler.go` and limits the number of times we "revive" to 3 before we fail the prowjob.
Revival happens in case the pod is in an unknown, evicted or unreachable state.

**Future work:** in a follow-up PR, I'll add support for handling pods that are in a terminated state (the node on which the pod was running was shut-down).

REVIEWER NOTE: "revival" might not be a good term, feel free to suggest a different name

### DIFF
--- a/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml
+++ b/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml
@@ -9788,6 +9788,10 @@ spec:
                   PrevReportStates stores the previous reported prowjob state per reporter
                   So crier won't make duplicated report attempt
                 type: object
+              revivalCount:
+                description: Amount of times the Pod was revived from an unexpected
+                  stop.
+                type: integer
               startTime:
                 description: StartTime is equal to the creation time of the ProwJob
                 format: date-time
--- a/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml
+++ b/config/prow/cluster/prowjob-crd/prowjob_customresourcedefinition.yaml
@@ -9780,6 +9780,12 @@ spec:
                   plank. This field should always be the same as
                   the ProwJob.ObjectMeta.Name field.
                 type: string
+              pod_revival_count:
+                description: |-
+                  PodRevivalCount applies only to ProwJobs fulfilled by
+                  plank. This field shows the amount of times the
+                  Pod was recreated due to an unexpected stop.
+                type: integer
               prev_report_states:
                 additionalProperties:
                   description: ProwJobState specifies whether the job is running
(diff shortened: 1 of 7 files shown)


petr-muller:
Feel free to ignore the AI review, I just wanted to try how it behaves on a PR like this :+1:

petr-muller:
From the API lifecycle perspective I am a bit worried about Prow instances that update their controlelrs often but not the CRD. Assuming `x-kubernetes-preserve-unknown-fields: false`, the `status.revivalCount` gets pruned, then the client will see the `RevivalCount` having zero value when reading from the server, leading to revivals until timeout. Could we prevent that by tracking remaining revival budget instead? That way the default zero would mean "do not revive" me.

Some other points but mostly thinking out loud:
1. This changes behavior for disabled `ErrorOnEviction`, right? Absent explicit configuration, previously we retried until timeout, now we default to 3. To me this is acceptable.
2. No strong feeling about `revival` name. My slight issue is that without context it may not be clear what the term means. I assume you had a reason for not calling it `retry`? Something like `DisruptedPodRetries` is a bit verbose buy maybe more explicit. Just throwing ideas here, I don't have a issue if "revival" merges.

petr-muller (commented):
From the API lifecycle perspective I am a bit worried about Prow instances that update their controlelrs often but not the CRD. Assuming `x-kubernetes-preserve-unknown-fields: false`, the `status.revivalCount` gets pruned, then the client will see the `RevivalCount` having zero value when reading from the server, leading to revivals until timeout. Could we prevent that by tracking remaining revival budget instead? That way the default zero would mean "do not revive" me.

Some other points but mostly thinking out loud:
1. This changes behavior for disabled `ErrorOnEviction`, right? Absent explicit configuration, previously we retried until timeout, now we default to 3. To me this is acceptable.
2. No strong feeling about `revival` name. My slight issue is that without context it may not be clear what the term means. I assume you had a reason for not calling it `retry`? Something like `DisruptedPodRetries` is a bit verbose buy maybe more explicit. Just throwing ideas here, I don't have a issue if "revival" merges.

inteon:
> From the API lifecycle perspective I am a bit worried about Prow instances that update their controlelrs often but not the CRD. Assuming `x-kubernetes-preserve-unknown-fields: false`, the `status.revivalCount` gets pruned, then the client will see the `RevivalCount` having zero value when reading from the server, leading to revivals until timeout. Could we prevent that by tracking remaining revival budget instead? That way the default zero would mean "do not revive" me.

In case the CRD is not updated, I think the "revivals until timeout" is more backwards compatible with the current behavior than never reviving and erroring on first failure.

> Some other points but mostly thinking out loud:
> 
> 1. This changes behavior for disabled `ErrorOnEviction`, right? Absent explicit configuration, previously we retried until timeout, now we default to 3. To me this is acceptable.

Indeed, now we will retry 3 times instead of just retrying until timeout.

PR #413, opened 2025-04-02
bug: set state to error on retired context

janisz:
When action set description to `Context retired without replacement.` it also sets status to :heavy_check_mark:  which leads to false assumption that job passed while it was not. Below is the example where required job was failing but prow set it's status to passed causing merge and breaking the master builds. 

The generated comment is correct https://github.com/stackrox/stackrox/pull/14800#issuecomment-2761889050
while  status is not

image (3)

Refs:
- https://github.com/stackrox/stackrox/pull/14800
- :red_circle: [stackrox-master-gke-nongroovy-e2e-tests](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/stackrox_stackrox/14800/pull-ci-stackrox-stackrox-master-gke-nongroovy-e2e-tests/1905584438360674304)

### DIFF
--- a/pkg/statusreconciler/migrator/migrator.go
+++ b/pkg/statusreconciler/migrator/migrator.go
@@ -130,11 +130,10 @@ func copyAction(origContext, newContext string) func(statuses []github.Status, s
 
 // retireAction creates a function that returns a retire action.
 // Specifically the returned function returns a RepoStatus that will update the origContext status
-// to 'success' and set it's description to mark it as retired and replaced by newContext.
+// to error and set it's description to mark it as retired and replaced by newContext.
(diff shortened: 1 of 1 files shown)


smg247:
If the context is being retired, and we set it as `error`, won't that mean that it would have to be overridden for the PR to merge? If the context is retired, doesn't that mean that it will no longer be required for a PR to merge?

stevekuznetsov:
> Then the new status will override it 

Isn't the logic behind retired without replacement is that there is no new status to override it? Did we incorrectly compute "retired without replacemenet"?

stevekuznetsov (changes requested):
This change is incorrect. I think we likely have a bug in detecting when something is retired without replacement.

PR #433, opened 2025-04-17
feat: add ghproxy valkey/redis ACL support

kairoaraujo:
add ghproxy support to valkey/redis using ACL (username and password)

when using `--redis-address` the user can pass the username and password if the redis server has auth

New parameters:
`--redis-username=username`
`--redis-password=password`

### DIFF
--- a/cmd/ghproxy/ghproxy.go
+++ b/cmd/ghproxy/ghproxy.go
@@ -93,7 +93,9 @@ type options struct {
 	sizeGB                                 int
 	diskCacheDisableAuthHeaderPartitioning bool
 
-	redisAddress string
+	redisAddress  string
+	redisUsername string
+	redisPassword string
 
 	port           int
 	upstream       string
@@ -143,6 +145,8 @@ func flagOptions() *options {
 	flag.IntVar(&o.sizeGB, "cache-sizeGB", 0, "Cache size in GB per unique token if using a disk cache.")
 	flag.BoolVar(&o.diskCacheDisableAuthHeaderPartitioning, "legacy-disable-disk-cache-partitions-by-auth-header", true, "Whether to disable partitioning a disk cache by auth header. Disabling this will start a new cache at $cache_dir/$sha256sum_of_authorization_header for each unique authorization header. Bigger setups are advise to manually warm this up from an existing cache. This option will be removed and set to `false` in the future")
 	flag.StringVar(&o.redisAddress, "redis-address", "", "Redis address if using a redis cache e.g. localhost:6379.")
+	flag.StringVar(&o.redisUsername, "redis-usersname", "", "Redis username")
+	flag.StringVar(&o.redisPassword, "redis-password", "", "Redis password")
 	flag.IntVar(&o.port, "port", 8888, "Port to listen on.")
 	flag.StringVar(&o.upstream, "upstream", "https://api.github.com", "Scheme, host, and base path of reverse proxy upstream.")
 	flag.IntVar(&o.maxConcurrency, "concurrency", 25, "Maximum number of concurrent in-flight requests to GitHub.")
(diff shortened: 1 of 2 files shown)


cmd/ghproxy/ghproxy.go
```diff
@@ -143,6 +145,8 @@ func flagOptions() *options {
 	flag.IntVar(&o.sizeGB, "cache-sizeGB", 0, "Cache size in GB per unique token if using a disk cache.")
 	flag.BoolVar(&o.diskCacheDisableAuthHeaderPartitioning, "legacy-disable-disk-cache-partitions-by-auth-header", true, "Whether to disable partitioning a disk cache by auth header. Disabling this will start a new cache at $cache_dir/$sha256sum_of_authorization_header for each unique authorization header. Bigger setups are advise to manually warm this up from an existing cache. This option will be removed and set to `false` in the future")
 	flag.StringVar(&o.redisAddress, "redis-address", "", "Redis address if using a redis cache e.g. localhost:6379.")
+	flag.StringVar(&o.redisUsername, "redis-usersname", "", "Redis username")
```
smg247, line 148:
```suggestion
	flag.StringVar(&o.redisUsername, "redis-username", "", "Redis username")
```

cmd/ghproxy/ghproxy.go
```diff
@@ -143,6 +145,8 @@ func flagOptions() *options {
 	flag.IntVar(&o.sizeGB, "cache-sizeGB", 0, "Cache size in GB per unique token if using a disk cache.")
 	flag.BoolVar(&o.diskCacheDisableAuthHeaderPartitioning, "legacy-disable-disk-cache-partitions-by-auth-header", true, "Whether to disable partitioning a disk cache by auth header. Disabling this will start a new cache at $cache_dir/$sha256sum_of_authorization_header for each unique authorization header. Bigger setups are advise to manually warm this up from an existing cache. This option will be removed and set to `false` in the future")
 	flag.StringVar(&o.redisAddress, "redis-address", "", "Redis address if using a redis cache e.g. localhost:6379.")
```
smg247, line 147: I would add some validation that states that `redis-address` must be supplied if either of the 2 new fields are

pkg/ghcache/ghcache.go
```diff
@@ -537,8 +537,18 @@ func NewFromCache(roundTripper http.RoundTripper, cache CachePartitionCreator, m
...
+
+	var conn redis.Conn
+	var err error
+	if redisUsername != "" {
+		conn, err = redis.Dial("tcp", redisAddress, redis.DialUsername(redisUsername), redis.DialPassword(redisPassword))
+	} else if redisPassword != "" {
+		conn, err = redis.Dial("tcp", redisAddress, redis.DialPassword(redisPassword))
+	} else {
+		conn, err = redis.Dial("tcp", redisAddress)
+	}
```
smg247, line 550: Rather than creating the connection in three separate statements, can we just always pass the two new options? it seems like passing empty strings for either of the options would have the same result as not including them, but that should be verified.

cmd/ghproxy/ghproxy.go
```diff
@@ -143,6 +145,8 @@ func flagOptions() *options {
...
+	flag.StringVar(&o.redisUsername, "redis-usersname", "", "Redis username")
+	flag.StringVar(&o.redisPassword, "redis-password", "", "Redis password")
```
petr-muller, line 149: I was not able to a full review but I see we add a sensitive value parameter.

1. It should not be passed as a value; a typical pattern in Prow is to pass a path to a file with the secret as content
2. Internally we should work with secrets through the [agent](https://github.com/kubernetes-sigs/prow/tree/main/pkg/config/secret). It handles reading the secret from the file, reloading (maybe not that useful here) and log censoring. Whatever needs the secret will fetch it from the agent.

petr-muller (approved):
Thanks! LGTM, I have added a small change to address a minor issue.

PR #439, opened 2025-04-25
adds regex support for branch matching in the Tide status controller

### LINKED ISSUE #438: Add regex support for branch matching in Tide status controller
Currently, the Tide status controller only supports exact string matching for branch names in the `includedBranches` and `excludedBranches` fields of Tide queries. However, the branchprotector component already supports regex pattern matching for branch names, which provides more flexibility.

This inconsistency means that users need to list each branch explicitly in Tide queries, even when they follow a pattern (like feature branches or release branches). Adding regex support would make the Tide status controller's behavior consistent with the branchprotector and provide a more powerful way to specify branch patterns.

Reference: The branchprotector implementation at `cmd/branchprotector/protect.go` around line 322 already supports regex for branch matching.

kaovilai:
### What this PR does / why we need it
This PR adds regex support for branch matching in the Tide status controller, similar to how it's implemented in the branchprotector. This allows for more flexible branch matching in Tide queries using regular expressions.

For example, with this change, users can now use patterns like:
- `feature-.*` to match all feature branches
- `release-\d+\.\d+` to match release branches with version numbers
- Complex patterns with alternation like `(main|master|release-.*)`

### Which issue(s) this PR fixes
Fixes #438

### Special notes for your reviewer
- The implementation compiles regex patterns from branch lists joined with the `|` operator
- It falls back to exact string matching if regex compilation fails, ensuring backward compatibility
- A comprehensive test file was added to verify the functionality

### How Has This Been Tested?
- Added unit tests in `pkg/tide/requirementdiff_test.go` that cover various scenarios:
  - Exact branch name matching
  - Regex pattern matching for feature branches
  - Regex pattern matching for release branches
  - Complex patterns with alternation
  - Multiple branch patterns
  - Invalid regex patterns (fallback to exact matching)
- Ran existing tests to ensure no regressions

### Does this PR introduce a user-facing change?
Yes. Users can now use regex patterns in the `includedBranches` and `excludedBranches` fields of Tide queries.

Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

### DIFF
--- a/pkg/tide/status.go
+++ b/pkg/tide/status.go
@@ -145,37 +145,12 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
 
 	// Weight incorrect branches with very high diff so that we select the query
 	// for the correct branch.
-	targetBranchDenied := false
-	if len(q.ExcludedBranches) > 0 {
-		branchExclusions, err := regexp.Compile(strings.Join(q.ExcludedBranches, `|`))
-		if err == nil && branchExclusions.MatchString(string(pr.BaseRef.Name)) {
-			targetBranchDenied = true
-		} else if err != nil {
-			// Fall back to exact matching if regex compilation fails
-			for _, excludedBranch := range q.ExcludedBranches {
-				if string(pr.BaseRef.Name) == excludedBranch {
-					targetBranchDenied = true
-					break
-				}
-			}
-		}
-	}
+	targetBranchDenied := matchBranchPattern(string(pr.BaseRef.Name), q.ExcludedBranches)
 	
 	// if no allowlist is configured, the target is OK by default
 	targetBranchAllowed := len(q.IncludedBranches) == 0
-	if !targetBranchAllowed && len(q.IncludedBranches) > 0 {
-		branchInclusions, err := regexp.Compile(strings.Join(q.IncludedBranches, `|`))
-		if err == nil && branchInclusions.MatchString(string(pr.BaseRef.Name)) {
-			targetBranchAllowed = true
-		} else if err != nil {
-			// Fall back to exact matching if regex compilation fails
-			for _, includedBranch := range q.IncludedBranches {
-				if string(pr.BaseRef.Name) == includedBranch {
-					targetBranchAllowed = true
-					break
-				}
-			}
-		}
+	if !targetBranchAllowed {
+		targetBranchAllowed = matchBranchPattern(string(pr.BaseRef.Name), q.IncludedBranches)
 	}
 	if targetBranchDenied || !targetBranchAllowed {
 		diff += 2000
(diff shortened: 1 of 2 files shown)


pkg/tide/status.go
```diff
@@ -145,18 +146,35 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
...
 			targetBranchDenied = true
-			break
+		} else if err != nil {
+			// Fall back to exact matching if regex compilation fails
+			for _, excludedBranch := range q.ExcludedBranches {
+				if string(pr.BaseRef.Name) == excludedBranch {
+					targetBranchDenied = true
+					break
+				}
+			}
```
smg247, line 160: Is it possible to extract a method to avoid the (mostly complete) duplication between this and the includedBranch logic?

petr-muller:
/hold

I do not think there is any benefit in doing this in Tide status controller, and unfortunately this does not do what you probably want to achieve, sorry.

Tide has two controllers, one that handles actual merges, one that manages the commit status. You modify the latter in a very peculiar way. GitHub Search API (which is what Tide uses to determine mergeable PRs; Tide `queries` configuration is literally queries passed almost directly to GH Search API to find mergeable PRs) does not support regular expressions, so it would not be possible to use regexes in Tide configuration queries in the current form.

The status controller basically only does a search for a superset of mergeable PRs: things that could become mergeable in the future, but not right now. Adding some kind of regex processing to status controller's `requirementDiff` (method used to compute the `Needs approved label.`-like strings to be put on GH commit statuses would be useful if the _rest of Tide_ would be able to work with regexes in its queries, but that is not the case. Hypothetically you could modify Tide to process the regexes itself and "explode" the regex into a query (Tide could obtain a list of branches in the repo and generate the necessary "precise" GH Search API-compliant queries internally) but that would be a much larger change.

petr-muller:
Additionally, anything that simply changes how the existing config is treated also needs to be careful to not change the semantics of the existing config. We cannot simply start treating existing strings as regexes, especially when they commonly contain dots (like branch names do).

kaovilai:
> you could modify Tide to process the regexes itself and "explode" the regex into a query (Tide could obtain a list of branches in the repo and generate the necessary "precise" GH Search API-compliant queries internally) but that would be a much larger change.

ok.

> to not change the semantics of the existing config.

Perhaps a v2 config?

petr-muller:
> Perhaps a v2 config?

You could add `includedBranchPatterns` and `excludedBranchPattern` fields to query configs but the main problem is to properly translate such user input into something Tide could query GH with.

smg247:
/lgtm cancel
I am not sure how I didn't see that this was only updating the status sync loop. Thanks for double-checking this @petr-muller

petr-muller:
No worries, Tide is not the easiest piece of Prow to grok. I think that if we ever manage to implement this we'll need to gate it by some real deployment testing. It is very hard to test how Tide would work without exposing it to the actual GH API.

kaovilai:
maybe a "prow-unit-test-repo"

PR #440, opened 2025-04-28
fix(pkg/github): skip skipped workflow runs when collecting failed runs

wuhuizuo:
This pull request updates the logic for filtering GitHub workflow runs and includes a minor import addition to support the changes. The primary goal is to refine the selection criteria for completed jobs by excluding both "skipped" and "successful" runs.

### Logic updates for filtering workflow runs:
* Updated the filtering condition in `GetFailedActionRunsByHeadBranch` to exclude runs with the conclusion "skipped" in addition to "success". This ensures that skipped workflows are no longer considered as failed runs. (`pkg/github/client.go`, [pkg/github/client.goR2082-R2086](diffhunk://#diff-2bb9ccd2fcbbcfd99cb8e400704ea77cc80dbf62f5982b1f6ed1a0966ac42ae5R2082-R2086))

### Codebase enhancements:
* Added the `slices` package to the import list to leverage its utility functions, specifically `slices.Contains`, for cleaner and more readable filtering logic. (`pkg/github/client.go`, [pkg/github/client.goR32](diffhunk://#diff-2bb9ccd2fcbbcfd99cb8e400704ea77cc80dbf62f5982b1f6ed1a0966ac42ae5R32))

### DIFF
--- a/pkg/github/client.go
+++ b/pkg/github/client.go
@@ -29,6 +29,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -2078,10 +2079,11 @@ func (c *client) GetFailedActionRunsByHeadBranch(org, repo, branchName, headSHA 
 	// See https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow
 	// This makes it hard to use directly. Instead, we loop through the runs and check them individually.
 	// A successful workflow will have status "completed" and conclusion "success".
+	// A skipped workflow will have status "completed" and conclusion "skipped".
 	// A failed workflow also have status "completed", but the conclusion can be either "failure" or "cancelled".
-	// We only want completed jobs that are not successful.
+	// We only want completed jobs that are not skipped and not successful.
 	for _, run := range runs.WorkflowRuns {
-		if run.Status == "completed" && run.Conclusion != "success" {
+		if run.Status == "completed" && !slices.Contains([]string{"success", "skipped"}, run.Conclusion) {
 			prRuns = append(prRuns, run)
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/github/client.go
```diff
@@ -2078,10 +2079,11 @@ func (c *client) GetFailedActionRunsByHeadBranch(org, repo, branchName, headSHA
 	// See https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow
 	// This makes it hard to use directly. Instead, we loop through the runs and check them individually.
 	// A successful workflow will have status "completed" and conclusion "success".
+	// A skipped workflow will have status "completed" and conclusion "skipped".
 	// A failed workflow also have status "completed", but the conclusion can be either "failure" or "cancelled".
-	// We only want completed jobs that are not successful.
+	// We only want completed jobs that are not skipped and not successful.
 	for _, run := range runs.WorkflowRuns {
-		if run.Status == "completed" && run.Conclusion != "success" {
+		if run.Status == "completed" && !slices.Contains([]string{"success", "skipped"}, run.Conclusion) {
```
smg247, line 2086: With only 2 values to check this seems like overkill, but if you really want to use `slices.Contains` you should make a const slice for the values to check outside of the function.

Personally, I would just use check both values individually.

wuhuizuo: GitHub Actions workflows return a conclusion value after completion. The possible conclusion values are:
- `success`: The workflow completed successfully
- `failure`: The workflow failed due to test failures, code errors, etc.
- `cancelled`: The workflow was manually cancelled
- `skipped`: The workflow was skipped (not run)
- `timed_out`: The workflow exceeded the maximum allowed runtime
- `action_required`: Requires a user action before continuing
- `neutral`: The workflow completed but neither succeeded nor failed (neutral outcome)

So I think the passed values are `["success", "skipped"]`, the `action_required` should be treat as failed state in triggering.

PR #441, opened 2025-04-29
Adding the ability to specify a custom backoff

bradmwilliams:
This PR adds the ability to specify a custom backoff function into the Jira Client's underlying retryablehttp client.  This provides a convenient way to allow consumers to specify their own Backoff logic if/when the `DefaultBackoff` ([here](https://github.com/hashicorp/go-retryablehttp/blob/390c1d807b1dfda09c64e992bdd5e58a00daa698/client.go#L551-L566)) is not sufficient.

### DIFF
--- a/pkg/flagutil/jira.go
+++ b/pkg/flagutil/jira.go
@@ -20,7 +20,9 @@ import (
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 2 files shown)


petr-muller (approved):
Thanks!

PR #456, opened 2025-05-13
Allow reason with /unhold too

stephenfin:
Since commit 69ffdfc, the `/hold` command allows a reason. For example:

```
/hold while we wait for foo to merge
```

However, you can't do the equivalent for `/unhold` and its variants. This is confusing. Correct it. This opens up the chance that someone could do e.g.

```
/hold cancel everything while I review this epic code
```

And get an unhold rather than a hold, but that seems an acceptable risk.


stephenfin:
/unhold

That regex obviously isn't aware that I've included the hold as examples in code blocks :sweat_smile:

PR #465, opened 2025-05-19
fix(generic comment): avoid ok-to-test message on IgnoreOkToTest

dhiller:
Emit a message similar to the welcome message in case of untrusted user and trigger.IgnoreOkToTest.

See https://github.com/kubevirt/kubevirt/pull/14110#issuecomment-2781210119

### DIFF
--- a/pkg/plugins/trigger/generic-comment.go
+++ b/pkg/plugins/trigger/generic-comment.go
@@ -92,7 +92,12 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 			return err
 		}
 		if !trusted {
-			resp := "Cannot trigger testing until a trusted user reviews the PR and leaves an `/ok-to-test` message."
+			var resp string
+			if trigger.IgnoreOkToTest {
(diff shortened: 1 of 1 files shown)


pkg/plugins/trigger/generic-comment.go
```diff
@@ -92,7 +92,12 @@ func handleGenericComment(c Client, trigger plugins.Trigger, gc github.GenericCo
 			return err
 		}
 		if !trusted {
-			resp := "Cannot trigger testing until a trusted user reviews the PR and leaves an `/ok-to-test` message."
+			var resp string
+			if trigger.IgnoreOkToTest {
+				resp = "PRs from untrusted users cannot be marked as trusted with `/ok-to-test` in this repo meaning untrusted PR authors can never trigger tests themselves. Collaborators can still trigger tests on the PR using `/test all`."
```
0xFelix, line 97: Mention that individual tests can be triggered too, not just `all`?

dhiller: @0xFelix I think I agree that we should not suggest to always run `/test` all, maybe just `/test` is enough. WDYT?

0xFelix (approved):
/lgtm

PR #466, opened 2025-05-20
Update config-bootstrapper.md

clubanderson:
add an example of usage as a periodic


clubanderson:
closing - I misunderstood how the config-updater works.

PR #469, opened 2025-05-30
plugin/label: Improve error when adding/removing a restricted label

stephenfin:
Currently, if you attempt to add a restricted label and `allowed_teams` is unset/empty, you get the following unhelpful message.

```
@username: Can not set label cherry-pick-approved: Must be member in one of these teams: []
```

Improve this so that we no longer suggest being a member of no team and instead suggest reaching out to one of the users in `allowed_users` instead:

```
@username: The label(s) `restricted-label` cannot be applied or removed, because you are not in one of the allowed teams and are not an allowed user. Consider assigning one of the following members: Mallory
```

We also improve the common path where `allowed_teams` is populated, aligning the error message with those of other errors paths:

```
@username: The label(s) `restricted-label` cannot be applied or removed, because you are not in one of the allowed teams and are not an allowed user. Must be a member of one of these teams: privileged-group
```

> [!NOTE]
> I suspect dumping a list of users could get "chatty", so I've intentionally
> avoided formatting these as links (i.e. we emit `user`, not `@user`) and only
> print up to 3 users. In most cases, this code should not be triggered since
> projects will usually make use of `allowed_teams`.

### DIFF
--- a/pkg/plugins/label/label.go
+++ b/pkg/plugins/label/label.go
@@ -332,7 +332,11 @@ func canUserSetLabel(ghc githubClient, org string, user string, label string, re
 		}
 	}
 
-	return false, fmt.Sprintf("Can not set label %s: Must be member in one of these teams: %v", label, config.AllowedTeams), nil
+	msg := fmt.Sprintf("The label(s) `%s` cannot be applied or removed, because you are not in one of the allowed teams and are not an allowed user.", label)
+	if len(config.AllowedTeams) > 0 {
+		msg += fmt.Sprintf(" Must be a member of one of these teams: %v", strings.Join(config.AllowedTeams, ", "))
+	}
+	return false, msg, nil
 }
 
 func handleLabelAdd(gc githubClient, log *logrus.Entry, config plugins.Label, e *github.PullRequestEvent) error {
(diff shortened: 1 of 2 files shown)


stephenfin:
/test pull-prow-integration

That...doesn't look related...

stephenfin:
/retest-required

`pull-prow-unit-test-race-detector-nonblocking` is failing on `pkg/interrupts/interrupts_test.go:183`, which I'm not touching here.

`pull-prow-integration` failed on the same test, but again this doesn't look related

smg247:
I see no reason why the message can't display the allowed teams and (a subset of) the allowed users as well. You can format a GH message so that it hides content by default to be expanded, maybe if the list is longer than 5 or so you can put an ellipsis, and hide the rest behind it. I am not sure that I would bother with returning random users. In situations where there are 20+ users, maybe just return the first 20... WDYT?

stephenfin:
> I see no reason why the message can't display the allowed teams and (a subset of) the allowed users as well. You can format a GH message so that it hides content by default to be expanded, maybe if the list is longer than 5 or so you can put an ellipsis, and hide the rest behind it. I am not sure that I would bother with returning random users. In situations where there are 20+ users, maybe just return the first 20... WDYT?

If you're happy, I'm happy. I'll bump the limit to 20 now.

PR #478, opened 2025-06-16
branchprotector: remove protection from excluded branches

### LINKED ISSUE #477: branchprotector: excluded branches retain existing protection instead of being removed
## Issue

When a branch is added to the `exclude` list in branchprotector configuration, the tool correctly stops applying new protection rules but does not remove existing protection from branches that were previously protected.

## Impact

Branches that match exclusion patterns (e.g., `^konflux-`, `^dependabot/`) remain protected on GitHub even though they are excluded in the configuration. This causes push failures when users expect these branches to be unprotected:

```
remote: error: GH006: Protected branch update failed for refs/heads/konflux-oadp-operator-oadp-1-5.
remote: - Changes must be made through a pull request.
! [remote rejected] konflux-oadp-operator-oadp-1-5 -> konflux-oadp-operator-oadp-1-5 (protected branch hook declined)
```

While branchprotector logs show the branch as excluded:
```
"msg":"openshift/oadp-operator=konflux-oadp-operator-oadp-1-5: excluded"
```

## Root Cause

In `cmd/branchprotector/protect.go:341-343`, excluded branches are filtered out early and never added to the updates channel. The protection removal logic at lines 184-188 only triggers when `Request` is set to `nil`, but excluded branches bypass this mechanism entirely.

## Workaround

1. Temporarily remove the branch pattern from the exclude list
2. Let branchprotector run once to detect the branch should be unmanaged (triggering removal)  
3. Add the pattern back to the exclude list

Alternatively, manually remove branch protection via GitHub's web interface or API.

kaovilai:
Fix issue where excluded branches retain existing protection instead of being removed.

When a branch is added to the exclude list in branchprotector configuration, the tool
correctly stops applying new protection rules but does not remove existing protection
from branches that were previously protected.

This change adds logic to detect excluded branches that are currently protected and
queue them for removal by sending requirements with Request: nil, which triggers
RemoveBranchProtection() in the configureBranches() function.

The fix prevents push failures like:
  remote: error: GH006: Protected branch update failed for refs/heads/konflux-branch
  remote: - Changes must be made through a pull request.

Changes:
- Add detection logic for excluded protected branches in UpdateRepo()
- Send removal requests (Request: nil) for such branches
- Update tests to expect removal requests for excluded protected branches
- Add code comment explaining the call flow to RemoveBranchProtection

Fixes kubernetes-sigs/prow#477

### DIFF
--- a/cmd/branchprotector/protect.go
+++ b/cmd/branchprotector/protect.go
@@ -326,11 +326,13 @@ func (p *protector) UpdateRepo(orgName string, repoName string, repo config.Repo
 	}
 
 	branches := map[string]github.Branch{}
+	var allBranches []github.Branch
 	for _, onlyProtected := range []bool{false, true} { // put true second so b.Protected is set correctly
 		bs, err := p.client.GetBranches(orgName, repoName, onlyProtected)
 		if err != nil {
 			return fmt.Errorf("list branches: %w", err)
 		}
+		allBranches = append(allBranches, bs...)
 		for _, b := range bs {
 			_, ok := repo.Branches[b.Name]
 			if !ok && branchInclusions != nil && branchInclusions.MatchString(b.Name) {
@@ -343,6 +345,27 @@ func (p *protector) UpdateRepo(orgName string, repoName string, repo config.Repo
 				continue
 			}
 			branches[b.Name] = b
+		}
+	}
+
+	// Handle excluded branches that are currently protected and need removal
+	if branchExclusions != nil {
+		seen := make(map[string]bool)
+		for _, b := range allBranches {
+			if b.Protected && branchExclusions.MatchString(b.Name) && !seen[b.Name] {
+				seen[b.Name] = true
+				// Check if this branch is not in the branches map (i.e., it was excluded)
+				if _, inBranches := branches[b.Name]; !inBranches {
+					logrus.Infof("%s/%s=%s: removing protection from excluded branch", orgName, repoName, b.Name)
+					p.updates <- requirements{
+						Org:     orgName,
+						Repo:    repoName,
+						Branch:  b.Name,
+						Request: nil, // nil Request triggers RemoveBranchProtection
+						// Flow: updates channel -> configureBranches() -> client.RemoveBranchProtection()
+					}
+				}
+			}
 		}
 	}
 
--- a/cmd/branchprotector/protect_test.go
+++ b/cmd/branchprotector/protect_test.go
@@ -1112,6 +1112,12 @@ branch-protection:
 				{
 					Org:     "kubernetes",
 					Repo:    "test-infra",
+					Branch:  "skip",
+					Request: nil, // nil Request triggers RemoveBranchProtection
+				},
+				{
+					Org:     "kubernetes",
+					Repo:    "test-infra",
 					Branch:  "master",
 					Request: &github.BranchProtectionRequest{EnforceAdmins: &no},
 				},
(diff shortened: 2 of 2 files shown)


cmd/branchprotector/protect.go
```diff
@@ -346,6 +348,27 @@ func (p *protector) UpdateRepo(orgName string, repoName string, repo config.Repo
 		}
 	}
 
+	// Handle excluded branches that are currently protected and need removal
+	if branchExclusions != nil {
+		seen := make(map[string]bool)
```
Prucek, line 353: Why is this map needed?

kaovilai: Problem

  The allBranches slice can contain duplicate entries for the same branch because it's populated from two separate API calls:
```go
  for _, onlyProtected := range []bool{false, true} { // runs twice
      bs, err := p.client.GetBranches(orgName, repoName, onlyProtected)
      allBranches = append(allBranches, bs...) // can add same branch twice
  }
```

  Without Deduplication

  If a branch like konflux-test appears in both API responses, we would send two removal requests:
  1. First iteration: {Org: "org", Repo: "repo", Branch: "konflux-test", Request: nil}
  2. Second iteration: {Org: "org", Repo: "repo", Branch: "konflux-test", Request: nil}

  With seen Map
```go
  if b.Protected && branchExclusions.MatchString(b.Name) && !seen[b.Name] {
      seen[b.Name] = true  // Mark as processed
      // Send removal request only once
  }
```
  The seen map ensures each excluded protected branch gets exactly one removal request, regardless of how many times it appears in allBranches.

  This prevents:
  - Duplicate API calls to GitHub
  - Confusing log messages
  - Potential race conditions in the removal process

kaovilai: Perhaps if p.client.GetBranches(orgName, repoName, false) return all branches we can initialize allBranches once.

Prucek: okay, get it. Thanks!

smg247:
Won't this remove protection from these branches that has been set up outside of prow? Historically, this has not been something that we have wanted Prow to do. I am concerned that this change is overreaching.

kaovilai:
Prow already undid branch protections from my prior repos setups with branchprotector. It overrides native protection settings for all mentioned branches.. hence I had to set up my repos branch protections in prow.. and now that we added exclusion patterns to prow, prow do not remove them. And it's very annoying for working in openshift org because one can't just be admin all the time and undo these settings manually.

smg247:
The concern I have here is that Prow will continue to remove the BP rules when the branch is listed as an exclusion. The problem you are describing requires an admin to go in and change something once, but if we add this logic there will be no possible way to set a BP rule outside of Prow for an excluded branch without Prow continually overwriting it.

kaovilai:
> will be no possible way to set a BP rule outside of Prow

That was already the case for my repo.. we had setup required approvals: 2 outside of prow

Then branchprotector undos them until we put it into the branchprotector yaml config.

kaovilai:
https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1688754465575399
https://github.com/openshift/release/pull/41066

Perhaps it's because we had prow branch-protection configured for `allow_force_pushes` for something else and it just kept overriding everything else not mentioned as false/disabled.

Perhaps prow should only update fields that are explicitly mentioned.. ie. if true do this if false do this, if empty, do nothing.

smg247:
> Perhaps prow should only update fields that are explicitly mentioned.. ie. if true do this if false do this, if empty, do nothing.

IMO, this would be ideal. I am curious of what other's thoughts are on this though.

petr-muller:
Moving towards closing this, rationale in https://github.com/kubernetes-sigs/prow/issues/477#issuecomment-5280228975

/hold

PR #481, opened 2025-06-21
Upgrade Golang v1.24

danilo-gemoli:
Upgrade Golang to `v1.24` and leverage the `go get -tool` feature to get rid of the `hack/tools/tools.go` workaround.

The Golang runtime upgrade hasn't been that hard this time but the new `go get -tool` feature deserves a bit more context.

The `hack/tools/tools.go` used to hold "fake" imports just to make sure the source code of some of the required tools would be vendored.

Let's take `gotestsum` as an example:
```golang
package tools

import (
	_ "gotest.tools/gotestsum"
)
```

A script consuming the `gotestsum` tool is supposed to build it:
```sh
$ cd 'hack/tools'
$ go build -o "${REPO_ROOT}/_bin/gotestsum" gotest.tools/gotestsum
```

 and finally use it:
```sh
$ "${REPO_ROOT}/_bin/gotestsum" ...
```

By introducing the new feature, this procedure is reduced to:
-  `go get -tool gotest.tools/gotestsum@v1.12.3` to install the tool.
-  `go tool gotestsum` for using it.

This new entry will be set in the `go.mod`:
```
tool (
    gotest.tools/gotestsum
)
```


cblecker:
Please note for the future:
When upgrading the minimum golang version, some @kubernetes/test-infra jobs need to be updated or they will fail: https://github.com/kubernetes/test-infra/pull/35019

BenTheElder:
I'm pretty sure we should not have removed the module from hack/tools, the new tools support allows you to avoid needing underscore imports in a tools.go file, but moving these into the main module makes the tools part of the main dependency graph which means anyone using the libraries now has them in their dependency graph.

Tools should be kept in a seperate module.

BenTheElder:
See previously: https://github.com/kubernetes/kubernetes/issues/130825

go.mod
```diff
@@ -195,13 +194,51 @@ require (
 )
 
 require (
```
BenTheElder, line 196: We can see that the indirect require graph grew enourmously, this impacts anyone using the prow API types

BenTheElder:
> A script consuming the gotestsum tool is supposed to build it:
> $ cd 'hack/tools'
> $ go build -o "${REPO_ROOT}/_bin/gotestsum" gotest.tools/gotestsum
> [...]

You can actually instead (without `go tool`)
```console
$ cd hack/tools
$ go install gotest.tools/gotestsum
```

`go install` is module aware and will read the go.mod from the current working module. We use this extensively in Kubernetes/Kubernetes

These scripts _should_ use a local directory though, and not install to the system GOBIN (which both pollutes the host as a side effect and is racy, a second process could reasonably update the binary in the system GOBIN right after we install, but shouldn't touch our local bindir, the package cache will be shared anyhow)

We could use `go tool` to avoid the underscore import but we should avoid putting it in the main module https://kubernetes.slack.com/archives/CHGFYJVAN/p1740498294466519?thread_ts=1740494621.909589&cid=CHGFYJVAN

kei01234kei:
@danilo-gemoli 
You don't need to share tool dependencies with the main module.
Please see: https://github.com/kubernetes/kubernetes/pull/133315

PR #483, opened 2025-06-26
Fix `/retitle` race condition

spowelljr:
Fixes race condition when executing `/retitle` and `/release-note-edit` in the same comment.

**Before:**
Running `/retitle` would retrieve the full PR, modify the title of the local PR object, and then push the PR object up to save it.

This left open a race condition where if `/release-note-edit` executed during the short window where the pull request was retrieved but wasn't saved yet, the release note change would get unintendedly reverted.

**After:**
Update `/retitle` to use `UpdatePullRequest` instead of `EditPullRequest`, `UpdatePullRequest` is a patch operation and we can exclusively update the title, removing the opportunity for the race condition to occur.


petr-muller (approved):
LGTM, thanks!

PR #496, opened 2025-07-22
Tide: Improve error handling when getting PR blockers

danilo-gemoli:
When a new organization in which the GitHub App has not been installed is on-boarded by mistake, Tide stops working entirely, making it impossible to merge any PR from any repository.

A sample error follows:
```
failed getting blockers: error searching for blocker issues: Post \"http://ghproxy/graphql\": failed to get installation id for org QUBIP: the github app is not installed in organization QUBIP
```

With this PR, Tide will report an error for any failing organization and then it will keep moving on.


danilo-gemoli:
/hold
I feel this is incomplete still

danilo-gemoli:
/unhold
good to review now

pkg/tide/blockers/blockers.go
```diff
@@ -222,7 +250,7 @@ type Issue struct {
 	}
 }
 
-type searchQuery struct {
+type SearchQuery struct {
```
petr-muller: I like everything in this PR except for making this type exported, for which the only reason seems to be the type assertion in https://github.com/danilo-gemoli/prow/blob/46c362307b4f99def031fcbb3beb0ec0f413af87/pkg/tide/tide_test.go#L2349 which to me is a rather weak reason to expose the struct.

If we only need that check at that one place in a unit test then I think using reflection is actually a lesser evil.

pkg/tide/tide.go
```diff
@@ -500,7 +501,18 @@ func (c *syncController) Sync() error {
...
+		// When the GitHubApp authentication is enabled, the blocker issues query is split by organization.
+		// Each query might fail individually, and if this happens, we continue with the rest of the work
+		// but we remove all PRs that belong to any of the affected organizations. This is why we don't
+		// want to take into account PRs for which we don't have all the necessary information to process them.
+		if orgBlockersErr := (&blockers.OrgError{}); errors.As(err, &orgBlockersErr) {
+			failedBlockerOrgs := orgBlockersErr.Orgs.UnsortedList()
+			c.logger.Warnf("failed getting blockers for organizations %q - won't merge any PR coming from those", strings.Join(failedBlockerOrgs, ","))
+			maps.DeleteFunc(prs, func(_ string, pr CodeReviewCommon) bool {
+				return orgBlockersErr.Orgs.Has(pr.Org)
+			})
```
petr-muller, line 514: nit: I would slightly prefer if this was a loop that logged each failed org separately in a structured log (`WithField("org", org)`) but not a big deal

pkg/tide/blockers/blockers.go
danilo-gemoli: I have no problem rewriting it, but are we sure we want to go down this route?
```golang
func IsBlockersSearchQueryType(iface any) bool {
    ifaceType := reflect.TypeOf(iface)
    return ifaceType.Kind().String() == "ptr" &&
        ifaceType.Elem().PkgPath() == "sigs.k8s.io/prow/pkg/tide/blockers" &&
        ifaceType.Elem().Name() == "searchQuery"
}
```

pkg/tide/tide.go
danilo-gemoli: IMHO the loop we should be using to log that kind of error is this [one](https://github.com/kubernetes-sigs/prow/blob/46c362307b4f99def031fcbb3beb0ec0f413af87/pkg/tide/blockers/blockers.go#L134) but then, in order to avoid redundancy, at the caller site we should turn this:
```golang
c.logger.Warnf("failed getting blockers for organizations %q - won't merge any PR coming from those", strings.Join(failedBlockerOrgs, ","))
```
into this:
```golang
c.logger.Warn("failed getting blockers for some organizations - won't merge any PR coming from those")
```

pkg/tide/blockers/blockers.go
petr-muller: yeah, if the only consumer that needs this is a single test from a sibling package, then I think an ugly reflection helper is a lesser evil than making a happily-internal-until-now type exported

petr-muller: ...ideally with a comment on that helper that says what I said above :+1:

PR #497, opened 2025-07-22
peribolos: Add repository collaborator management support

hoxhaeris:
# Add repository collaborator support to peribolos

## Summary

This PR adds support for managing repository collaborators through peribolos configuration, enabling declarative management of external repository access alongside existing org members and teams.

## Changes

**Core functionality:**
- Add `collaborators` field to repository configuration in YAML
- Add `--fix-collaborators` flag for opt-in collaborator management (backwards compatible)
- Support for all GitHub permission levels: `read`, `triage`, `write`, `maintain`, `admin`
- Add/update/remove collaborator operations based on configuration

**Implementation details:**
- New `collaboratorClient` interface for collaborator operations
- `configureCollaborators()` function handles collaborator lifecycle management
- Respects pending repository invitations to avoid duplicate invites
- Integrates with existing dump functionality

**Testing:**
- Comprehensive unit tests covering all collaboration scenarios
- Tests for error handling and edge cases
- Integration with existing test patterns

## Example Usage

```yaml
orgs:
  my-org:
    repos:
      my-repo:
        description: "Example repository"
        collaborators:
          external-contributor: write
          contractor: read
          partner-org-user: maintain
```

Command usage:
```bash
# Apply collaborator changes
peribolos --config-path=config.yaml --fix-collaborators --confirm
```

## Backwards Compatibility

- New functionality is opt-in via `--fix-collaborators` flag
- Existing configurations and workflows remain unchanged
- No breaking changes to existing peribolos behavior

## Testing

All functionality is covered by comprehensive unit tests, including:
- Add/update/remove collaborator scenarios
- Error handling and edge cases
- Integration with existing dump and configuration patterns
- Backwards compatibility verification


hoxhaeris:
/hold
I will push a few improvements

hoxhaeris:
During testing, I discovered a fundamental issue with GitHub's REST API.

**Problem:** GitHub's `ListCollaborators` REST endpoint returns ALL users with repository access, including those with inherited permissions through org membership, teams, or default settings. As stated in GitHub's API documentation: *"There is presently not a way to differentiate between an organization level grant and a repository level grant from this endpoint response."*

This caused Peribolos to attempt removing org members who appeared in the collaborators list due to inherited permissions, not explicit grants.

**Solution:** Switched to GraphQL API using `affiliation: DIRECT` parameter, which returns only explicitly added collaborators.

REST API behavior:
```
GET /repos/org/repo/collaborators
Returns: [explicit-collaborator, admin1, admin2, member-with-team-access]
```

GraphQL API behavior:
```graphql
repository { collaborators(affiliation: DIRECT) { ... } }
Returns: [explicit-collaborator]
```
cc @bradmwilliams 
cc @petr-muller

hoxhaeris:
I have tested the following scenarios using a test GitHub organization:
### Core Behavior
- [x] **Permissions matrix transitions** - All permission level changes (Admin/Maintain/Write/Triage/Read)
- [x] **Non-org member flow** - Invitation-based workflow for external users
- [x] **Org member flow** - Direct permission application without invitations
- [x] **Idempotency** - No changes when current state matches desired (includes matching pending invitations)
- [x] **Empty config handling** - Removes all direct collaborators and pending invites; inherited access untouched

### Normalization and Identity
- [x] **Username casing** - Handles case variations across config/current/pending states with correct invite ID mapping
- [x] **Duplicate normalization** - Post-normalization duplicates (Alice/alice) handled correctly; last wins, no double actions

### Inherited vs Direct Access
- [x] **Inherited access overlay** - Direct permissions overlay cleanly on top of inherited org/team access
- [x] **Direct permission removal** - Removing direct permissions doesn't impact inherited access
- [x] **Member removal edge case** - When org member is removed, next reconcile treats as external and re-invites if still in config (expected behavior)

### Pending Invitation Lifecycle
- [x] **Full invitation flow** - Create invite -> permission update via `UpdateRepoInvitation` -> accept -> subsequent updates via `AddCollaborator`
- [x] **Invitation cleanup** - Remove pending invitations not in config using `DeleteRepoInvitation`
- [x] **Conflict resolution** - Handle cases with both pending invite and direct collaborator entry without conflicting actions

### Failure Modes and Fallbacks
- [x] **GraphQL failure handling** - `ListDirectCollaboratorsWithPermissions` failure returns error; does not proceed
- [x] **Invitation listing fallback** - `ListRepoInvitations` failure proceeds with empty pending set; warns appropriately
- [x] **Invitation ID fallback** - `repoInvitationIDs` failure falls back to `AddCollaborator`; warns appropriately
- [x] **API error aggregation** - Add/Remove/Update 4xx/5xx errors are aggregated and surfaced

### Operational Modes
- [x] **Dry-run mode** - `--fix-collaborators + --confirm=false` performs no writes; planned actions logged
- [x] **Feature gating** - `--fix-collaborators=false` skips collaborator reads/writes (except `--dump`)
- [x] **Dump functionality** - Includes collaborators with correct permissions; performs no writes

### Scale and Pagination
- [x] **Pagination support** - Multi-page GraphQL collaborators processed end-to-end *(unit tests only)*
- [x] **Large dataset handling** - Many collaborators/invitations processed with acceptable performance; no missed actions *(unit tests only)*

### Re-entrancy
- [x] **Idempotent reconciliation** - Second run with same desired state makes no API calls; returns nil error

----
cc @petr-muller, cc @bradmwilliams
/unhold

pkg/github/client.go
```diff
@@ -4062,6 +4070,211 @@ func (c *client) ListCollaborators(org, repo string) ([]User, error) {
...
+// ListDirectCollaboratorsWithPermissions gets direct repository collaborators with their permissions using GraphQL.
+// This only returns users who were explicitly added as collaborators, not those with inherited org/team access.
+//
+// See GraphQL schema: repository.collaborators(affiliation: DIRECT)
+func (c *client) ListDirectCollaboratorsWithPermissions(org, repo string) (map[string]RepoPermissionLevel, error) {
+	durationLogger := c.log("ListDirectCollaboratorsWithPermissions", org, repo)
+	defer durationLogger()
+
+	if c.fake {
+		return nil, nil
```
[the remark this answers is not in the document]
hoxhaeris: The fake flag is meant to be a simple no-op, not a delegation to another fake client.  This aligns with the current code patterns.

pkg/github/client.go
```diff
@@ -4889,3 +5102,33 @@ func (c *client) CreatePullRequestReviewComment(org, repo string, number int, rc
...
 }
+
+// ListRepoInvitations returns a list of invitations for the repository.
+//
+// See https://docs.github.com/en/rest/reference/repos#list-repository-invitations
+func (c *client) ListRepoInvitations(org, repo string) ([]RepoInvitation, error) {
+	durationLogger := c.log("ListRepoInvitations", org, repo)
+	defer durationLogger()
+
+	if c.fake {
```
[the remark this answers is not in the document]
hoxhaeris: Same as above. This wouldn't even compile.

cmd/peribolos/main.go
```diff
@@ -369,7 +382,9 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ap
 			AllowRebaseMerge: &full.AllowRebaseMerge,
 			Archived:         &full.Archived,
 			DefaultBranch:    &full.DefaultBranch,
+			Collaborators:    collaborators,
 		})
+		out.Repos[full.Name] = repoConfig
```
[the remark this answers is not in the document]
hoxhaeris: I’d prefer the current approach for readability and maintainability over a minor optimization.

pkg/github/client.go
```diff
@@ -4062,6 +4070,211 @@ func (c *client) ListCollaborators(org, repo string) ([]User, error) {
...
+
+		// Process this page of results
+		for _, edge := range query.Repository.Collaborators.Edges {
+			login := string(edge.Node.Login)
+			permission := mapGraphQLPermissionToRepoLevel(string(edge.Permission))
+			result[login] = permission
+		}
+
+		// Check if there are more pages
+		if !bool(query.Repository.Collaborators.PageInfo.HasNextPage) {
```
bradmwilliams: nit..This bool() conversion is redundant

pkg/github/client.go
```diff
@@ -4062,6 +4070,211 @@ func (c *client) ListCollaborators(org, repo string) ([]User, error) {
...
+			break
+		}
+		vars["after"] = githubql.String(query.Repository.Collaborators.PageInfo.EndCursor)
```
bradmwilliams: nit..The String() conversion is redundant

pkg/github/client.go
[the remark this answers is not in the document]
bradmwilliams: I agree with @hoxhaeris assessment as well.

cmd/peribolos/main.go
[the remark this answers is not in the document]
bradmwilliams: Not a hill I would die on, but seems like a reasonable change to me.

pkg/github/client.go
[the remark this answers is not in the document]
bradmwilliams: Agree with @hoxhaeris

pkg/github/client.go
hoxhaeris: Redundant conversion removed

pkg/github/client.go
hoxhaeris: Redundant conversion removed

bradmwilliams:
I have thoroughly reviewed this PR and think that it is ready for acceptance.  The vast majority of the code is gated by the `--fix-collaborators` flag which will safeguard all pre-existing peribolos users out in the field.  We plan on enabling this functionality, as soon as this PR merges, and are fully committed to supporting it moving forward.  Thanks!

pkg/github/types.go
```diff
@@ -1202,6 +1202,17 @@ type UserRepoInvitation struct {
 	Permission   RepoPermissionLevel `json:"permissions"`
 }
 
+// RepoInvitation contains details about repository invitations returned by list repository invitations endpoint.
+type RepoInvitation struct {
+	ID         int                 `json:"id"`
```
petr-muller: nit: `InvitationID` for consistency with `UserRepoInvitation` struct

pkg/github/types.go
```diff
@@ -1202,6 +1202,17 @@ type UserRepoInvitation struct {
 	Permission   RepoPermissionLevel `json:"permissions"`
 }
 
+// RepoInvitation contains details about repository invitations returned by list repository invitations endpoint.
```
petr-muller: Can we place a link to GH API docs to specify what repo invitations endpoint is meant here? I guess it is this one:

https://docs.github.com/en/rest/collaborators/invitations?apiVersion=2022-11-28#list-repository-invitations

pkg/config/org/org.go
```diff
@@ -73,6 +73,9 @@ type Repo struct {
 
 	Previously []string `json:"previously,omitempty"`
 
+	// Collaborators is a map of username to their permission level for this repository
+	Collaborators map[string]github.RepoPermissionLevel `json:"collaborators,omitempty"`
```
petr-muller, line 77: I am not entirely sure about adding the new member on this level, tbh. Pretty much all other members (xcept `previously` and `on_create`) correspond to the actual payload in https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#update-a-repository ; they are actual properties of a "repo" as a GH entity, but `collaborators` is not.

It seems like we're doing something similar in `Team` where we at least separate `TeamMetadata` from ` Team` but we do not do that here.

Probably not a stopper though, if anything then this should probably be refactored so that synthesized and API-matching members are separated again, but your PR change does not introduce it, so 🤷

cmd/peribolos/main.go
```diff
@@ -357,7 +360,17 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ap
 			return nil, fmt.Errorf("failed to get repo: %w", err)
 		}
 		logrus.WithField("repo", full.FullName).Debug("Recording repo.")
-		out.Repos[full.Name] = org.PruneRepoDefaults(org.Repo{
+
+		// Get direct collaborators (explicitly added) via GraphQL
+		var collaborators map[string]github.RepoPermissionLevel
+		directCollabs, err := client.ListDirectCollaboratorsWithPermissions(orgName, repo.Name)
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to list direct collaborators for %s/%s", orgName, repo.Name)
```
petr-muller: This looks suspicious to me; we hit this error, continue with a warning, have empty `repoConfig.Collaborators` (just as if we succeeded and had empty directCollabs). I _think_ this will result in incorrect reconciliation to empty collaborators later =removal of all existing collaborators?

cmd/peribolos/main.go
```diff
@@ -357,7 +360,17 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ap
...
+		} else if len(directCollabs) > 0 {
+			collaborators = directCollabs
```
petr-muller: nit: I think preparing `repoConfig` earlier with `Collaborators` unset (=zero value) and then setting it conditionally would result in fewer & more tightly scoped variables

```
			if collabs, err := ; if err != nil {
				...
			} else {
				repoConfig.Collaborators = collabs
			}
```

cmd/peribolos/main.go
```diff
@@ -517,11 +532,80 @@ func normalize(s sets.Set[string]) sets.Set[string] {
 	return out
 }
 
+// collaboratorMap manages collaborator usernames to permissions with normalization support
+type collaboratorMap struct {
+	permissions map[string]github.RepoPermissionLevel // normalized_username -> permission
+	origNames   map[string]string                     // normalized_username -> original_username
```
petr-muller: nit: I think I'd slightly prefer a `map[normalized_username]struct{permission, original}` over two maps but no big deal

cmd/peribolos/main.go
```diff
@@ -517,11 +532,80 @@ func normalize(s sets.Set[string]) sets.Set[string] {
...
+func (cm collaboratorMap) has(normalizedUser string) bool {
+	_, exists := cm.permissions[normalizedUser]
+	return exists
+}
+
+// get returns the permission for a normalized username
+func (cm collaboratorMap) get(normalizedUser string) (github.RepoPermissionLevel, bool) {
+	perm, exists := cm.permissions[normalizedUser]
+	return perm, exists
+}
```
petr-muller: I'm not _entirely_ convinced these have value over inline map access expressions (`_, has =`, `permission, has := `) but I have not seen the callsites yet so 🤷

cmd/peribolos/main.go
```diff
@@ -517,11 +532,80 @@ func normalize(s sets.Set[string]) sets.Set[string] {
...
+		return nil, err
+	}
+	for _, i := range is {
+		if i.Invitee == nil || i.Invitee.Login == "" {
+			continue
+		}
+		invitationIDs[github.NormLogin(i.Invitee.Login)] = i.ID
+	}
+	return invitationIDs, nil
+}
```
petr-muller: These two seem to be used together (sequentially) resulting in two API calls resulting in two maps with same keys but different values. I think this should be a single call, value should be a ministruct and the callsite should just extract the value it needs. Or a single API call that returns two maps if the use case really needs them.

cmd/peribolos/main.go
petr-muller: ah okay, this is `dumpOrgConfig`, that's not used for later reconciliation

cmd/peribolos/main.go
hoxhaeris: Merged repoInvitationsWithPermissions and repoInvitationIDs into single repoInvitationsData function (single call)

cmd/peribolos/main.go
hoxhaeris: Not much value here, removed unnecessary helper methods, replaced `has()` and `get()` with direct map access

cmd/peribolos/main.go
hoxhaeris: Changed to use a single map with struct

cmd/peribolos/main.go
hoxhaeris: Agreed, tightened variable scoping

pkg/config/org/org.go
hoxhaeris: You're right, and I agree – this was a pragmatic decision when I started implementing it. Collaborators is a relationship, not a repo property, and indeed it is a separate GitHub API resource. However, as you mentioned, this PR didn’t introduce that pattern.
A future refactor could address this by separating repo metadata from repo relationships, but that would be a much larger change affecting the entire config structure and backward compatibility.

pkg/github/types.go
hoxhaeris: Added documentation link

pkg/github/types.go
hoxhaeris: Renamed the stuct to `CollaboratorRepoInvitation` and `ID` -> `InvitationID`

PR #510, opened 2025-08-21
Fix to properly merge lists when label is configured at global level

jmguzik:
**Problem**: Can not set label backport-risk-assessed: Must be member in one of these teams: [] 

**Solution**: fixed the RestrictedLabelsFor() function in pkg/plugins/config.go to merge restricted label configurations across global (*) -> org -> repo levels instead of overwriting them, which was causing global teams to be lost when repository configs specified only users.

### DIFF
--- a/pkg/plugins/config_test.go
+++ b/pkg/plugins/config_test.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	fuzz "github.com/google/gofuzz"
 
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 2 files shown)


pkg/plugins/config_test.go
```diff
@@ -2416,3 +2417,174 @@ func TestValidatePluginsDupes(t *testing.T) {
...
+		name     string
+		config   Label
+		org      string
+		repo     string
+		expected map[string]RestrictedLabel
+	}{
+		{
+			name: "simple case - no overrides",
+			config: Label{
+				RestrictedLabels: map[string][]RestrictedLabel{
```
petr-muller, line 2432: nit: all tests seem to only use `RestrictedLabels` field so the top-level struct could use that directly instead of `Label`, saving some lines and indent levels

PR #511, opened 2025-08-28
fix: prevent GitHub token leakage in git command logs

ravisantoshgudimetla:
This commit addresses a security vulnerability where GitHub tokens were being exposed in logs when executing git commands. The issue occurred in two places:

1. In executor.go, command arguments containing credentials were being logged without censoring sensitive information
2. In publisher.go, remote URLs with embedded tokens were being logged directly

This fix ensures that sensitive information like GitHub tokens is not exposed in logs, improving the security posture of the application.

### DIFF
--- a/pkg/git/v2/executor.go
+++ b/pkg/git/v2/executor.go
@@ -17,7 +17,9 @@ limitations under the License.
 package git
 
 import (
+	"net/url"
 	"os/exec"
+	"regexp"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -30,6 +32,20 @@ type executor interface {
 
 // Censor censors content to remove secrets
 type Censor func(content []byte) []byte
+
+// censorURLCredentials censors credentials in URLs
+// It replaces the password/token part of URLs with "xxxxx"
+func censorURLCredentials(s string) string {
+	if u, err := url.Parse(s); err == nil && u.User != nil {
+		return u.Redacted()
+	}
+
+	// Fallback to regex for URLs embedded in command arguments
+	// Match URLs like git clone https://username:token@gitlab.com:443/group/project.git
+	// Groups: 1:(https://username:) 2:(token) 3:(@gitlab.com:443)
+	re := regexp.MustCompile(`(https?://[^:]+:)([^@]+)(@[^/\s:]+(?::[0-9]+)?)`)
+	return re.ReplaceAllString(s, "${1}xxxxx${3}")
+}
 
 func NewCensoringExecutor(dir string, censor Censor, logger *logrus.Entry) (executor, error) {
 	g, err := exec.LookPath("git")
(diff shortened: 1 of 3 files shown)


pkg/git/v2/executor.go
```diff
@@ -31,6 +33,30 @@ type executor interface {
...
+			result += "?" + censoredURL.RawQuery
+		}
+		if censoredURL.Fragment != "" {
+			result += "#" + censoredURL.Fragment
+		}
+		return result
+	}
+
+	// Fallback to regex for URLs embedded in command arguments
+	// Match URLs like git clone https://username:token@gitlab.com:443/group/project.git
```
michelle192837, line 44: Really nitpicky: Consider also adding what groups this URL would be broken into (e.g. I assume it should be `1:(https://username:) 2:(token) 3:(@gitlab.com:443/group/project.git)`?)

pkg/git/v2/executor.go
```diff
@@ -31,6 +33,30 @@ type executor interface {
 // Censor censors content to remove secrets
 type Censor func(content []byte) []byte
 
+// censorURLCredentials censors credentials in URLs
+// It replaces the password/token part of URLs with "***"
+func censorURLCredentials(s string) string {
+	if u, err := url.Parse(s); err == nil && u.User != nil {
```
michelle192837, line 39: nit: I was going to ask about something else, but instead, is it easier to use https://pkg.go.dev/net/url#URL.Redacted since you have the parsed URL?

ravisantoshgudimetla: Makes sense, can make the change.

pkg/git/v2/executor.go
```diff
@@ -63,7 +79,14 @@ type censoringExecutor struct {
 }
 
 func (e *censoringExecutor) Run(args ...string) ([]byte, error) {
-	logger := e.logger.WithField("args", strings.Join(args, " "))
+	censoredArgs := make([]string, len(args))
+	for i, arg := range args {
+		censoredArgs[i] = censorURLCredentials(arg)
```
petr-muller, line 84: I wonder if  the existing censoring method (which has an advantage of censoring all secrets possibly known to the process) could be enough here, like:

```suggestion
		censoredArgs[i] = e.censor(arg)
```

(plus the bytes/string conversion)

 Have you tried that?

But the _actual_ benefit of that is likely just hypothetical compared to your fix (_other_ secrets leaked through _other_ arguments, hypothetical) so /shrug

pkg/git/v2/executor.go
```diff
@@ -31,6 +33,20 @@ type executor interface {
...
+// It replaces the password/token part of URLs with "xxxxx"
+func censorURLCredentials(s string) string {
+	if u, err := url.Parse(s); err == nil && u.User != nil {
+		return u.Redacted()
+	}
+
+	// Fallback to regex for URLs embedded in command arguments
+	// Match URLs like git clone https://username:token@gitlab.com:443/group/project.git
+	// Groups: 1:(https://username:) 2:(token) 3:(@gitlab.com:443)
+	re := regexp.MustCompile(`(https?://[^:]+:)([^@]+)(@[^/\s:]+(?::[0-9]+)?)`)
```
petr-muller, line 46: Compilation should be extracted to be done just once, not in every `censorURLCredentials` call

petr-muller (approved):
Let's merge this and fix the minor code issue in a followup

pkg/git/v2/executor.go
petr-muller: https://github.com/kubernetes-sigs/prow/pull/513

PR #519, opened 2025-09-19
Fix cmd that retrieve the version of plank

liangxia:
According to https://github.com/kubernetes/test-infra/blob/1faa9de05541d29656d832b997997c688c273d58/config/prow/cluster/starter/starter-gcs.yaml#L971C10-L971C33 , the label should be `app=prow-controller-manager` ( rather than `app=plank` ).


petr-muller (approved):

petr-muller:
Thanks!

Issue #509, opened 2025-08-20
Improve accessibility: Red text on grey background fails WCAG contrast guidelines


kaovilai:
## Problem

The Prow job status UI displays red text on grey backgrounds that fail WCAG accessibility contrast guidelines. This affects readability, particularly for users with visual impairments.

### Current Colors
- Red text: ~#FF4040
- Grey background: #505050
- **Contrast ratio: 2.32** (fails all WCAG levels - AA requires 4.5:1, AAA requires 7:1)

### Examples
- "Tests Failed" header text in job status pages
- Example URL: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-oadp-operator-oadp-1.5-4.20-e2e-test-kubevirt-aws-periodic/1958016609893224448

## Proposed Solution

To achieve WCAG AA compliance (4.5:1 contrast ratio), we could:

1. **Adjust background to darker grey**: Use ~#141414 background with #242424 cards
2. **Use full red**: #FF0000 on the darker background achieves the required contrast
3. **Alternative**: Use a lighter shade of red like #FFA3A3 (though this approaches pink/white)

## Impact

This affects accessibility for users with:
- Visual impairments
- Color vision deficiencies  
- Users viewing on different devices/lighting conditions

## References

- [WebAIM Contrast Checker](https://webaim.org/resources/contrastchecker/)
- [WCAG 2.1 Guidelines](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html)

> [!Note]
> Responses generated with Claude

kaovilai:
Thread: https://kubernetes.slack.com/archives/CDECRSC5U/p1755707468893709

tsj-30:
Hi @petr-muller, @BenTheElder, @kaovilai 👋
One possible fix is to adjust the colors to meet WCAG AAA contrast.

If we only change the text color, we can switch from #FF4040 → #B00000 (keeping the same background #505050). This achieves a 7:1 contrast ratio.

Alternatively, if we’re open to adjusting the background as well, we could keep a brighter red (e.g., #FF0000) and pair it with a darker grey to hit the required contrast.

This way, we preserve the red/grey theme while ensuring accessibility compliance.

Please let me know what would work the best. 😃

petr-muller:
I do not have a strong opinion. I'm not opposed to adjusting background colors if it ends up working better than just tweaking font colors.

tsj-30:
[...]

PR #520, opened 2025-09-22
Fix accessibility: increase contrast for red-on-gray in Spyglass lens

tsj-30:
This adjusts the CSS styling in the Spyglass component so that red
text on a gray background meets WCAG contrast guidelines (Issue #509).
Update color values and test contrast ratios to ensure readability for
users with low vision or color sensitivity.


michelle192837:
Hi, thanks for the change!

Just from a first look, I think the individual colors might need a bit more adjustment? (2B2B2B vs. FF0000 is better, but still fails some of the accessibility checks on https://webaim.org/resources/contrastchecker/).

(Also if you have a screenshot of how this looks via any method that'd be awesome. Thanks!)

tsj-30:
image

tsj-30:
Hi @michelle192837 ,

I’ve updated the PR to ensure that all red-colored components now meet WCAG contrast guidelines.

- The new red shade has a contrast ratio > 7:1 against the background.

- Verified using the [WebAIM Contrast Checker](https://webaim.org/resources/contrastchecker/?fcolor=FF9999&bcolor=1F1F1F).

- Attached a screenshot showing the updated background and text for reference.

**Exception**: .match-highlighted ["Resolved release" in the attached screenshot] currently has a ratio of 4.64:1, which is still above the WCAG AA minimum (4.5:1) but does not meet AAA.

Please take a look and let me know if you’d like me to explore alternative adjustments for .match-highlighted 🙏.

cmd/deck/static/spyglass/spyglass.css
```diff
@@ -1,5 +1,5 @@
 #announcement {
-  background-color: #424242;
+  background-color: #1f1f1f;
```
michelle192837: Thanks!

nit: I think elevation (e.g. in our case the different greys to show the different layers of elements on the page) is a bit wonky with these changes. It's definitely a readability improvement, but it looks a bit odd.

I think a couple small changes would fix it (`.expander` background-color should be slightly lighter than these background colors (I think `#222222` would work, `#333333` on a stretch), and the `body` background-color (below, currently `#303030` should be darker than any of the other background colors. (`#000000` would be safe, can probably do slightly lighter though like `#111111`).

tsj-30: image
@michelle192837 
This would look something like this, I have added this piece of code and pushed it to https://github.com/kubernetes-sigs/prow/commit/64fb2165f16483ebd790cf41b13a394ab7b0b69c
Please let me know if any more changes are required 😃

michelle192837 (approved):
Awesome, thank you! That looks great :D

tsj-30:
> Awesome, thank you! That looks great :D

Welcome, 😄

petr-muller:
Folks on our Prow noticed today that Spyglass looks slicker, thanks!

tsj-30:
@petr-muller Welcome 😃

PR #524, opened 2025-10-06
Use github DefaultHost Constant Where Possible

midnightconman:
Wherever possible, use the github `DefaultHost` constant instead of a static string.


matthyx:
I suspect you want to override it?
/ok-to-test

midnightconman:
> I suspect you want to override it?

Yeah, we run a local fork against GitHub Enterprise and we override these constants. At the moment we go through and update all these static strings too, but we would prefer to remove all those patches if possible 😸

midnightconman:
@matthyx Tests all fixed, please take another look 😸

matthyx (approved):
/lgtm

PR #528, opened 2025-10-13
Fix: no replacement context should be failure not success

jmguzik:
No replacement context leads to success job state which can result in merging PR with jobs not ran. The fix changes success state to failure in order to address this issue.

### DIFF
--- a/pkg/statusreconciler/migrator/migrator.go
+++ b/pkg/statusreconciler/migrator/migrator.go
@@ -131,21 +131,25 @@ func copyAction(origContext, newContext string) func(statuses []github.Status, s
 // retireAction creates a function that returns a retire action.
 // Specifically the returned function returns a RepoStatus that will update the origContext status
 // to 'success' and set it's description to mark it as retired and replaced by newContext.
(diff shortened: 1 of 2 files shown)


petr-muller:
/hold

status-reconciler sets this for jobs that were removed but their contexts are still on open PRs - the green status mimics the "this job would not run now because it no longer exists". So making it a fail is absolutely not the right action.

petr-muller:
/close

Retiring contexts via a failing status would defeat the purpose of that operation, which is to gracefully remove traces of jobs that were removed (or renamed) from PRs that were open while the removed jobs still existed. Contexts cannot be removed from commits (otherwise retiring would probably do that), so making them green is the next best thing.

I acknowledge the issue where status-reconciler failed to correctly read the job config, which resulted in it seeing **all jobs as removed** and it started retiring all contexts everywhere. This needs to be fixed, status-reconciler should never act on a corrupt / suspiciously empty job config by retiring all existing job results, and if that ever happens then Tide should not respect retired contexts in its merge decisions.

I will file some followup status-reconciler issues to address this.

PR #529, opened 2025-10-15
gangway: add scheduling state

Prucek:
The `scheduling` state of a PJ was added a long time ago, but the `.proto` file was not updated.


pkg/gangway/gangway.proto
```diff
@@ -94,12 +94,13 @@ message JobExecution {
 // in prow/apis/prowjobs/v1/types.go.
 enum JobExecutionStatus {
   JOB_EXECUTION_STATUS_UNSPECIFIED = 0;
-  TRIGGERED = 1;
-  PENDING = 2;
-  SUCCESS = 3;
-  FAILURE = 4;
-  ABORTED = 5;
-  ERROR = 6;
+  SCHEDULING = 1;
```
droslean: To keep backwards compatibility, it would be better to assign the `SCHEDULING` enum to a new value instead of changing them all.

PR #530, opened 2025-10-17
github: enable dry-run mode when using apps auth

hoxhaeris:
## Summary

This PR fixes a limitation where GitHub Apps could not use dry-run mode in Peribolos. Previously, attempting to run peribolos without the `--confirm` flag while using GitHub App authentication would fail with: `not requesting GitHub App access_token in dry-run mode`.
The `allowInDryRun` flag includes mechanical enforcement via hardcoded allowlist to prevent accidental misuse.

## Problem

GitHub Apps authenticate differently than personal access tokens:
1. Apps need to fetch installation tokens via POST request to `/app/installations/{id}/access_tokens`
2. These tokens are required for **all** subsequent API calls, including read-only GET requests
3. The existing dry-run mode blocked **all** non-GET requests, including token fetching
4. This made it impossible to use dry-run mode with GitHub Apps

## Solution

Introduce an `allowInDryRun` flag on the `request` struct to selectively allow specific read-only operations that enable other reads (specifically: GitHub App token acquisition) while keeping all actual mutations blocked.

### Why This Works

- **Token fetching is effectively read-only**: No state changes in org/repos/teams/members
- **Token enables reads**: Required for all API calls, including GETs
- **Mutations still blocked**: All POST/PUT/PATCH/DELETE to resources remain blocked in dry-run mode
- **Time-limited**: Installation tokens expire after 1 hour
- **Matches user expectations**: Dry-run = "see what would change" without making changes

### With GitHub Apps
```bash
# Dry-run mode (preview changes)
./peribolos \
  --github-app-id=2126429 \
  --github-app-private-key-path=/path/to/key.pem \
  --config-path=config.yaml \
  --require-self=false \
  --min-admins=2 \
  --fix-org \
  --fix-org-members \
  --fix-teams
  # No --confirm flag = dry-run mode

# Apply changes
./peribolos \
  --github-app-id=2126429 \
  --github-app-private-key-path=/path/to/key.pem \
  --config-path=config.yaml \
  --require-self=false \
  --min-admins=2 \
  --fix-org \
  --fix-org-members \
  --fix-teams \
  --confirm
```

### DIFF
--- a/pkg/github/client.go
+++ b/pkg/github/client.go
@@ -824,6 +824,12 @@ type request struct {
 	org         string
 	requestBody interface{}
 	exitCodes   []int
+	// allowInDryRun allows this request even in dry-run mode.
+	// WARNING: This should ONLY be used for read-only operations that enable other reads,
+	// such as GitHub App installation token acquisition. NEVER use this for actual mutations
+	// (creating/updating/deleting org members, teams, repos, etc.) as it would defeat the
+	// purpose of dry-run mode. Currently only used for: /app/installations/{id}/access_tokens
+	allowInDryRun bool
 }
 
 type requestError struct {
@@ -923,6 +929,27 @@ func (c *client) requestWithContext(ctx context.Context, r *request, ret interfa
 	return statusCode, nil
 }
 
+// isDryRunAllowed returns true if this request should be allowed in dry-run mode.
+// Enforces a hardcoded allowlist; currently only allows GitHub App token acquisition.
+func isDryRunAllowed(r *request) bool {
+	if !r.allowInDryRun {
+		return false
+	}
+
+	// Hardcoded allowlist: ONLY allow GitHub App token acquisition
+	// Pattern: POST /app/installations/{installation_id}/access_tokens
+	if r.method == http.MethodPost &&
+		strings.Contains(r.path, "/app/installations/") &&
+		strings.HasSuffix(r.path, "/access_tokens") {
+		return true
+	}
+
+	// If allowInDryRun is set but doesn't match the allowlist, this is a bug
+	// Log an error to catch misuse during development
+	logrus.Errorf("SECURITY: allowInDryRun=true set for non-allowed endpoint: %s %s. This is a bug - allowInDryRun should ONLY be used for GitHub App token acquisition.", r.method, r.path)
+	return false
+}
+
 // requestRaw makes a request with retries and returns the response body.
 // Returns an error if the exit code is not one of the provided codes.
 func (c *client) requestRaw(r *request) (int, []byte, error) {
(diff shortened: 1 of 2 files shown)


pkg/github/client.go
```diff
@@ -923,14 +929,36 @@ func (c *client) requestWithContext(ctx context.Context, r *request, ret interfa
...
+	// Pattern: POST /app/installations/{installation_id}/access_tokens
+	if r.method == http.MethodPost &&
+		strings.Contains(r.path, "/app/installations/") &&
+		strings.HasSuffix(r.path, "/access_tokens") {
+		return true
+	}
+
+	// If allowInDryRun is set but doesn't match the allowlist, this is a bug
+	// Log an error to catch misuse during development
+	logrus.Errorf("SECURITY: allowInDryRun=true set for non-allowed endpoint: %s %s. This is a bug - allowInDryRun should ONLY be used for GitHub App token acquisition.", r.method, r.path)
```
petr-muller, line 949: I wonder if this should be `Fatalf` and blow up into developer's face to make it hard to ignore

pkg/github/client.go
```diff
@@ -923,14 +929,36 @@ func (c *client) requestWithContext(ctx context.Context, r *request, ret interfa
 	return statusCode, nil
 }
 
+// isDryRunAllowed returns true if this request should be allowed in dry-run mode.
+// Enforces a hardcoded allowlist; currently only allows GitHub App token acquisition.
+func isDryRunAllowed(r *request) bool {
```
petr-muller, line 934: I'd even move the check for method being `GET` from the L961 condition to this method to centralize the logic

petr-muller (commented):
I like the intent and I especially like the implementation, very elegant. I have two minor points inline, but they are not blocking.

Maybe I'd add something loud (warning?) to be transparent to the user about this happening.

This will affect all clients that use dry-run, not just peribolos:
/retitle github: enable dry-run mode when using apps auth

I'll need to think a little about this though - we definitely want _some_ form of this because it is obviously useful, but the current state is intentional - I remember this was discussed back when Alvaro added apps auth to our client library. I do not remember if the current state was a strong decision or something to be revisited later or something like that. So I'm not sure if the best way to move forward is really to change the `--dry-run` behavior, or if we should add something new (`--dry-run-with-apps`) or something like that. I think I'll talk to some people to see if anyone has concerns about just changing `--dry-run` (I'd personally prefer that) and get back here.

kaovilai (commented):
Works for me.

petr-muller (approved):

petr-muller:
Let's move forward with this one

pkg/github/client.go
petr-muller: This would make the logic hard to to test and the value is not that significant

pkg/github/client.go
petr-muller: I've filed https://github.com/kubernetes-sigs/prow/pull/756 as a followup

PR #532, opened 2025-10-25
Fix Kubernetes link on landing page

stmcginnis:
There is a link to the main k/k repo on the docs landing page. This index file is html, not markdown, so the existing markdown link formatting does not get rendered into the expected html.

This updates the link to use standard html.

image


matthyx:
/approve
thanks @stmcginnis

PR #537, opened 2025-10-30
tide: changes requested block merge when branch protection requires reviews

### LINKED ISSUE #269: PR with "Change requested" leads to Tide repeatedly attempting MERGE

Prucek:
Fixes #269
When a PR has "changes requested", Tide was attempting to merge it even though GitHub blocks it when branch protection requires reviews.
Changes:
 - Added branch protection check in `isAllowedToMerge()` to block merges only when branch protection requires reviews and the PR has "changes requested"
- Updated `requirementDiff()` to respect branch protection settings for status messages
- Added `branchRequiresReviews()` helper with caching to minimize API calls

Result: Tide no longer attempts to merge PRs with "changes requested" when branch protection would block the merge.
🤖 Assisted by Claude.

### DIFF
--- a/pkg/tide/status.go
+++ b/pkg/tide/status.go
@@ -253,11 +253,17 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
 			desc = fmt.Sprintf(" Jobs %s have not succeeded.", strings.Join(trunced, ", "))
 		}
 	}
-
 	if q.ReviewApprovedRequired && pr.ReviewDecision != githubql.PullRequestReviewDecisionApproved {
 		diff += 50
 		if desc == "" {
 			desc = " PullRequest is missing sufficient approving GitHub review(s)"
+		}
+	}
+	// Check GitHub's mergeStateStatus which reflects all GitHub-side blocking conditions
+	if pr.MergeStateStatus == "BLOCKED" {
+		diff += 100
+		if desc == "" {
+			desc = " Blocked by GitHub (branch rulesets or protection)"
 		}
 	}
 	return desc, diff
--- a/pkg/tide/github.go
+++ b/pkg/tide/github.go
@@ -621,6 +621,11 @@ func (m *mergeChecker) isAllowedToMerge(crc *CodeReviewCommon) (string, error) {
 	} else if !allowed {
 		return fmt.Sprintf("Merge type %q disallowed by repo settings", *mergeMethod), nil
 	}
+	// Check GitHub's mergeStateStatus which reflects all GitHub-side merge blocking conditions
+	// including branch protection rules, rulesets, required reviews, status checks, etc.
+	if pr.MergeStateStatus == "BLOCKED" {
+		return "PR is blocked from merging by GitHub (check branch protection, required reviews, or rulesets)", nil
+	}
 	return "", nil
 }
 
--- a/pkg/tide/status_test.go
+++ b/pkg/tide/status_test.go
@@ -870,6 +870,536 @@ func TestExpectedStatus(t *testing.T) {
 	}
 }
 
+func TestRequirementDiff(t *testing.T) {
+	testCases := []struct {
+		name                   string
+		prLabels               []string
+		prAuthor               string
+		prMilestone            *Milestone
+		prBaseBranch           string
+		prContexts             []Context
+		prCheckRuns            []CheckRun
+		reviewDecision         githubql.PullRequestReviewDecision
+		mergeStateStatus       string
+		queryLabels            []string
+		queryForbiddenLabels   []string
+		queryAuthor            string
+		queryMilestone         string
+		queryExcludedBranches  []string
+		queryIncludedBranches  []string
+		reviewApprovedRequired bool
+		expectedDiff           int
+		expectedDescContains   string
+	}{
+		// MergeStateStatus tests
+		{
+			name:                   "Blocked merge state with changes requested",
+			reviewDecision:         githubql.PullRequestReviewDecisionChangesRequested,
+			mergeStateStatus:       "BLOCKED",
+			reviewApprovedRequired: false,
+			expectedDiff:           100,
+			expectedDescContains:   "Blocked by GitHub (branch rulesets or protection)",
+		},
+		{
+			name:                   "Blocked merge state with specific review decision, review required",
+			reviewDecision:         githubql.PullRequestReviewDecisionChangesRequested,
+			mergeStateStatus:       "BLOCKED",
+			reviewApprovedRequired: true,
+			expectedDiff:           150,
+			expectedDescContains:   "PullRequest is missing sufficient approving GitHub review(s)",
+		},
+		{
+			name:                   "Blocked merge state without specific review decision",
+			mergeStateStatus:       "BLOCKED",
+			reviewApprovedRequired: false,
+			expectedDiff:           100,
+			expectedDescContains:   "Blocked by GitHub (branch rulesets or protection)",
+		},
+		{
+			name:                   "Clean merge state allows merge",
+			mergeStateStatus:       "CLEAN",
+			reviewApprovedRequired: false,
+			expectedDiff:           0,
+			expectedDescContains:   "",
+		},
+		{
+			name:                   "Missing approval when required",
+			reviewDecision:         "",
+			reviewApprovedRequired: true,
+			expectedDiff:           50,
+			expectedDescContains:   "missing sufficient approving",
+		},
+		{
+			name:                   "Approved review passes",
+			reviewDecision:         githubql.PullRequestReviewDecisionApproved,
+			reviewApprovedRequired: true,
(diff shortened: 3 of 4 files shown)


pkg/tide/status.go
```diff
@@ -254,7 +254,13 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
 		}
 	}
 
-	if q.ReviewApprovedRequired && pr.ReviewDecision != githubql.PullRequestReviewDecisionApproved {
+	// PullRequestReviewDecisionChangesRequested is similar to not approved
+	if pr.ReviewDecision == githubql.PullRequestReviewDecisionChangesRequested {
```
petr-muller: Even if this was sufficient to fix the issue (which I'm afraid it isn't :( ), would this _always_ be correct to do? I believe that "prevent PRs from merge if there are changes requested" is actually configurable behavior (part of branch protection), so depending on what the repository has configured (and I think also depending on who authored the "request changes" review - maybe that part is propagated to the `pr.ReviewDecision` here though) such review may or may not block merging, but here we'd treat it as if such branch protection was always enabled.

petr-muller (commented):
Some good news and bad news here ;)

Good news: the `requirementDiff` test coverage alone is an excellent contribution and I'd like to get it merged no matter what, likely separately from the actual attempted fix (see the bad news).

Bad news: I do not think this actually resolves the underlying problem. The `requirementDiff` method is used in Tide's status-managing controller (the one that manages the GH status context on the PR commit, including the message shown on the commit) but not in the controller that actually manages mergeability decisions and attempts merge. There are some [notes](https://github.com/kubernetes-sigs/prow/issues/269#issuecomment-2363883843) on this in the linked issue.

pkg/tide/status.go
Prucek: right, understood. So now, when I get that info from branch protection and update the `isAllowedToMerge` that should work?

petr-muller: Yes, but it's complicated under that "get that info from branch protection" sentence. "Will this PR be mergeable per all GitHub-side configuration" is a complicated problem, because branch protection is now actually two kinds of GH-side concepts ([rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets#about-rulesets) -- new thing and [protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#about-branch-protection-rules) -- old thing). And both are not trivial to intepret. Getting info from branch protection would mean you'd need to interpret them, basically reverse engineering GH functionality, in Tide which I believe is not sustainable. It may not be that hard to do for specific merge-blocking setting (like "changes requested" here) but the whole thing would become a whack-a-mole where various GH-side settings prevent PRs from merging, and Tide would need to learn all of them them.

I believe the suitable medium-term approach is to improve how this whole class of failure cases is surfaced. One of the main problems today is that when you hit any case where Tide thinks something is mergeable but GH prohibits that (basically whenever you hit [this place](https://github.com/Prucek/prow/blob/854b8a8908a25c421588852b9587000d36fbb257/pkg/tide/github.go#L290-L291)), it is not surfaced to the user in any way. The user will see something like a green dot saying `In merge pool` and Tide will be logging these "user errors"  in its log.

If we could somehow improve this behavior in general (no matter what the actual reason GH has for disallowing the merge), it would benefit this whole class of failure cases, and we would have more space to experiment with how to make Tide play better with these GH concepts.

Obtaining GH-side information for each merge attempt has its flaws, too; it is API-intensive (although it can be cached, like you are doing in the current revision) and imagine doing these checks for every cause that can prevent the merge.

One additional thought I have is whether GH surfaces the result of its mergeability checking machine _somethow_ through APIs available to Tide. The GitHub UI itself can disable (grey-out) the merge button if the PR does not meet the mergeability criteria - this means GH is able to expose the data at least to its UI. Now the question is whether this is also exposed through the public APIs.

petr-muller:
Conversation from https://github.com/kubernetes-sigs/prow/pull/537#discussion_r2481980666 aside, I am not opposed to trying to improve it for this specific case with the approach you are trying, and I will review the code as it is now.

What kind of testing have you done on this? Have you tried reproducing the failure case from the linked issue and verifying the modified Tide really improves the behavior as expected?

pkg/tide/status.go
petr-muller: Some breadcrumbs on:

> whether GH surfaces the result of its mergeability checking machine somethow through APIs available to Tide

1. There is something called `mergeable_state` in the v3 PR API; we do not seem to use that at all (we use `mergeable` which AFAIK surfaces the git-level mergeability). It is not documented in https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28&versionId=free-pro-team%40latest&category=pulls (at least I have not ofund anything relevant but it is listed in the example API responses)
2. v4 docs have something called `MergeStateStatus` which looks worth looking into: https://docs.github.com/en/graphql/reference/enums#mergestatestatus

Sources: I interrogated Claude Code which then lead me to the follwoing links:

- https://github.com/orgs/community/discussions/24504
- https://github.com/pascalgn/automerge-action/issues/122
- https://stackoverflow.com/questions/30619549/why-does-github-api-return-an-unknown-mergeable-state-in-a-pull-request
- https://github.com/orgs/community/discussions/24299

Prucek:
@petr-muller could you take a look now, please? I created my own tide deployment in my repo: https://github.com/Prucek/GNU_go_sort/pull/20 and it runs this code.

petr-muller:
image

This is awesome

pkg/tide/status.go
```diff
@@ -254,7 +254,13 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
...
 	}
 
-	if q.ReviewApprovedRequired && pr.ReviewDecision != githubql.PullRequestReviewDecisionApproved {
+	// Check GitHub's mergeStateStatus which reflects all GitHub-side blocking conditions
+	if pr.MergeStateStatus == "BLOCKED" {
+		diff += 50
+		if desc == "" {
+			desc = " PR is blocked by GitHub (check branch protection, reviews, or rulesets)"
+		}
+	} else if q.ReviewApprovedRequired && pr.ReviewDecision != githubql.PullRequestReviewDecisionApproved {
```
petr-muller: Two things here:

1. I think the `else` branch is a subset of the `if` branch (we'll see `BLOCKED` if a PR needs an approving review but does not have any), which would mean the `if` branch will prevent us from ever reaching the `else`, even when the `else` case is a better message to show to the user ('review is missing" is better than "GH blocks us"). I think the `BLOCKED` case should be a standalone, last-resort condition: if we see `BLOCKED` and `desc` is empty, set the `desc`.

2. I _think_ we may see `BLOCKED` in some of the otherwise legit not-mergeable states (failing required jobs), so a low `diff` here would make us show this, over a potentially more useful another query (like another query that specifies `q.ReviewApprovedRequired`, which has a diff of `50`). So something like 100 sounds more appropriate?

pkg/tide/status.go
```diff
@@ -254,7 +254,13 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
 		}
 	}
 
-	if q.ReviewApprovedRequired && pr.ReviewDecision != githubql.PullRequestReviewDecisionApproved {
+	// Check GitHub's mergeStateStatus which reflects all GitHub-side blocking conditions
+	if pr.MergeStateStatus == "BLOCKED" {
+		diff += 50
+		if desc == "" {
+			desc = " PR is blocked by GitHub (check branch protection, reviews, or rulesets)"
```
petr-muller:
```suggestion
			desc = " Blocked by GitHub (branch rulesets or protection)"
```

Brief is good. Required reviews is a subset of rulesets/protection. Not many users can actually check rulesets/branch protection, so I would not instruct them to do so.

ameukam:
It was reported this is a possible breaking change blocking PR merge.   
See: https://kubernetes.slack.com/archives/C7J9RP96G/p1765546995447379

PR #538, opened 2025-10-30
tide: remove PR from batch when batch fails

### LINKED ISSUE #474: 2 pr's are not compatible in merge pool
When 2 PRs are in the merge pool that aren't compatible, prow will get stuck. This is buggy behavior and it should be resolved by breaking the tie, merging one, and retesting the other PR.

Prucek:
Fixes: https://github.com/kubernetes-sigs/prow/issues/474

This PR adds the exclusion of the lowest-priority PR when a batch fails.
Previously, when a Batch failed to merge, it didn't account for the possibility of failure; the action in `takeAction` was always `MergeBatch`, regardless of the error. This sometimes led to /hold-ing one of the PRs from the batch, and then it would merge. 
Now, Tide excludes the lowest-priority PR from the failed batch and triggers a smaller batch, breaking the cycle.

🤖 Assisted by Claude.

### DIFF
--- a/pkg/tide/tide.go
+++ b/pkg/tide/tide.go
@@ -934,6 +934,33 @@ func pickHighestPriorityPR(log *logrus.Entry, prs []CodeReviewCommon, cc map[int
 	return false, smallestPR
 }
 
+// pickLowestPriorityPR returns the PR with the highest number (newest, lowest
+// priority in Tide's ordering) from the list. Used to select which PR to
+// exclude when bisecting a failed batch.
+func pickLowestPriorityPR(prs []CodeReviewCommon) (bool, CodeReviewCommon) {
+	if len(prs) == 0 {
+		return false, CodeReviewCommon{}
+	}
+	result := prs[0]
+	for _, pr := range prs[1:] {
+		if pr.Number > result.Number {
+			result = pr
+		}
+	}
+	return true, result
+}
+
+// withoutPR returns a copy of prs excluding the PR with the given number.
+func withoutPR(prs []CodeReviewCommon, excludeNumber int) []CodeReviewCommon {
+	result := make([]CodeReviewCommon, 0, len(prs))
+	for _, pr := range prs {
+		if pr.Number != excludeNumber {
+			result = append(result, pr)
+		}
+	}
+	return result
+}
+
 // accumulateBatch looks at existing batch ProwJobs and, if applicable, returns:
 // * A list of PRs that are part of a batch test that finished successfully
 // * A list of PRs that are part of a batch test that hasn't finished yet but
@@ -943,7 +970,7 @@ func pickHighestPriorityPR(log *logrus.Entry, prs []CodeReviewCommon, cc map[int
 // successBatch, it's possible that these jobs haven't run yet, and in the case
 // we should consider this batch as failed so that takeAction can trigger a new
 // batch.
-func (c *syncController) accumulateBatch(sp subpool) (successBatch []CodeReviewCommon, pendingBatch []CodeReviewCommon) {
+func (c *syncController) accumulateBatch(sp subpool) (successBatch, pendingBatch, failedBatch []CodeReviewCommon) {
 	sp.log.Debug("accumulating PRs for batch testing")
 	prNums := make(map[int]CodeReviewCommon)
 	for _, pr := range sp.prs {
@@ -1032,9 +1059,11 @@ func (c *syncController) accumulateBatch(sp subpool) (successBatch []CodeReviewC
 			pendingBatch = state.prs
 		case successState:
 			successBatch = state.prs
+		case failureState:
+			failedBatch = state.prs
 		}
 	}
-	return successBatch, pendingBatch
+	return successBatch, pendingBatch, failedBatch
 }
 
 // prowJobsFromContexts constructs ProwJob objects from all successful presubmit contexts that include a baseSHA.
@@ -1539,7 +1568,7 @@ func (c *syncController) nonFailedBatchForJobAndRefsExists(jobName string, refs 
 	return len(pjs.Items) > 0
 }
 
-func (c *syncController) takeAction(sp subpool, batchPending, successes, pendings, missings, batchMerges []CodeReviewCommon, missingSerialTests map[int][]config.Presubmit) (Action, []CodeReviewCommon, error) {
+func (c *syncController) takeAction(sp subpool, batchPending, batchFailed, successes, pendings, missings, batchMerges []CodeReviewCommon, missingSerialTests map[int][]config.Presubmit) (Action, []CodeReviewCommon, error) {
 	var merged []CodeReviewCommon
 	var err error
 	defer func() {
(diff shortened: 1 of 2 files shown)


pkg/tide/github.go
```diff
@@ -287,8 +287,9 @@ func (gi *GitHubProvider) mergePRs(sp subpool, prs []CodeReviewCommon, dontUpdat
 			return gi.ghc.Merge(sp.org, sp.repo, pr.Number, ghMergeDetails)
 		})
 		if err != nil {
-			// These are user errors, shouldn't be printed as tide errors
 			log.WithError(err).Debug("Merge failed.")
+			failed = append(failed, pr.Number)
+			errs = append(errs, err)
```
petr-muller: This change looks very suspicious to me. This condition handles an actual failure to merge something that should be mergeable in the first place, otherwise we would not be here (btw, this is the place where Tide hits the annoying errors where it attempts to merge something blocked by incompatible branch protection settings, like #269). That's why the (now removed) "these are user errors" comment was here. Maybe we need to treat this case differently when we are trying to merge a batch, but for non-batch merges I don't see why this behavior would need to change.

I'll need to doublecheck.
/hold

pkg/tide/tide_test.go
```diff
@@ -1959,6 +1959,17 @@ func testTakeAction(clients localgit.Clients, t *testing.T) {
 			action:           Trigger,
 			enableScheduling: true,
 		},
+		{
+			name:        "batch merge fails, falls back to individual PR merge (issue #474)",
```
petr-muller: Isn't this what the "batch merge errors but continues if a PR is unmergeable" test on L1899 is supposed to test?

Prucek: That is a situation where something was merged (merged: 2), but the bug I am trying to solve is that nothing got merged

pkg/tide/github.go
Prucek: I updated the conditions to only `if len(merged) > 0`

pkg/tide/tide.go
```diff
@@ -1439,7 +1439,15 @@ func (c *syncController) takeAction(sp subpool, batchPending, successes, pending
 	// Merge the batch!
 	if len(batchMerges) > 0 {
 		merged, err = c.provider.mergePRs(sp, batchMerges, c.statusUpdate.dontUpdateStatus)
-		return MergeBatch, batchMerges, err
+		// If the entire batch failed to merge (e.g., due to conflicts), fall through
+		// to attempt merging individual PRs to break the tie.
```
petr-muller: #474 is really weird to me and I'm really not sure if the change proposed in this PR actually helps with anything.

Have you managed to reproduce or at least really understand what was actually happening in #474 (Tide logs would really help with that, it's a pity we haven't captured them when the problem was happening).

My thinking is: if we are here, then we have at least one successful batch job of a presubmit that is required for merge. That implies that all pull requests involved must be git-mergeable (not necessarilty GH-mergeable) without conflicts, otherwise the batch job would fail right at the beginning, when clonerefs merges all involved pulls into the base ref. If that happened, we would not have a successful batch job and we would not get here to try merge PRs based on that fact.

Additionally, if at least one PR from the batch is GH-mergeable (and therefore, if the fallback single merge on L1448 has any chance of succeeding) , it **would** merge already. The PRs are attempted to be [merged sequentially](https://github.com/kubernetes-sigs/prow/blob/98a0c9d48d175803e1c616b856f39362011abef3/pkg/tide/github.go#L285-L288) and the code tracks `keepTrying` for merge failures that should only affect that single PR. It [only returns "stop trying"](https://github.com/kubernetes-sigs/prow/blob/98a0c9d48d175803e1c616b856f39362011abef3/pkg/tide/tide.go#L1349-L1361) on "we do not have permissions to merge" errors and "merge commits are forbidden" errors.  Otherwise it continues and tries to merge other PRs. If none of the PRs merged it means that all of them failed to merge individually (and fallback individual merge has no chance of merging) or Tide hit one of the hard errors (and fallback merge has no chance of merging).

This is all really strange to me, because I remember that we actually saw #474 when it was happening. I just do not understand what is the actual failure case from which the user can recover through individual merges.

I think the idea that if we once have a passing batch job that fails to merge all its PRs then Tide gets stuck on this and does not let any other PR to merge (because it will always prefer to act on the passing Batch job it sees) is sound, I just do not understand the conditions that lead to that.

Prucek: I remember looking at the logs, but didn't find anything interesting since it is thousands of lines, mostly tide queries. I am not even sure if any log would show up, as Tide thinks everything is ok.
Here is at least the pool https://prow.ci.openshift.org/tide-history?repo=openshift%2Fdpu-operator&branch=main&pull=405, it kept choosing the trigger_batch action? Or was it still the same batch job?

Prucek: I think it is always choosing the TRIGGER_BATCH action. That would mean, that the `accumulateBatch()` does not return a pending nor a successful batch https://github.com/kubernetes-sigs/prow/blob/main/pkg/tide/tide.go#L944-L946 otherwise the `takeAction()` would choose a different action, right?

Prucek:
It happened again, and now we captured the log properly, and here is the reasoning in the Azure/ARO_HCP repo:
```
     TIME     | BATCH-PENDING          | ACTION        | REASON
     --------------------------------------------------------------------------------
     17:25:31 |                        | WAIT          |
     17:25:31 | [3957, 3962, 3964]     |               |
     17:28:29 |                        | WAIT          |
     17:28:29 | [3957, 3962, 3964]     |               |
     17:31:46 |                        | WAIT          |
     17:31:46 | [3957, 3962, 3964]     |               |
     17:34:39 |                        | WAIT          |
     17:34:39 | [3957, 3962, 3964]     |               |
     17:37:25 |                        | WAIT          |
     17:37:25 | [3957, 3962, 3964]     |               |
     17:40:34 | None                   |               |
     17:40:34 |                        |               | baseimage-generator-images not passing
     17:40:35 |                        | TRIGGER_BATCH |
     17:43:18 |                        | TRIGGER_BATCH |
     17:43:18 | None                   |               |
     17:43:18 |                        |               | baseimage-generator-images not passing
     17:46:07 |                        | WAIT          |
     17:46:07 | [3957, 3962, 3964]     |               |
     17:48:43 |                        | WAIT          |
     17:48:43 | [3957, 3962, 3964]     |               |
     17:51:44 |                        | TRIGGER_BATCH |
     17:51:44 | None                   |               |
     17:51:44 |                        |               | baseimage-generator-images not passing
     17:54:34 |                        | TRIGGER_BATCH |
     17:54:34 | None                   |               |
     17:54:34 |                        |               | baseimage-generator-images not passing
     17:57:21 |                        | WAIT          |
     17:57:21 | [3957, 3962, 3964]

Prucek:
@petr-muller I think this is the place that is causing the issue: https://github.com/kubernetes-sigs/prow/blob/main/pkg/tide/tide.go#L1028-L1037 
The code doens't care about the failure state. I am not sure how to handle it properly. Should I in case of batch failure, default to merging a single PR?

PR #539, opened 2025-10-30
tide: add more metrics

Prucek:
This PR adds more metrics to see how much time tide spends retesting.

🤖  Assisted by Claude.

### DIFF
--- a/pkg/tide/tide.go
+++ b/pkg/tide/tide.go
@@ -216,6 +216,13 @@ var (
 
 		// Per controller
 		syncHeartbeat *prometheus.CounterVec
+
+		// Retesting metrics
+		retests             *prometheus.CounterVec
+		poolMissingPRs      *prometheus.GaugeVec
+		poolPendingPRs      *prometheus.GaugeVec
+		poolSuccessfulPRs   *prometheus.GaugeVec
+		poolBatchPendingPRs *prometheus.GaugeVec
 	}{
 		pooledPRs: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "pooledprs",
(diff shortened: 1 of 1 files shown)


pkg/tide/tide.go
```diff
@@ -280,6 +289,63 @@ var (
...
+		}),
+		retestsByAction: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "tide_retests_by_action_total",
+			Help: "Total number of retests by action type (TRIGGER for serial, TRIGGER_BATCH for batch). Helps identify whether batch or serial testing is causing more retests.",
+		}, []string{
+			"org",
+			"repo",
+			"branch",
+			"action",
+		}),
```
petr-muller: I do not think we need both tide_retests_by_action_total and tide_retests_total, if you need the latter  you can get it from the former by aggregating. We should only have `tide_retests_total` but let it have `action` label.

pkg/tide/tide.go
```diff
@@ -280,6 +289,63 @@ var (
...
+			Name: "tide_batch_failures_total",
+			Help: "Total number of times a batch test completes but PRs move back to missing state (indicating batch failure). This is expensive as all PRs in the batch need retesting.",
+		}, []string{
+			"org",
+			"repo",
+			"branch",
+		}),
+		poolMissingPRs: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "tide_pool_missing_prs",
+			Help: "Number of PRs with missing or failed tests in each pool. High values indicate testing bottlenecks.",
```
petr-muller: What is "missing or failed tests" supposed to mean? A PR is in a merge pool only if it is passing tests.

pkg/tide/tide.go
```diff
@@ -1705,8 +1778,22 @@ func (c *syncController) syncSubpool(sp subpool, blocks []blockers.Blocker) (Poo
...
+	tideMetrics.poolMissingPRs.WithLabelValues(sp.org, sp.repo, sp.branch).Set(float64(len(missings)))
+	tideMetrics.poolPendingPRs.WithLabelValues(sp.org, sp.repo, sp.branch).Set(float64(len(pendings)))
+	tideMetrics.poolSuccessfulPRs.WithLabelValues(sp.org, sp.repo, sp.branch).Set(float64(len(successes)))
+	tideMetrics.poolBatchPendingPRs.WithLabelValues(sp.org, sp.repo, sp.branch).Set(float64(len(batchPending)))
+	if act == Trigger || act == TriggerBatch {
+		tideMetrics.retests.WithLabelValues(sp.org, sp.repo, sp.branch).Add(float64(len(targets)))
+		tideMetrics.retestsByAction.WithLabelValues(sp.org, sp.repo, sp.branch, string(act)).Add(float64(len(targets)))
+	}
+	if len(batchPending) == 0 && len(missings) > 0 && len(pendings) == 0 {
+		if act == Trigger {
```
petr-muller: Can you help me understand why this indicates a batch failure? I _think_ we can be in this state simply because the HEAD moved and now a PR needs to be (maybe individually, not necessarily in a batch) retested?

pkg/tide/tide.go
```diff
@@ -280,6 +289,63 @@ var (
...
+			"action",
+		}),
+		batchFailures: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "tide_batch_failures_total",
+			Help: "Total number of times a batch test completes but PRs move back to missing state (indicating batch failure). This is expensive as all PRs in the batch need retesting.",
+		}, []string{
+			"org",
+			"repo",
+			"branch",
+		}),
```
petr-muller: Does this really need to be a Tide metric? Batch jobs are a separate type of a ProwJob, I guess we already have that number from crier (or plank maybe) in a metric that tracks job results?

Prucek: yes, you are right: [pkg/kube/metrics.go](https://github.com/kubernetes-sigs/prow/blob/main/pkg/kube/metrics.go)

stmcginnis:
Looks like all comments have been address, and the changes look good to me.

/lgtm

PR #548, opened 2025-11-14
Upgraded Tekton Pipeline from v0.61.0 to v1.6.0 LTS

prad9192:
## What
Upgraded Tekton Pipeline from v0.61.0 to v1.6.0 LTS. Upgraded controller-runtime from v0.18.5 to v0.20.1 and k8s.io dependencies from v0.30.1 to v0.32.x. Added integration testing infrastructure for Tekton Pipeline controller.

## Why
Tekton v1.6.0 LTS requires k8s.io v0.32.x dependencies, which forced controller-runtime upgrade to v0.20.1.


pkg/entrypoint/run_test.go
```diff
@@ -147,7 +147,7 @@ exit 3
 }
 trap cleanup SIGINT SIGTERM EXIT
 echo process started
-sleep infinity &
+sleep 9999 &
```
prad9192, line 150: Changed this for macOS compatibility.

pkg/crier/controller.go
```diff
@@ -68,7 +68,7 @@ func New(
 		Named(fmt.Sprintf("crier_%s", reporter.GetName())).
 		For(&prowv1.ProwJob{}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: numWorkers,
-			RateLimiter: workqueue.DefaultControllerRateLimiter()}).
```
prad9192, line 71: Deprecated hence using the new method - https://pkg.go.dev/k8s.io/client-go@v0.32.8/util/workqueue#DefaultTypedControllerRateLimiter

pkg/entrypoint/run_test.go
stmcginnis: _Might_ be better to just do this in its own PR, but this is a pretty small change and just a couple lines, so probably no harm in including it here.

stmcginnis:
Looks fine to me. The only question would be k8s skew version with the newer dependencies, but I think everything should be updated to be within an acceptable range of this anyway.

/lgtm

prad9192:
Hello @stmcginnis 
Could you please take a look when get a chance?

PR #549, opened 2025-11-17
Allow filtering jobs by owner, repo and org

rikatz:
This PR changes Prow handler prowjob.js, allowing new query parameters that can be used to filter the returned jobs by org, repo and owner.

The idea is that instead of returning the full array of jobs running on this Prow instance, just the expected jobs are returned reducing a lot the json size when dealing just with a specific org, or with a big prow instance.

### DIFF
--- a/cmd/deck/main.go
+++ b/cmd/deck/main.go
@@ -786,40 +786,66 @@ func handleProwJobs(ja *jobs.JobAgent, log *logrus.Entry) http.HandlerFunc {
 		setHeadersNoCaching(w)
 		jobs := ja.ProwJobs()
 		omit := r.URL.Query().Get("omit")
+		omitSet := sets.New(strings.Split(omit, ",")...)
 
-		if set := sets.New[string](strings.Split(omit, ",")...); set.Len() > 0 {
-			for i := range jobs {
-				jobs[i].ManagedFields = nil
-				if set.Has(Annotations) {
-					jobs[i].Annotations = nil
+		org := r.URL.Query().Get("org")
+		repo := r.URL.Query().Get("repo")
+		owner := r.URL.Query().Get("owner")
+
+		ownerMatch := func(job prowapi.ProwJob, owner string) bool {
+			if job.Spec.Refs == nil {
+				return false
+			}
+			for _, pull := range job.Spec.Refs.Pulls {
+				if pull.Author == owner {
+					return true
 				}
-				if set.Has(Labels) {
-					jobs[i].Labels = nil
-				}
-				if set.Has(DecorationConfig) {
-					jobs[i].Spec.DecorationConfig = nil
-				}
-				if set.Has(PodSpec) {
-					// when we omit the podspec, we don't set it completely to nil
-					// instead, we set it to a new podspec that just has an empty container for each container that exists in the actual podspec
-					// this is so we can determine how many containers there are for a given prowjob without fetching all of the podspec details
-					// this is necessary for prow/cmd/deck/static/prow/pkg.ts to determine whether the logIcon should link to a log endpoint or to spyglass
-					if jobs[i].Spec.PodSpec != nil {
-						emptyContainers := []coreapi.Container{}
-						for range jobs[i].Spec.PodSpec.Containers {
-							emptyContainers = append(emptyContainers, coreapi.Container{})
-						}
-						jobs[i].Spec.PodSpec = &coreapi.PodSpec{
-							Containers: emptyContainers,
-						}
+			}
+			return false
+		}
+
+		finalJobs := make([]prowapi.ProwJob, 0)
+		for i := range jobs {
+			if org != "" && (jobs[i].Spec.Refs == nil || jobs[i].Spec.Refs.Org != org) {
+				continue
+			}
+			if repo != "" && (jobs[i].Spec.Refs == nil || jobs[i].Spec.Refs.Repo != repo) {
+				continue
+			}
+			if owner != "" && !ownerMatch(jobs[i], owner) {
+				continue
+			}
+			jobs[i].ManagedFields = nil
+			if omitSet.Has(Annotations) {
+				jobs[i].Annotations = nil
+			}
+			if omitSet.Has(Labels) {
+				jobs[i].Labels = nil
+			}
+			if omitSet.Has(DecorationConfig) {
+				jobs[i].Spec.DecorationConfig = nil
+			}
+			if omitSet.Has(PodSpec) {
+				// when we omit the podspec, we don't set it completely to nil
+				// instead, we set it to a new podspec that just has an empty container for each container that exists in the actual podspec
+				// this is so we can determine how many containers there are for a given prowjob without fetching all of the podspec details
+				// this is necessary for prow/cmd/deck/static/prow/pkg.ts to determine whether the logIcon should link to a log endpoint or to spyglass
+				if jobs[i].Spec.PodSpec != nil {
+					emptyContainers := []coreapi.Container{}
+					for range jobs[i].Spec.PodSpec.Containers {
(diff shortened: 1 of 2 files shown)


cmd/deck/main.go
```diff
@@ -785,40 +785,66 @@ func handleProwJobs(ja *jobs.JobAgent, log *logrus.Entry) http.HandlerFunc {
...
-					jobs[i].Labels = nil
-				}
-				if set.Has(DecorationConfig) {
-					jobs[i].Spec.DecorationConfig = nil
+		org := r.URL.Query().Get("org")
+		repo := r.URL.Query().Get("repo")
+		owner := r.URL.Query().Get("owner")
+
+		ownerMatch := func(job prowapi.ProwJob, owner string) bool {
+			if job.Spec.Refs == nil {
```
ivankatliarchuk, line 805: The org/repo filter only checks Spec.Refs. ProwJobs (especially postsubmit/periodic jobs) can have a Spec.ExtraRefs []Refs pointing to additional repos. 

This is particularly impactful for periodic jobs, which typically have Spec.Refs == nil, meaning they are unconditionally excluded when any org or repo filter is applied - even if they test that org's code.

cmd/deck/main_test.go
```diff
@@ -426,6 +427,185 @@ func TestProwJob(t *testing.T) {
...
+			t.Errorf("Error making request: %v", err)
+		}
+
+		rr := httptest.NewRecorder()
+		handler.ServeHTTP(rr, req)
+		if rr.Code != http.StatusOK {
+			t.Errorf("Bad error code: %d", rr.Code)
+		}
+		resp := rr.Result()
+		defer resp.Body.Close()
```
ivankatliarchuk: defer resp.Body.Close() inside a loop. The body won't be closed until TestHandleProwJobsWithFilter returns

The fix is straightforward — just drop defer:

```js
  body, err := io.ReadAll(resp.Body)
  resp.Body.Close()   // closes immediately after reading, each iteration
```

  Or wrap the iteration body in an anonymous function so defer scopes correctly:
```js
  for _, tc := range testCases {
      func() {
          // ...
          resp := rr.Result()
          defer resp.Body.Close()   // now scoped to this func, fires at end of each iteration
          body, err := io.ReadAll(resp.Body)
          // ...
      }()
  }
```

cmd/deck/main.go
```diff
@@ -785,40 +785,66 @@ func handleProwJobs(ja *jobs.JobAgent, log *logrus.Entry) http.HandlerFunc {
 		setHeadersNoCaching(w)
 		jobs := ja.ProwJobs()
 		omit := r.URL.Query().Get("omit")
+		omitSet := sets.New(strings.Split(omit, ",")...)
```
ivankatliarchuk: worth to use `sets.New[string](....)` to match project conventions

cmd/deck/main.go
ivankatliarchuk, line 792: Probably worth to add a doc comment. Something like

```
// handleProwJobs returns an http.HandlerFunc that serves the /prowjobs.js endpoint.
// It accepts the following optional query parameters:
//
//   - omit: comma-separated list of fields to strip from each job in the response.
//   - org: filter jobs to those whose Spec.Refs.Org matches exactly (case-sensitive).
//   - repo: filter jobs to those whose Spec.Refs.Repo matches exactly (case-sensitive).
//   - owner: filter jobs to those with a pull in Spec.Refs.Pulls whose Author matches exactly (case-sensitive)
```

cmd/deck/main.go
ivankatliarchuk:
```
// Refs is the code under test, determined at runtime by Prow itself
  Refs *Refs `json:"refs,omitempty"`

// ExtraRefs are auxiliary repositories that need to be cloned, determined from config
  ExtraRefs []Refs `json:"extra_refs,omitempty"`
```

Two separate fields. The filter only touches one.

How each job type uses them:

  - Presubmit: Prow sets Spec.Refs automatically — it's the PR's repo. ExtraRefs can optionally clone additional repos.
  - Postsubmit: Same — Spec.Refs is the repo being merged into.
  - Periodic: Has no PR and no base repo, so Prow never populates Spec.Refs. If a periodic job needs to clone a repo, the config author puts it in ExtraRefs instead.
  
  So for a periodic job that tests kubernetes/kubernetes (fabricated example):

```
  periodics:
    - name: periodic-e2e-kubernetes
      extra_refs:
        - org: kubernetes
          repo: kubernetes
          base_ref: main
```

  The resulting ProwJob has:
  - Spec.Refs → nil
  - Spec.ExtraRefs[0].Org → "kubernetes"

This is very theoretical, just looking at codebase. PeriodicsMatchingExtraRefs at https://github.com/kubernetes-sigs/prow/blob/6097a43c84021c53de2058ed1668c3be59b84761/pkg/config/jobs.go#L736 exists in this very repo and does exactly what I described - it looks up periodic jobs by org/repo via ExtraRefs, not Refs.

ivankatliarchuk (commented):
Interesting, never knew something like https://prow.k8s.io/prowjobs.js is actually possible.

ExtraRefs not checked — periodic jobs will disappear when an org/repo filter is applied, which could silently return an incomplete picture

If I understand the change correctly, most likely new query params could be documented. The file currently documents running/debugging/rerunning but says nothing about the HTTP API. A new section at the bottom something similar

```
## HTTP API

  ### GET /prowjobs.js

  Returns a JSON object with an `items` array of all active ProwJobs known to Deck.

  **Query parameters**

  | Parameter | Description |
  |-----------|-------------|
  | `omit` | Comma-separated list of fields to omit from each job. Accepted values: `annotations`, `labels`, `decoration_config`, `pod_spec`. When `pod_spec` is omitted, a
  stub with empty containers is returned instead of `null` so callers can still determine container count. |
  | `org` | Filter to jobs whose `spec.refs.org` matches exactly (case-sensitive). Jobs with no `spec.refs` are excluded. |
  | `repo` | Filter to jobs whose `spec.refs.repo` matches exactly (case-sensitive). Can be combined with `org`. |
  | `owner` | Filter to jobs that have a pull request authored by this user in `spec.refs.pulls` (case-sensitive). |

  **Example — fetch only jobs for a specific org, without labels and annotations:**

  GET /prowjobs.js?org=kubernetes&omit=annotations,labels
```

petr-muller:
/remove-lifecycle stale

@rikatz do you plan to address @ivankatliarchuk 's review? The change seems useful to me and I'd like to see it polished and merged.

rikatz:
I totally missed the comments, sorry. Will try to be back to it by EOW

petr-muller:
Appreciate it! No need to hurry, not urgent.

cmd/deck/main.go
rikatz: sure, I did without it mostly because linter was complaining about unnecessary type argument

cmd/deck/main.go
rikatz: good call, copy pasted as I liked the description :)

cmd/deck/main_test.go
rikatz: I'm going to the first one, I agree going with the approach and as this is test we can simply "Close and assert close() error" directly instead of adding an anonymous func

cmd/deck/main.go
rikatz: would it make sense to say "if specs.ref == nil then we check for extraRefs for match"? similarly to the code you posted above but adjusting for a slices.ContainsFunc (or something similar)

rikatz: I mean, I will need to adjust every match of "org, repo, owner" etc but this is doable.

rikatz:
@petr-muller @ivankatliarchuk done, addressed the comments, filtering will also use extraRefs in case refs is null (for periodic jobs)

petr-muller (approved):
Thanks!

PR #555, opened 2025-11-24
`peribolos`: add org roles feature

hoxhaeris:
# Add Organization Roles Support to Peribolos

This PR adds support for managing GitHub organization custom roles in Peribolos, enabling declarative configuration of role assignments to teams and users.

### 1. Configuration Schema (`pkg/config/org/org.go`)
- Added `Roles` field to `Config` struct for declaring role assignments
- Implemented `ValidateRoles()` for pre-flight validation with case-insensitive team/user matching
- Uses `github.NormLogin()` for consistent username normalization

### 2. GitHub API Client (`pkg/github/client.go`)
Added methods for organization role management:
- `ListOrganizationRoles()`, `ListTeamsWithRole()`, `ListUsersWithRole()`
- `AssignOrganizationRoleToTeam()`, `RemoveOrganizationRoleFromTeam()`
- `AssignOrganizationRoleToUser()`, `RemoveOrganizationRoleFromUser()`

All mutating operations respect dry-run mode.

### 3. Main Implementation (`cmd/peribolos/main.go`)
- Added `--fix-org-roles` flag to enable role management (gatekeeper)
- Implemented `configureOrgRoles()` 
- Team name -> slug resolution for correct API calls
- Slug -> team name mapping in dump mode for idempotent configurations
- Roles configured after teams are fully populated

### 4. Test Coverage (`pkg/config/org/org_test.go`, `cmd/peribolos/main_test.go`)
- 12 validation test cases (nested teams, case-insensitivity, duplicates)
- 6 team assignment test cases
- 8 user assignment test cases

## Behavior

- **Only roles declared in config are managed.** Roles absent from the config, including GitHub's built-in predefined roles (`all_repo_*`) and any roles managed outside peribolos, are left untouched. peribolos never removes assignments from a role it was not asked to manage.
- **To remove all assignments from a role, declare it in config with no teams or users.** Assignments are reconciled to the declared state, so an empty role clears its assignments.
- **Indirect (team-derived) user assignments are preserved.** A user who holds a role via team membership is not removed for being absent from a role's `users` list; only direct assignments are reconciled.
- **`--fix-org-roles` requires `--fix-teams`** so role team names can be resolved to slugs from the reconciled team set. A role's user references are validated only when the config declares org membership (`members`/`admins`); when membership is managed elsewhere those references are not validated. Team references are always validated.
- **Dumping roles is best-effort.** If the organization role APIs are unavailable (for example a token without the required scope), roles are omitted from the dump rather than failing the whole dump. Teams excluded via `--ignore-secret-teams`/`--ignore-enterprise-teams` are also excluded from dumped role assignments.

## Example Configuration

```yaml
orgs:
  my-org:
    teams:
      security-team:
        members: [user1, user2]
      compliance-team:
        members: [user3, user4]

    admins:
      - admin-user

    roles:
      security-manager:
        teams: [security-team]
        users: [admin-user]

      billing-manager:
        teams: [compliance-team]
        users: [admin-user]
```

## Usage

```bash
# Validate configuration only
peribolos --config-path=org.yaml

# Apply with dry-run
peribolos --config-path=org.yaml --fix-org-roles --confirm=false

# Apply changes
peribolos --config-path=org.yaml --fix-org-roles --confirm=true

# Dump current organization state
peribolos --dump my-org --github-token-path=token
```

### DIFF
--- a/cmd/peribolos/main.go
+++ b/cmd/peribolos/main.go
@@ -409,9 +409,6 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ig
 		return nil, fmt.Errorf("failed to list organization roles: %w", err)
 	}
 	logrus.Debugf("Found %d organization roles", len(roles))
-	if len(roles) > 0 {
-		out.Roles = make(map[string]org.Role, len(roles))
-	}
 	for _, role := range roles {
 		logrus.WithField("role", role.Name).Debug("Recording organization role.")
 
@@ -447,9 +444,14 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ig
 			userLogins = append(userLogins, user.Login)
 		}
 
-		out.Roles[role.Name] = org.Role{
-			Teams: teamNames,
-			Users: userLogins,
+		if len(teamNames) > 0 || len(userLogins) > 0 {
+			if out.Roles == nil {
+				out.Roles = make(map[string]org.Role)
+			}
+			out.Roles[role.Name] = org.Role{
+				Teams: teamNames,
+				Users: userLogins,
+			}
 		}
 	}
 
--- a/cmd/peribolos/main.go
+++ b/cmd/peribolos/main.go
@@ -59,6 +59,7 @@ type options struct {
 	fixTeamRepos          bool
 	fixRepos              bool
 	fixCollaborators      bool
+	fixOrgRoles           bool
 	ignoreInvitees        bool
 	ignoreSecretTeams     bool
 	ignoreEnterpriseTeams bool
@@ -97,6 +98,7 @@ func (o *options) parseArgs(flags *flag.FlagSet, args []string) error {
 	flags.BoolVar(&o.fixTeamRepos, "fix-team-repos", false, "Add/remove team permissions on repos if set")
 	flags.BoolVar(&o.fixRepos, "fix-repos", false, "Create/update repositories if set")
 	flags.BoolVar(&o.fixCollaborators, "fix-collaborators", false, "Add/remove/update repository collaborators if set")
+	flags.BoolVar(&o.fixOrgRoles, "fix-org-roles", false, "Assign/remove organization roles to teams and users if set")
 	flags.BoolVar(&o.allowRepoArchival, "allow-repo-archival", false, "If set, archiving repos is allowed while updating repos")
 	flags.BoolVar(&o.allowRepoPublish, "allow-repo-publish", false, "If set, making private repos public is allowed while updating repos")
 	flags.StringVar(&o.logLevel, "log-level", logrus.InfoLevel.String(), fmt.Sprintf("Logging level, one of %v", logrus.AllLevels))
@@ -148,6 +150,10 @@ func (o *options) parseArgs(flags *flag.FlagSet, args []string) error {
 
 	if o.fixTeamRepos && !o.fixTeams {
 		return fmt.Errorf("--fix-team-repos requires --fix-teams")
+	}
+
+	if o.fixOrgRoles && !o.fixTeams {
+		return fmt.Errorf("--fix-org-roles requires --fix-teams")
 	}
 
 	return nil
@@ -212,6 +218,9 @@ type dumpClient interface {
 	GetRepo(owner, name string) (github.FullRepo, error)
 	GetRepos(org string, isUser bool) ([]github.Repo, error)
 	ListDirectCollaboratorsWithPermissions(org, repo string) (map[string]github.RepoPermissionLevel, error)
+	ListOrganizationRoles(org string) ([]github.OrganizationRole, error)
+	ListTeamsWithRole(org string, roleID int) ([]github.OrganizationRoleAssignment, error)
+	ListUsersWithRole(org string, roleID int) ([]github.OrganizationRoleAssignment, error)
 	BotUser() (*github.UserData, error)
 }
 
@@ -275,6 +284,7 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ig
 	idMap := map[int]org.Team{} // metadata for a team
 	children := map[int][]int{} // what children does it have
 	var tops []int              // what are the top-level teams
+	slugToName := map[string]string{}
 
 	for _, t := range teams {
 		logger := logrus.WithFields(logrus.Fields{"id": t.ID, "name": t.Name})
@@ -287,6 +297,7 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ig
 			logger.Debug("Ignoring secret team.")
 			continue
 		}
+		slugToName[t.Slug] = t.Name
 		d := t.Description
 		nt := org.Team{
 			TeamMetadata: org.TeamMetadata{
@@ -390,6 +401,53 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ig
 			repoConfig.Collaborators = directCollabs
 		}
 		out.Repos[full.Name] = repoConfig
+	}
+
+	// Dump organization roles
+	roles, err := client.ListOrganizationRoles(orgName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list organization roles: %w", err)
+	}
+	logrus.Debugf("Found %d organization roles", len(roles))
+	if len(roles) > 0 {
+		out.Roles = make(map[string]org.Role, len(roles))
+	}
+	for _, role := range roles {
+		logrus.WithField("role", role.Name).Debug("Recording organization role.")
+
+		// Get teams with this role
+		teamsWithRole, err := client.ListTeamsWithRole(orgName, role.ID)
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to list teams with role %s", role.Name)
+			continue
+		}
+
+		// Get users with this role
+		usersWithRole, err := client.ListUsersWithRole(orgName, role.ID)
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to list users with role %s", role.Name)
+			continue
+		}
+
+		// Build team and user lists
+		var teamSlugs []string
+		for _, team := range teamsWithRole {
+			if name, ok := slugToName[team.Slug]; ok {
+				teamSlugs = append(teamSlugs, name)
+			} else {
+				teamSlugs = append(teamSlugs, team.Slug)
+			}
+		}
+
+		var userLogins []string
+		for _, user := range usersWithRole {
+			userLogins = append(userLogins, user.Login)
+		}
+
+		out.Roles[role.Name] = org.Role{
+			Teams: teamSlugs,
+			Users: userLogins,
+		}
 	}
 
 	return &out, nil
@@ -541,6 +599,8 @@ func configureOrgMembers(opt options, client orgClient, orgName string, orgConfi
 			}
 		} else if om.State == github.StatePending {
 			logrus.Infof("Invited %s to %s as a %s", user, orgName, role)
+			// Track the new invitation so role assignment can skip this user
+			invitees.Insert(github.NormLogin(user))
 		} else {
 			logrus.Infof("Set %s as a %s of %s", user, role, orgName)
 		}
@@ -945,6 +1005,13 @@ func orgFailedInvitations(opt options, client failedInviteClient, orgName string
 }
 
 func configureOrg(opt options, client github.Client, orgName string, orgConfig org.Config) error {
+	// Validate role configuration early (before any API calls) if we're going to configure roles
+	if opt.fixOrgRoles {
+		if err := orgConfig.ValidateRoles(); err != nil {
+			return fmt.Errorf("invalid role configuration: %w", err)
+		}
+	}
+
 	// Ensure that metadata is configured correctly.
 	if !opt.fixOrg {
 		logrus.Infof("Skipping org metadata configuration")
@@ -968,6 +1035,9 @@ func configureOrg(opt options, client github.Client, orgName string, orgConfig o
 	} else if err := configureOrgMembers(opt, client, orgName, orgConfig, invitees, failedInvites); err != nil {
 		return fmt.Errorf("failed to configure %s members: %w", orgName, err)
 	}
+
+	// Note: New invitations sent by configureOrgMembers are tracked in the invitees set,
+	// so role assignment can skip users with pending invitations without an extra API call.
 
 	// Create repositories in the org
 	if !opt.fixRepos {
@@ -1012,6 +1082,14 @@ func configureOrg(opt options, client github.Client, orgName string, orgConfig o
 			return fmt.Errorf("failed to configure %s team %s repos: %w", orgName, name, err)
 		}
 	}
+
+	// Configure organization roles
+	if !opt.fixOrgRoles {
+		logrus.Infof("Skipping organization roles configuration")
+	} else if err := configureOrgRoles(client, orgName, orgConfig, githubTeams, invitees); err != nil {
+		return fmt.Errorf("failed to configure %s organization roles: %w", orgName, err)
+	}
+
 	return nil
 }
 
@@ -1529,6 +1607,220 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
 	}
 
 	return utilerrors.NewAggregate(updateErrors)
+}
+
+type orgRolesClient interface {
+	ListOrganizationRoles(org string) ([]github.OrganizationRole, error)
+	AssignOrganizationRoleToTeam(org, teamSlug string, roleID int) error
+	RemoveOrganizationRoleFromTeam(org, teamSlug string, roleID int) error
+	AssignOrganizationRoleToUser(org, user string, roleID int) error
+	RemoveOrganizationRoleFromUser(org, user string, roleID int) error
+	ListTeamsWithRole(org string, roleID int) ([]github.OrganizationRoleAssignment, error)
+	ListUsersWithRole(org string, roleID int) ([]github.OrganizationRoleAssignment, error)
+}
+
+// configureOrgRoles configures organization roles for teams and users
+func configureOrgRoles(client orgRolesClient, orgName string, orgConfig org.Config, githubTeams map[string]github.Team, invitees sets.Set[string]) error {
+	// Note: Role configuration is validated at the start of configureOrg() before any API calls
+
+	// Get current organization roles from GitHub
+	roles, err := client.ListOrganizationRoles(orgName)
+	if err != nil {
+		return fmt.Errorf("failed to list organization roles: %w", err)
+	}
+
+	if len(roles) == 0 {
+		logrus.Debugf("No organization roles exist in %s", orgName)
+		return nil
+	}
+
+	// Create a map of configured role names (lowercase) for quick lookup
+	configuredRoles := make(map[string]org.Role)
+	for roleName, roleConfig := range orgConfig.Roles {
+		configuredRoles[strings.ToLower(roleName)] = roleConfig
+	}
+
+	unconfiguredCount := len(roles) - len(configuredRoles)
+	logrus.Debugf("Processing %d organization roles (%d configured, %d unconfigured to check for cleanup)",
+		len(roles), len(configuredRoles), unconfiguredCount)
+
+	var allErrors []error
+
+	// Iterate over ALL GitHub roles to handle both configured and unconfigured roles
+	for _, role := range roles {
+		roleNameLower := strings.ToLower(role.Name)
+
+		if roleConfig, isConfigured := configuredRoles[roleNameLower]; isConfigured {
+			// Role is in config - sync to match desired state
+			if err := configureRoleTeamAssignments(client, orgName, role.Name, role.ID, roleConfig.Teams, githubTeams); err != nil {
+				allErrors = append(allErrors, fmt.Errorf("failed to configure team assignments for role %s: %w", role.Name, err))
+			}
+			if err := configureRoleUserAssignments(client, orgName, role.Name, role.ID, roleConfig.Users, invitees); err != nil {
+				allErrors = append(allErrors, fmt.Errorf("failed to configure user assignments for role %s: %w", role.Name, err))
+			}
+		} else {
+			// Role is NOT in config - remove all assignments (clean up orphaned assignments)
+			logrus.Debugf("Role %q not in config, checking for assignments to clean up", role.Name)
+			if err := configureRoleTeamAssignments(client, orgName, role.Name, role.ID, []string{}, githubTeams); err != nil {
+				allErrors = append(allErrors, fmt.Errorf("failed to remove team assignments for unconfigured role %s: %w", role.Name, err))
+			}
+			if err := configureRoleUserAssignments(client, orgName, role.Name, role.ID, []string{}, invitees); err != nil {
+				allErrors = append(allErrors, fmt.Errorf("failed to remove user assignments for unconfigured role %s: %w", role.Name, err))
+			}
+		}
+	}
+
+	// Check if any configured roles don't exist in GitHub
+	for roleName := range orgConfig.Roles {
+		found := false
+		for _, role := range roles {
+			if strings.EqualFold(role.Name, roleName) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("role %q does not exist in organization %s - create the role in GitHub before assigning it", roleName, orgName)
+		}
+	}
+
+	return utilerrors.NewAggregate(allErrors)
+}
+
+// configureRoleTeamAssignments configures team assignments for a specific role
+func configureRoleTeamAssignments(client orgRolesClient, orgName, roleName string, roleID int, wantTeams []string, githubTeams map[string]github.Team) error {
+	// Get current team assignments for this role
+	currentTeams, err := client.ListTeamsWithRole(orgName, roleID)
+	if err != nil {
+		return fmt.Errorf("failed to list teams with role %s: %w", roleName, err)
+	}
+
+	// If we want no teams and have no teams, we're done
+	if len(wantTeams) == 0 && len(currentTeams) == 0 {
+		return nil
+	}
+
+	// Build a map of normalized team name to team slug for the teams we have in config
+	// This allows resolving "MyTeam" (config name) to "my-team" (GitHub slug)
+	normalizedTeams := make(map[string]string)
+	for name, team := range githubTeams {
+		normalizedTeams[strings.ToLower(name)] = team.Slug
+	}
+
+	// Create sets for comparison using slugs
+	wantSet := sets.New[string]()
+	for _, teamName := range wantTeams {
+		// Resolve config team name to slug
+		if slug, ok := normalizedTeams[strings.ToLower(teamName)]; ok {
+			wantSet.Insert(slug)
+		} else {
+			return fmt.Errorf("team %q referenced in role %q could not be resolved to a GitHub team slug - ensure the team exists in your teams configuration and was successfully created", teamName, roleName)
+		}
+	}
+
+	haveSet := sets.New[string]()
+	for _, team := range currentTeams {
+		haveSet.Insert(team.Slug)
+	}
+
+	// Teams to add
+	var errors []error
+	toAdd := wantSet.Difference(haveSet)
+	for teamSlug := range toAdd {
+		if err := client.AssignOrganizationRoleToTeam(orgName, teamSlug, roleID); err != nil {
+			errors = append(errors, fmt.Errorf("failed to assign role %s to team %s: %w", roleName, teamSlug, err))
+			logrus.WithError(err).Warnf("Failed to assign role %s to team %s", roleName, teamSlug)
+		} else {
+			logrus.Infof("Assigned role %s to team %s", roleName, teamSlug)
+		}
+	}
+
+	// Teams to remove
+	toRemove := haveSet.Difference(wantSet)
+	for teamSlug := range toRemove {
+		if err := client.RemoveOrganizationRoleFromTeam(orgName, teamSlug, roleID); err != nil {
+			errors = append(errors, fmt.Errorf("failed to remove role %s from team %s: %w", roleName, teamSlug, err))
+			logrus.WithError(err).Warnf("Failed to remove role %s from team %s", roleName, teamSlug)
+		} else {
+			logrus.Infof("Removed role %s from team %s", roleName, teamSlug)
+		}
+	}
+
+	return utilerrors.NewAggregate(errors)
+}
+
+// configureRoleUserAssignments configures user assignments for a specific role
+func configureRoleUserAssignments(client orgRolesClient, orgName, roleName string, roleID int, wantUsers []string, invitees sets.Set[string]) error {
+	// Get current user assignments for this role
+	currentUsers, err := client.ListUsersWithRole(orgName, roleID)
+	if err != nil {
+		return fmt.Errorf("failed to list users with role %s: %w", roleName, err)
+	}
+
+	// Create maps to preserve original casing while comparing normalized usernames
+	wantMap := make(map[string]string) // normalized -> original
+	for _, user := range wantUsers {
+		wantMap[github.NormLogin(user)] = user
+	}
+
+	// Only consider DIRECT assignments when building haveMap.
+	// Users with "indirect" assignment have the role via team membership and should not be
+	// removed just because they're not in the users list - they keep the role through their team.
+	haveMap := make(map[string]string) // normalized -> original
+	for _, user := range currentUsers {
+		if user.Assignment == "indirect" {
+			logrus.Debugf("Skipping indirect role assignment for user %s (has role via team membership)", user.Login)
+			continue
+		}
+		haveMap[github.NormLogin(user.Login)] = user.Login
+	}
+
+	// If we want no direct users and have no direct users, we're done
+	if len(wantUsers) == 0 && len(haveMap) == 0 {
+		return nil
+	}
+
+	// Create sets for comparison with normalized usernames
+	wantSet := sets.New[string]()
+	for normalized := range wantMap {
+		wantSet.Insert(normalized)
+	}
+	haveSet := sets.New[string]()
+	for normalized := range haveMap {
+		haveSet.Insert(normalized)
+	}
+
+	// Users to add
+	var errors []error
+	toAdd := wantSet.Difference(haveSet)
+	for normalizedUser := range toAdd {
+		originalUser := wantMap[normalizedUser]
+		// Skip users who have pending org invitations - they must accept before we can assign roles
+		if invitees.Has(normalizedUser) {
+			logrus.Infof("Waiting for %s to accept org invitation before assigning role %s", originalUser, roleName)
+			continue
+		}
+		if err := client.AssignOrganizationRoleToUser(orgName, originalUser, roleID); err != nil {
+			errors = append(errors, fmt.Errorf("failed to assign role %s to user %s: %w", roleName, originalUser, err))
+			logrus.WithError(err).Warnf("Failed to assign role %s to user %s", roleName, originalUser)
+		} else {
+			logrus.Infof("Assigned role %s to user %s", roleName, originalUser)
+		}
+	}
+
+	// Users to remove
+	toRemove := haveSet.Difference(wantSet)
+	for normalizedUser := range toRemove {
+		originalUser := haveMap[normalizedUser]
+		if err := client.RemoveOrganizationRoleFromUser(orgName, originalUser, roleID); err != nil {
+			errors = append(errors, fmt.Errorf("failed to remove role %s from user %s: %w", roleName, originalUser, err))
+			logrus.WithError(err).Warnf("Failed to remove role %s from user %s", roleName, originalUser)
+		} else {
+			logrus.Infof("Removed role %s from user %s", roleName, originalUser)
+		}
+	}
+
+	return utilerrors.NewAggregate(errors)
 }
 
 // teamMembersClient can list/remove/update people to a team.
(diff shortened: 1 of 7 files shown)


hoxhaeris:
/hold
Holding while I finish a set of manual tests using a test GitHub org
/cc @petr-muller 
/cc @bradmwilliams

hoxhaeris:
Tests summary: 

## What Was Tested

### Basic Operations
- Assign role to team
- Assign role to user
- Assign role to both team and user simultaneously
- Remove role from team
- Remove role from user

### Cleanup Scenarios
- Role removed from config -> all assignments cleaned up
- Empty `roles: {}` -> orphaned assignments removed
- Omitted `roles` key -> assignments cleaned up

### Case Sensitivity
- Role names matched case-insensitively (`Security_Manager` -> `security_manager`)
- Team references matched case-insensitively
- User references matched case-insensitively

### Idempotency
- Running Peribolos twice produces no mutations on second run

### Flag Gating
- Without `--fix-org-roles` flag -> no role changes made
- Dry-run mode (`--confirm=false`) -> no actual API mutations
- `--fix-org-roles` requires `--fix-teams` -> fails with clear error if missing

### Dump Functionality
- `--dump` output includes `roles:` section
- Dump uses team names (not slugs) for idempotent round-trip

### Validation
- Error for role not defined in GitHub
- Error for team not in config
- Error for user not an org member
- Validation runs before any API mutations (fail-fast)

### Multiple/Complex Assignments
- Team AND different user assigned to same role
- Partial update: add team + remove user in single run
- Admin users can be assigned roles (not just members)

### Pending Invitations
- Users with pending org invitations gracefully skipped with log message

### Indirect Assignments (Team Membership)
- Users who have role via team membership are NOT removed when `users: []`
- Only direct assignments are managed; indirect ones preserved

---

## Indirect Assignment Handling

GitHub's API returns users who have roles through team membership with `"assignment": "indirect"`. Peribolos will:

- Filters out indirect assignments when comparing to config
- Only manages **direct** user assignments
- Team members keep their role through the team even with `users: []`

This prevents accidental removal of team members who inherit roles.

---

cmd/peribolos/main.go
```diff
@@ -1453,6 +1531,220 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
...
+	logrus.Debugf("Processing %d organization roles (%d configured, %d unconfigured to check for cleanup)",
+		len(roles), len(configuredRoles), unconfiguredCount)
+
+	var allErrors []error
+
+	// Iterate over ALL GitHub roles to handle both configured and unconfigured roles
+	for _, role := range roles {
+		roleNameLower := strings.ToLower(role.Name)
+
+		if roleConfig, isConfigured := configuredRoles[roleNameLower]; isConfigured {
```
ivankatliarchuk: the "check if configured roles don't exist in GitHub" loop runs after the main processing loop that already makes API calls:

```
 // First loop: makes API calls for ALL roles
  for _, role := range roles {
      if roleConfig, isConfigured := configuredRoles[roleNameLower]; isConfigured {
          configureRoleTeamAssignments(...)  // API mutation happens here
          configureRoleUserAssignments(...)  // API mutation happens here
      }
  }

  // Second loop: checks validity of configured roles
  for roleName := range orgConfig.Roles {
      if !found {
          return fmt.Errorf("role %q does not exist...")  // Too late!
      }
  }
```

 If config has roles "A" (exists) and "B" (doesn't exist), role A gets fully configured before returning the error. The check should be moved before the mutation loop.

pkg/github/client.go
```diff
@@ -5164,3 +5176,163 @@ func (c *client) ListRepoInvitations(org, repo string) ([]CollaboratorRepoInvita
 	}
 	return ret, nil
 }
+
+// ListOrganizationRoles lists all organization roles
+//
+// https://docs.github.com/en/rest/orgs/organization-roles#list-organization-roles
```
ivankatliarchuk, line 5528: ListOrganizationRoles doesn't paginate. The GitHub API for listing org roles supports pagination. The implementation makes a single request. While org roles are typically few, this is inconsistent with other list methods in the codebase that use `readPaginatedResults`.

cmd/peribolos/main.go
```diff
@@ -496,6 +554,8 @@ func configureOrgMembers(opt options, client orgClient, orgName string, orgConfi
 			}
 		} else if om.State == github.StatePending {
 			logrus.Infof("Invited %s to %s as a %s", user, orgName, role)
+			// Track the new invitation so role assignment can skip this user
+			invitees.Insert(github.NormLogin(user))
```
ivankatliarchuk, line 627: Weak invitee-tracking logic. The invitees set passed to configureOrgRoles combines pre-existing pending invitees (fetched from API) and newly sent invitations. This is correct for the current run, but the comment says "without an extra API call" — this is an API-level optimization that conflates two concerns (pending pre-run invitations vs. just-sent invitations). On subsequent runs, newly sent invitations will show up in the API-fetched set anyway, so this is fine, but it's fragile.

cmd/peribolos/main.go
```diff
@@ -385,6 +396,53 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ap
...
+	}
+	for _, role := range roles {
+		logrus.WithField("role", role.Name).Debug("Recording organization role.")
+
+		// Get teams with this role
+		teamsWithRole, err := client.ListTeamsWithRole(orgName, role.ID)
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to list teams with role %s", role.Name)
+			continue
+		}
```
ivankatliarchuk, line 434: dumpOrgConfig silently continues on role listing errors. Silent omission from a config dump is dangerous. Someone dumping config to backup/restore will end up with an incomplete config. This should either propagate the error or at minimum produce a dump that is clearly incomplete.

cmd/peribolos/main_test.go
```diff
@@ -3970,3 +4059,683 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
 		}
 	}
 }
+
+// Test organization roles functionality - comprehensive tests
+func TestOrganizationRoles(t *testing.T) {
```
ivankatliarchuk: The test only creates structs and checks field counts - it doesn't test any function logic. While comment states
```
// Test organization roles functionality - comprehensive tests
```

cmd/peribolos/main_test.go
```diff
@@ -3970,3 +4059,683 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
...
+			t.Error("security-manager role should exist")
+		}
+		if len(secRole.Teams) != 2 {
+			t.Errorf("Expected 2 teams in security-manager role, got %d", len(secRole.Teams))
+		}
+	})
+}
+
+// Test organization roles fail-fast scenarios
+func TestOrganizationRolesFailFast(t *testing.T) {
```
ivankatliarchuk: the "error messages contain expected information" sub-test checks string matching against hardcoded strings in the test, not against actual function outputs. The test always passes regardless of what the code does:

```
testCases := []struct {
      errorMsg         string  // hardcoded, not from actual function call!
      expectedContains []string
  }{
      {
          errorMsg: "role non-existent-role does not exist...",  // This is not an actual call
          ...
      },
  }
  // Then checks: strings.Contains(tc.errorMsg, expected)
  // This is tautologically true — it just tests the test data itself
```

Additionally, the error message in the test ("check your configuration") doesn't match what the actual code emits ("create the role in GitHub before assigning it"). These tests should be replaced with actual calls to configureOrgRoles or removed.

cmd/peribolos/main_test.go
```diff
@@ -3970,3 +4059,683 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
...
+
+			if tc.expectError && err == nil {
+				t.Error("Expected error but got none")
+			}
+			if !tc.expectError && err != nil {
+				t.Errorf("Unexpected error: %v", err)
+			}
+
+			// Compare slices with nil-safe comparison
+			if !slicesEqual(client.assignedTeamRoles, tc.expectAssigned) {
```
ivankatliarchuk: This could be flaky with multiple assignments.  For the current test cases this happens to work because each only has 1 add or 1 remove, but if tests expand to multiple assignments the non-deterministic map/set iteration order will cause intermittent failures.

cmd/peribolos/main.go
```diff
@@ -1453,6 +1531,220 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
...
+	}
+
+	// Create a map of configured role names (lowercase) for quick lookup
+	configuredRoles := make(map[string]org.Role)
+	for roleName, roleConfig := range orgConfig.Roles {
+		configuredRoles[strings.ToLower(roleName)] = roleConfig
+	}
+
+	unconfiguredCount := len(roles) - len(configuredRoles)
+	logrus.Debugf("Processing %d organization roles (%d configured, %d unconfigured to check for cleanup)",
```
ivankatliarchuk: This assumes all configured roles exist in GitHub. If the config has 3 roles but GitHub has 2, this goes negative. It's only used in a debug log, but it's misleading.

cmd/peribolos/main_test.go
ivankatliarchuk: Additionally slicesEqual reinvents the wheel.  The Go 1.21+ has slices.Equal

pkg/github/client.go
```diff
@@ -5164,3 +5176,163 @@ func (c *client) ListRepoInvitations(org, repo string) ([]CollaboratorRepoInvita
...
+		org:       org,
+		exitCodes: []int{204},
+	}, nil)
+	return err
+}
+
+// ListTeamsWithRole lists all teams assigned to a specific organization role
+//
+// https://docs.github.com/en/rest/orgs/organization-roles#list-teams-assigned-to-an-organization-role
+func (c *client) ListTeamsWithRole(org string, roleID int) ([]OrganizationRoleAssignment, error) {
```
ivankatliarchuk: ListTeamsWithRole and ListUsersWithRole in client.go are nearly identical. The only difference is the URL suffix. Consider extracting a shared listRoleAssignments helper.

cmd/peribolos/main.go
```diff
@@ -147,6 +149,10 @@ func (o *options) parseArgs(flags *flag.FlagSet, args []string) error {
 		return fmt.Errorf("--fix-team-repos requires --fix-teams")
 	}
 
+	if o.fixOrgRoles && !o.fixTeams {
```
ivankatliarchuk, line 155: `--fix-org-roles` requiring `--fix-teams` is too strict when only user (not team) assignments are configured. A user configuring only user role assignments is forced to also enable team management. It could be correct behaviour, but worth to mention.

pkg/github/client.go
```diff
@@ -5164,3 +5176,163 @@ func (c *client) ListRepoInvitations(org, repo string) ([]CollaboratorRepoInvita
...
 }
+
+// ListOrganizationRoles lists all organization roles
+//
+// https://docs.github.com/en/rest/orgs/organization-roles#list-organization-roles
+func (c *client) ListOrganizationRoles(org string) ([]OrganizationRole, error) {
+	c.log("ListOrganizationRoles", org)
+	if c.fake {
+		return nil, nil
+	}
```
ivankatliarchuk, line 5533: since ListOrganizationRoles returns nil, nil in fake mode, and configureOrgRoles early-exits when len(roles) == 0, any test that accidentally uses the real github.Client fake (rather than fakeOrgRolesClient) would pass trivially without exercising any role logic. This is a test-trap worth noting

pkg/github/client.go
```diff
@@ -5164,3 +5176,163 @@ func (c *client) ListRepoInvitations(org, repo string) ([]CollaboratorRepoInvita
...
+	}, nil)
+	return err
+}
+
+// AssignOrganizationRoleToUser assigns an organization role to a user
+//
+// https://docs.github.com/en/rest/orgs/organization-roles#assign-an-organization-role-to-a-user
+func (c *client) AssignOrganizationRoleToUser(org, user string, roleID int) error {
+	c.log("AssignOrganizationRoleToUser", org, user, roleID)
+	if c.dry {
```
ivankatliarchuk: One thing the PR diff does reveal though: the explicit if c.dry { return nil } guards in the mutating methods are actually redundant. requestRawWithContext already handles
   this:

```goo
  if c.fake || (c.dry && r.method != http.MethodGet) {
      return r.exitCodes[0], nil, nil  // short-circuits for PUT/DELETE in dry mode
  }
```

ivankatliarchuk (changes requested):

cmd/peribolos/main.go
hoxhaeris: Fixed - validation now runs before any mutations. `configureOrgRoles` builds a `githubRolesByName` lookup from the API response, validates all configured roles exist against it, and returns an error before entering the mutation loop.

pkg/github/client.go
hoxhaeris: Fixed -- ListOrganizationRoles now uses readPaginatedResults, consistent with other list methods in the codebase.

Thanks, I was misled by the API spec not explicitly listing the parameters, but the actual API response headers (Link: rel="next") prove it.

cmd/peribolos/main.go
hoxhaeris: Agreed it conflates the two sources, but as you noted it works correctly in practice -- on subsequent runs the API-fetched set covers both cases. I would opt to keep the current approach as-is (I can't think on a better option).

cmd/peribolos/main.go
hoxhaeris: Fixed -- `ListTeamsWithRole` and `ListUsersWithRole` errors in `dumpOrgConfig` now return errors instead of logging and continuing. A dump will fail explicitly rather than produce silently incomplete output.

cmd/peribolos/main_test.go
hoxhaeris: Removed

cmd/peribolos/main_test.go
hoxhaeris: Removed -- both the "validate role assignments" and "error messages contain expected information" subtests were tautological and have been deleted. 
The remaining subtests already call `configureOrgRoles` with a fake client and check real error output, so the coverage is there. Renamed the function to `TestConfigureOrgRoles`.

cmd/peribolos/main_test.go
hoxhaeris: Fixed -- replaced the custom slicesEqual with slicesEqualUnordered that sorts both slices before comparing using slices.Equal. All call sites (both team and user tests) now use this helper, eliminating both the ordering issue and the stdlib reimplementation.

cmd/peribolos/main.go
hoxhaeris: Fixed -- unconfiguredCount is now computed by iterating over GitHub roles and counting those not in the configured set, so it's always accurate and non-negative.

pkg/github/client.go
hoxhaeris: Fixed -- extracted a shared listRoleAssignments(org, roleID, entityType) helper. Both ListTeamsWithRole and ListUsersWithRole are now one-line delegations to this helper.

pkg/github/client.go
hoxhaeris: Removed redundant guards.

hoxhaeris:
Thanks for the thorough review @ivankatliarchuk -- all valid points. I've pushed fixes addressing the feedback:                                                                                                                  
                                                                                                                                                                                                                                    
  - Validation now runs before mutations                                                                                                                                                                                          
  - ListOrganizationRoles paginates correctly
  - Dump filters indirect users and secret teams, propagates errors
  - Removed phantom fields from OrganizationRoleAssignment
  - Fixed invitee fetching when only --fix-org-roles is set
  - Added case-insensitive role name collision detection
  - Removed redundant dry-run guards, deduplicated list methods
  - Replaced tautological tests with real ones, added coverage for error paths and indirect filtering
  - Added FakeClient stubs for OrganizationRolesClient

 --fix-org-roles requiring --fix-teams is kept as-is -- the upfront error is clearer than failing mid-run on team resolution.

 /hold

 Holding while I test the changes manually against a test GitHub org

hoxhaeris:
Rebased on latest main and added a small fix: the dump output now excludes organization roles that have no team or user assignments, so predefined GitHub roles no longer clutter the output as empty objects.

cmd/peribolos/main.go
```diff
@@ -1531,6 +1629,203 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
...
+		if err := client.AssignOrganizationRoleToTeam(orgName, teamSlug, roleID); err != nil {
+			errors = append(errors, fmt.Errorf("failed to assign role %s to team %s: %w", roleName, teamSlug, err))
+			logrus.WithError(err).Warnf("Failed to assign role %s to team %s", roleName, teamSlug)
+		} else {
+			logrus.Infof("Assigned role %s to team %s", roleName, teamSlug)
+		}
+	}
+
+	// Teams to remove
+	toRemove := haveSet.Difference(wantSet)
```
cblecker, line 1742: This will remove role assignments from teams that `--ignore-secret-teams` / `--ignore-enterprise-teams` excluded. `haveSet` includes every team GitHub returns for the role, but `wantSet` can only contain teams in `githubTeams`, which `configureTeams` has already filtered. So any ignored team holding a declared role ends up in `toRemove` on every run. And there's no way to declare it in config either, since resolution against `githubTeams` would fail.

The dump side already builds `ignoredTeamSlugs` for this reason (and the comment around L443 implies dump → apply is safe), but the apply path doesn't honor it. Could you thread the ignored slug set through to `configureOrgRoles` (e.g. have `configureTeams` return it, or compute it the same way `dumpOrgConfig` does) and drop those slugs from `haveSet` before diffing? A test for each flag where `ListTeamsWithRole` returns an ignored team and asserts no removal would be good.

cmd/peribolos/main.go
```diff
@@ -1531,6 +1629,203 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
...
+			resolveErrors = append(resolveErrors, fmt.Errorf("team %q referenced in role %q could not be resolved to a GitHub team slug", teamName, roleName))
+		}
+	}
+	if len(resolveErrors) > 0 {
+		return utilerrors.NewAggregate(resolveErrors)
+	}
+
+	haveSet := sets.New[string]()
+	for _, team := range currentTeams {
+		haveSet.Insert(team.Slug)
```
cblecker, line 1726: The user path skips `Assignment == "indirect"`, but the team path inserts every returned team into `haveSet` without checking it. `OrganizationRoleAssignment` documents that `assignment` is populated for teams too. If a child team inherits a role from its parent, I'd expect GitHub to report the child as `indirect`. With only the parent declared, this would try to remove the role from the child on every sync. The dump loop (~L444) has the same gap, so dump → apply would turn an inherited grant into a direct one that survives revoking the parent's role.

Can you confirm what the API returns for child teams? If it's as documented, skip `indirect` here and in the dump (treating `mixed` as direct, like users), and add a team test case with a parent/child pair.

cmd/peribolos/main_test.go
```diff
@@ -2335,6 +2343,136 @@ func TestDumpOrgConfig(t *testing.T) {
 				Repos:   map[string]org.Repo{},
 			},
 		},
+		{
+			name: "dump organization with roles",
```
cblecker, line 2347: This case has no `teams` fixture, so `slugToName` is empty and every assignment goes through the raw-slug fallback. The name mapping at main.go L448-449 is never exercised. The expected output also references `security-team` in roles but has an empty `Teams` map, so it wouldn't pass `ValidateRoles`. Could you add a case with a team whose name differs from its slug (e.g. `Name: "Security Team", Slug: "security-team"`), expect the name in the dumped role, and assert `ValidateRoles()` succeeds on the result? A secret-team variant of the ignore test would be nice too.

cmd/peribolos/main_test.go
```diff
@@ -4417,3 +4587,692 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
...
+
+func (f *fakeOrgRolesClient) ListUsersWithRole(org string, roleID int) ([]github.OrganizationRoleAssignment, error) {
+	if f.listUsersWithRoleErr != nil {
+		return nil, f.listUsersWithRoleErr
+	}
+	return f.usersWithRole[roleID], nil
+}
+
+// Test configureRoleTeamAssignments
+func TestConfigureRoleTeamAssignments(t *testing.T) {
```
cblecker, line 4834: In all of these rows the `githubTeams` key equals the slug, so the tests would still pass if the code used the raw config name instead of resolving it. Could you add rows where the config name differs from the slug (`"Security Team"` → `security-team`), including a differently-cased reference with the assignment already present (expect no add/remove)?

cmd/peribolos/main_test.go
```diff
@@ -4417,3 +4587,692 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
...
+		invitees := sets.Set[string]{}
+
+		err := configureOrgRoles(client, "test-org", orgConfig, githubTeams, invitees)
+		if err != nil {
+			t.Errorf("Unexpected error: %v", err)
+		}
+
+		// security-manager (role ID 1) must not be touched at all.
+		for _, removal := range client.removedTeamRoles {
+			if strings.Contains(removal, "/1") {
```
cblecker, line 4739: nit: `strings.Contains(removal, "/1")` also matches role IDs 10–19. It works with the current fixtures, but an exact `slicesEqualUnordered` check like the rest of the file would be more robust. Also, the `listTeamsErr`/`assignErr`/`removeErr` columns in the table tests are never set. Either drop them or add rows that use them, ideally one asserting that a removal still happens after an assign fails.

cmd/peribolos/main.go
```diff
@@ -949,6 +1029,13 @@ func orgFailedInvitations(opt options, client failedInviteClient, orgName string
 }
 
 func configureOrg(opt options, client github.Client, orgName string, orgConfig org.Config) error {
+	// Validate role configuration early (before any API calls) if we're going to configure roles
+	if opt.fixOrgRoles {
```
cblecker, line 1033: `ValidateRoles` only runs when `--fix-org-roles` is set, but the PR description says a plain `peribolos --config-path=org.yaml` validates the config. As written, broken role references aren't caught until a mutating run. Should this run whenever `len(orgConfig.Roles) > 0`? Otherwise the usage docs should be adjusted.

cmd/peribolos/main.go
```diff
@@ -396,6 +413,67 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ig
 		out.Repos[full.Name] = repoConfig
 	}
 
+	// Dump organization roles. Listing roles requires org-roles read scope;
```
cblecker, line 416: nit: there isn't an "org-roles read" scope. Classic tokens need `admin:org`, and listing role assignees requires org admin. Worth naming the actual requirement so someone chasing the warning knows what to fix.

cmd/peribolos/main.go
```diff
@@ -1531,6 +1629,203 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
...
+		return fmt.Errorf("failed to list users with role %s: %w", roleName, err)
+	}
+
+	// Create maps to preserve original casing while comparing normalized usernames
+	wantMap := make(map[string]string) // normalized -> original
+	for _, user := range wantUsers {
+		wantMap[github.NormLogin(user)] = user
+	}
+
+	// Only consider DIRECT assignments when building haveMap.
```
cblecker, line 1769: nit: the comment says only DIRECT assignments are considered, but the code only skips `indirect`, so `mixed` is reconciled too (correctly, and the test covers it). Could you reword it so nobody "fixes" this to `== "direct"`?

pkg/config/org/org.go
```diff
@@ -122,6 +123,89 @@ type Config struct {
...
+// See https://docs.github.com/en/rest/orgs/organization-roles#assign-an-organization-role-to-a-team
+// See https://docs.github.com/en/rest/orgs/organization-roles#assign-an-organization-role-to-a-user
+type Role struct {
+	// Teams is a list of team names (from config keys) that have this role assigned
+	Teams []string `json:"teams,omitempty"`
+	// Users is a list of usernames that have this role assigned
+	Users []string `json:"users,omitempty"`
+}
+
+// ValidateRoles checks that all teams and users referenced in role assignments exist in the configuration
```
cblecker, line 140: nit: the doc says all referenced users are checked, but users are only validated when `Members`/`Admins` is non-empty. It also doesn't mention the case-insensitive role name collision check. Could the doc describe both?

pkg/config/org/org.go
```diff
@@ -122,6 +123,89 @@ type Config struct {
...
+func (c *Config) ValidateRoles() error {
+	if len(c.Roles) == 0 {
+		return nil
+	}
+
+	// Build a set of all team slugs (including nested teams)
```
cblecker, line 146: nit: these are config team names (lowercased), not slugs. Could you update the comment and rename `teamSlug` → `teamName` below? The name/slug distinction matters later in main.go.

cblecker (changes requested):
Thanks for sticking with this through the earlier rounds. The pre-mutation role validation and the "only manage declared roles" behavior look good. A couple of correctness issues in the team side of role reconciliation (ignored teams and inherited/indirect team assignments) that I think need to be addressed before this merges, plus some test gaps around the name↔slug handling and a few comment fixes inline.

cmd/peribolos/main.go
hoxhaeris: Good catch, fixed in ef9f06e0a. `configureTeams` now returns the set of team slugs it excluded per the ignore flags, and that set is threaded through `configureOrgRoles` into `configureRoleTeamAssignments`, which drops those slugs from `haveSet` before diffing. An ignored team that holds a declared role is now left untouched instead of being removed on every run. Added table-test rows for the ignored-team case, plus an `expectedIgnored` assertion in `TestConfigureTeams` covering both `--ignore-secret-teams` and `--ignore-enterprise-teams`.

cmd/peribolos/main.go
hoxhaeris: Confirmed against the live API and fixed in ef9f06e0a. Assigning a role to a parent team only, then listing `organization-roles/{id}/teams`, returns the child with `assignment: "indirect"` (and `"mixed"` if the child is also assigned directly), exactly as documented. So the team path now mirrors the user path: it skips `indirect` in both the apply path and the dump loop, and treats `mixed` as direct. This stops an inherited child grant from being stripped on every sync, and stops dump from promoting an inherited grant into a direct one. Added apply-path rows for indirect/mixed and a dump case that excludes an inherited child.

cmd/peribolos/main_test.go
hoxhaeris: Done in ef9f06e0a. Added a dump case with a team whose name differs from its slug (`Name: "Security Team"`, `Slug: "security-team"`), asserting the config name is what lands in the dumped role and that `ValidateRoles()` succeeds on the result. Also added a secret-team variant of the ignore case.

cmd/peribolos/main_test.go
hoxhaeris: Done in ef9f06e0a. Added rows to `TestConfigureRoleTeamAssignments` where the config name differs from the slug (`"Security Team"` -> `security-team`), including a differently-cased reference to an already-present assignment that expects no add and no remove.

cmd/peribolos/main_test.go
hoxhaeris: Done in ef9f06e0a. Replaced the `strings.Contains(removal, "/1")` checks with exact `slicesEqualUnordered` assertions on the recorded call sets. The `listTeamsErr`/`assignErr`/`removeErr` columns are now exercised, including a row asserting that a removal still happens after an assign fails.

cmd/peribolos/main.go
hoxhaeris: Went with adjusting the docs rather than broadening when validation fires. Like the other subsystems, role config is only validated when its fix flag (`--fix-org-roles`) is set, so I have corrected the PR description. Note that validation runs regardless of `--confirm`, so a dry run with `--fix-org-roles` still surfaces broken references before anything is applied. I added a comment at the call site spelling this out.

cmd/peribolos/main.go
hoxhaeris: Fixed in ef9f06e0a. Reworded to say listing roles and their assignees requires an org admin token (classic tokens: `admin:org`).

cmd/peribolos/main.go
hoxhaeris: Fixed in ef9f06e0a. Reworded to say only `indirect` is skipped and that `mixed` is reconciled like a direct assignment, with an explicit note not to change it to `== "direct"`.

pkg/config/org/org.go
hoxhaeris: Fixed in ef9f06e0a. The doc now describes all three checks: the case-insensitive role name collision check, team reference validation, and that user validation only runs when the config declares membership (Members/Admins).

pkg/config/org/org.go
hoxhaeris: Fixed in ef9f06e0a. Renamed `teamSlug` to `teamName` and corrected the comment to say these are config team names, not slugs.

hoxhaeris:
Thanks for the thorough review @cblecker. All points are addressed in ef9f06e0a, with per-thread replies inline.

The two correctness issues are fixed:
- Ignored teams (`--ignore-secret-teams`/`--ignore-enterprise-teams`) are no longer stripped of role assignments. `configureTeams` returns the excluded slugs, which are dropped from the current-state set before diffing.
- Indirect (inherited) team assignments are skipped in both the apply and dump paths, mirroring the user path; `mixed` is treated as direct.

I also confirmed the API behavior you asked about before implementing: assigning a role to a parent team only, then listing the role's teams, returns the child with `assignment: "indirect"` (and `"mixed"` when the child is also assigned directly), matching the docs. I verified the apply path end to end against a live test org (child and secret teams retain their roles across a `--confirm` run, and normal add/remove still works), and the dump path is covered by new unit tests including a `ValidateRoles` round-trip.

I have also corrected the PR description to reflect that role config is validated under `--fix-org-roles` (it runs in dry-run too), rather than on a plain config run.

cmd/peribolos/main.go
```diff
@@ -1531,6 +1649,222 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
...
+	// If we want no teams and have no teams, we're done
+	if len(wantTeams) == 0 && len(currentTeams) == 0 {
+		return nil
+	}
+
+	// Build a map of normalized team name to team slug for the teams we have in config
+	// This allows resolving "MyTeam" (config name) to "my-team" (GitHub slug)
+	normalizedTeams := make(map[string]string)
+	for name, team := range githubTeams {
+		normalizedTeams[strings.ToLower(name)] = team.Slug
```
cblecker: This will strip the role from a team that's being renamed via `previously`. `configureTeams` stores the pre-rename team (`matches[name] = *t`), and `configureTeam` discards the result of `EditTeam`, so by the time roles run `githubTeams[name].Slug` is still the old slug. `ListTeamsWithRole` returns the new slug, so the new slug lands in `toRemove` and gets removed, while the add goes to the old slug. That's a 404 if the old slug is gone, or a no-op if GitHub redirects, and the role ends up removed either way. The dry run skips `EditTeam`, so the preview looks clean.

Could you reconcile by team ID instead (`OrganizationRoleAssignment.ID` is stable across renames, and you can map ID → current slug from the `ListTeamsWithRole` results for the remove calls)? Alternatively, write the `EditTeam` result back into `githubTeams`. A test row where `githubTeams` has `"New Name": {ID: 10, Slug: "old-name"}` and the current assignment is `{ID: 10, Slug: "new-name", Assignment: "direct"}`, expecting no add and no remove, would cover it.

cmd/peribolos/main_test.go
```diff
@@ -2505,6 +2836,27 @@ func (c fakeDumpClient) ListRepoInvitations(org, repo string) ([]github.Collabor
 	return []github.CollaboratorRepoInvitation{}, nil
 }
 
+func (c fakeDumpClient) ListOrganizationRoles(org string) ([]github.OrganizationRole, error) {
```
cblecker: The PR documents dump as best-effort when the role APIs are unavailable, but nothing tests that. `fakeDumpClient`'s role methods can't inject errors. Could you add error fields here and dump cases for: `ListOrganizationRoles` failing (dump succeeds, no `Roles`); `ListTeamsWithRole` failing for one role (that role is skipped, others still dumped); and a user with `Assignment: "indirect"` next to a direct user (only the direct one is emitted)? The team-side indirect filter has a case but the user side doesn't.

cmd/peribolos/main_test.go
```diff
@@ -4417,3 +4774,809 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
...
+			roles: []github.OrganizationRole{
+				{ID: 1, Name: "other-role"},
+			},
+		}
+
+		githubTeams := map[string]github.Team{}
+		invitees := sets.Set[string]{}
+
+		err := configureOrgRoles(client, "test-org", orgConfig, githubTeams, nil, invitees)
+		if err == nil {
```
cblecker: nit: this should be `t.Fatal`. If the error ever comes back nil, the `err.Error()` below panics instead of failing cleanly.

cmd/peribolos/main_test.go
```diff
@@ -4417,3 +4774,809 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
...
+
+	err := configureOrgRoles(client, "test-org", orgConfig, map[string]github.Team{}, nil, sets.Set[string]{})
+	if err == nil {
+		t.Fatal("Expected error for non-existent role")
+	}
+	if !strings.Contains(err.Error(), "not-exists") {
+		t.Errorf("Expected error about 'not-exists', got: %v", err)
+	}
+
+	// Verify NO mutations were made (validation should have blocked them)
```
cblecker: This only checks assignments. It'd be good to cover removals too, but asserting `removedUserRoles` is empty won't prove anything with this fixture since role 1 has no current holders. Could you seed an undeclared direct holder (e.g. `usersWithRole: {1: {{Login: "stale", Assignment: "direct"}}}`) and then assert nothing was removed? Removal is the dangerous side of this feature.

cmd/peribolos/main.go
```diff
@@ -1531,6 +1649,222 @@ func configureTeamRepos(client teamRepoClient, githubTeams map[string]github.Tea
...
+	}
+
+	logrus.Debugf("Processing %d configured organization roles (of %d total in org)", len(orgConfig.Roles), len(roles))
+
+	var allErrors []error
+
+	// Only manage roles that are explicitly declared in config. Roles absent
+	// from config - including GitHub's built-in predefined roles and any roles
+	// managed out-of-band - are left untouched. To empty a role, declare it in
+	// config with no teams/users. (This mirrors how peribolos avoids mutating
```
cblecker: nit: this doesn't quite match how peribolos behaves elsewhere. `configureTeams` deletes undeclared teams and `configureOrgMembers` removes undeclared members. Leaving undeclared roles alone is a deliberate difference, which is fine, but the comment should say so, maybe something like "Unlike teams, roles are opt-in per role because GitHub ships predefined roles and roles may be managed out-of-band."

pkg/config/org/org.go
```diff
@@ -122,6 +123,96 @@ type Config struct {
 	Members []string        `json:"members,omitempty"`
 	Admins  []string        `json:"admins,omitempty"`
 	Repos   map[string]Repo `json:"repos,omitempty"`
+	Roles   map[string]Role `json:"roles,omitempty"`
+}
+
+// Role declares an organization role and its assignments to teams and users
```
cblecker: nit: since this is where people read the schema, could the `Role` doc cover the semantics? Something like: the role must already exist in GitHub (declaring it doesn't create it), only roles listed here are managed, an empty `teams`/`users` removes all direct assignments, and indirect (inherited) assignments are never removed.

cblecker (changes requested):
Thanks for the thorough turnaround on the last round. I've resolved the threads that are fully addressed and replied on the rest. One blocking issue below (team renames via `previously` stripping role assignments) plus some smaller test/doc items.

Separately: `site/content/en/docs/components/cli-tools/peribolos.md` doesn't mention `roles:` at all. Since role validation now only runs with `--fix-org-roles`, and the semantics (roles must already exist, only declared roles are managed, an empty role clears direct assignments, indirect assignments are kept, `--fix-org-roles` requires `--fix-teams`, ignored teams are left alone) only live in the PR description right now, could you add a `roles:` stanza to the example config and a short summary of those rules?

cmd/peribolos/main.go
cblecker: Thanks, the docs route works for me. One follow-up on the new comment: "a dry run surfaces broken references before applying" is a bit broader than what `ValidateRoles` checks. It only covers team/user references in the config. A misspelled role name isn't caught until `configureOrgRoles`, which runs after meta, members, repos, collaborators and teams have all been applied. Could you narrow the wording to "undefined team/user references", or move the role-existence check up front?

pkg/config/org/org.go
cblecker: Doc looks good now. Small follow-up: the membership rationale is now duplicated almost verbatim in the inline comment at L187–192, and both copies name peribolos CLI flags from inside `pkg/config/org`. Could you keep it once in the func doc, phrased in config terms ("membership may be managed elsewhere"), and trim the inline one?

cmd/peribolos/main.go
cblecker: Fix looks good, thanks. One small thing: if someone empties a role in config to revoke it, a grant on an ignored team silently survives and is only logged at Debug. Could that log be Warn (naming role + slug) so it's visible at default log level?

cmd/peribolos/main_test.go
cblecker: Thanks, the new cases look good. Related nit: the final `else` in the dump loop (main.go L461) still emits a raw slug silently when a team shows up in `ListTeamsWithRole` but not `ListTeams` (a race, or a team the token can't see). That produces exactly the non-round-tripping config the comment above it warns about. Could it log a Warn with role + slug, or skip the entry?

PR #556, opened 2025-11-25
plugins: add new plugin for linking and unlinking issues to a PR

### LINKED ISSUE #359: Command for linking a PR to an issue
It would be nice to have a command for [linking a PR to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue).
Something like `/issue 42`.

Amulyam24:
This PR adds a new plugin issue-management which has commands for linking and unlinking issues to a PR.
Ref - https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

- The commands can be used to link an issue to a PR in the current repository or in a different repository as well as handle multiple issues
- This is done by adding the supported keyword `Fixes` to the body of the PR if it doesn't already exist or by appending the issue to the existing Fixes line
- Supported formats are `issue-number` and `org/repo-name#issue-number`

A new plugin has been added to accommodate any existing issue commands or provide flexibility to support more issue commands in the future.

Fixes #359

### DIFF
--- a/pkg/plugins/issue-management/issue_management.go
+++ b/pkg/plugins/issue-management/issue_management.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package issuemanagement implements issue management commands.
+package issuemanagement
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/prow/pkg/config"
+	"sigs.k8s.io/prow/pkg/github"
+	"sigs.k8s.io/prow/pkg/pluginhelp"
+	"sigs.k8s.io/prow/pkg/plugins"
+)
+
+const pluginName = "issue-management"
+
+var (
+	linkIssueRegex   = regexp.MustCompile(`(?mi)^/link-issue\s+(.+)$`)
+	unlinkIssueRegex = regexp.MustCompile(`(?mi)^/unlink-issue\s+(.+)$`)
+)
+
+type githubClient interface {
+	CreateComment(org, repo string, number int, comment string) error
+	GetIssue(org, repo string, number int) (*github.Issue, error)
+	GetPullRequest(org, repo string, number int) (*github.PullRequest, error)
+	GetRepo(org, name string) (github.FullRepo, error)
+	IsMember(org, user string) (bool, error)
+	UpdatePullRequest(org, repo string, number int, title, body *string, open *bool, branch *string, canModify *bool) error
+}
+
+func helpProvider(_ *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	pluginHelp := &pluginhelp.PluginHelp{
+		Description: "The issue management plugin provides commands for linking and unlinking issues to a PR.",
+	}
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/link-issue <issue(s)>",
+		Description: "Links issue(s) to a PR in the same or different repo.",
+		WhoCanUse:   "Org members",
+		Examples:    []string{"/link-issue 1234", "/link-issue org/repo#789"},
+	})
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/unlink-issue <issue(s)>",
+		Description: "Unlinks issue(s) from a PR in the same or different repo.",
+		WhoCanUse:   "Org members",
+		Examples:    []string{"/unlink-issue 1234", "/unlink-issue org/repo#789"},
+	})
+	return pluginHelp, nil
+}
+
+func init() {
+	plugins.RegisterGenericCommentHandler(pluginName, handleGenericComment, helpProvider)
+}
+
+func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error {
+	return handleIssues(pc.GitHubClient, pc.Logger.WithFields(logrus.Fields{
+		"org":    e.Repo.Owner.Login,
+		"repo":   e.Repo.Name,
+		"number": e.Number,
+		"user":   e.User.Login,
+	}), e)
+}
+
+func handleIssues(gc githubClient, log *logrus.Entry, e github.GenericCommentEvent) error {
+	toLink, toUnlink, err := parseCommentForLinkCommands(e.Body)
+	if err != nil {
+		return err
+	}
+	if len(toLink) > 0 || len(toUnlink) > 0 {
+		return handleLinkIssue(gc, log, e, toLink, toUnlink)
+	}
+
+	return nil
+}
+
+func parseCommentForLinkCommands(commentBody string) ([]string, []string, error) {
+	extractIssues := func(re *regexp.Regexp) []string {
+		var issues []string
+		allMatches := re.FindAllStringSubmatch(commentBody, -1)
+
+		for _, match := range allMatches {
+			if len(match) > 1 {
+				issues = append(issues, strings.Fields(match[1])...)
+			}
+		}
+		return issues
+	}
+
+	return extractIssues(linkIssueRegex), extractIssues(unlinkIssueRegex), nil
+}
--- a/pkg/plugins/issue-management/link-issue.go
+++ b/pkg/plugins/issue-management/link-issue.go
@@ -1,0 +1,209 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// The `/link-issue` and `/unlink-issue` command allows
+// members of the org to link and unlink issues to PRs.
+package issuemanagement
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/prow/pkg/github"
+	"sigs.k8s.io/prow/pkg/plugins"
+)
+
+var (
+	fixesRegex = regexp.MustCompile(`(?i)^fixes\s+(.*)$`)
+)
+
+type IssueRef struct {
+	Org  string
+	Repo string
+	Num  int
+}
+
+func handleLinkIssue(gc githubClient, log *logrus.Entry, e github.GenericCommentEvent, linkIssues []string, unlinkIssues []string) error {
+	org := e.Repo.Owner.Login
+	repo := e.Repo.Name
+	number := e.Number
+	user := e.User.Login
+
+	var (
+		errs []error
+		sb   strings.Builder
+	)
+
+	if !e.IsPR || e.Action != github.GenericCommentActionCreated {
+		return gc.CreateComment(org, repo, number, plugins.FormatResponseRaw(
+			e.Body, e.HTMLURL, user, "This command can only be used on pull requests."))
+	}
+
+	isMember, err := gc.IsMember(org, user)
+	if err != nil {
+		return fmt.Errorf("unable to fetch if %s is an org member of %s: %w", user, org, err)
+	}
+	if !isMember {
+		return gc.CreateComment(org, repo, number, plugins.FormatResponseRaw(
+			e.Body, e.HTMLURL, user, "You must be an org member to use this command."))
+	}
+
+	// Verify if issues priovided in the comment are valid and format them accoordingly.
+	validateIssues := func(issues []string) []string {
+		var issueRefs []string
+		for _, issue := range issues {
+			issueRef, err := parseIssueRef(issue, org, repo)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Invalid format for issue **%s**. Supported formats are **issue-number** and **organization/repository#issue-number**", issue))
+				continue
+			}
+
+			// If repo or org of the issue reference is different from the one in which the PR is created, check if it exists
+			if org != issueRef.Org || repo != issueRef.Repo {
+				if _, err := gc.GetRepo(issueRef.Org, issueRef.Repo); err != nil {
+					errs = append(errs, fmt.Errorf("Failed to get repository details with name **%s**", issueRef.Repo))
+					continue
+				}
+			}
+			// Verify if the issue exists
+			fetchedIssue, err := gc.GetIssue(issueRef.Org, issueRef.Repo, issueRef.Num)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Failed to get issue **#%d** from **%s/%s**", issueRef.Num, issueRef.Org, issueRef.Repo))
+				continue
+			}
+			// Skip linking the issue if the provided issue number is a pull request
+			if fetchedIssue.IsPullRequest() {
+				errs = append(errs, fmt.Errorf("Skipping #%d of **%s/%s** as it is a ***pull request***.", fetchedIssue.Number, issueRef.Repo, issueRef.Org))
+				continue
+			}
+			issueRefs = append(issueRefs, formatIssueRef(issueRef, org, repo))
+		}
+		return issueRefs
+	}
+
+	toLink := validateIssues(linkIssues)
+	toUnlink := validateIssues(unlinkIssues)
+
+	if len(toLink) > 0 || len(toUnlink) > 0 {
+		pr, err := gc.GetPullRequest(org, repo, number)
+		if err != nil {
+			return fmt.Errorf("failed to get pull request: %w", err)
+		}
+
+		newBody := updateFixesLine(pr.Body, toLink, toUnlink)
+		if newBody != pr.Body {
+			if err := gc.UpdatePullRequest(org, repo, number, nil, &newBody, nil, nil, nil); err != nil {
+				return fmt.Errorf("failed to update PR body: %w", err)
+			}
+
+			log.Infof("Successfully updated the PR body by linking issues: %s and unlinking issues: %s", toLink, toUnlink)
+			sb.WriteString("Updated the `Fixes` line in the pull request body.\nHowever, ")
+		} else {
+			log.Debug("PR body is already up-to-date. No changes needed.")
+		}
+	}
+
+	if len(errs) > 0 {
+		sb.WriteString("There are one or more errors with issue references provided, please cross check and retry.\n")
+
+		for _, err := range errs {
+			fmt.Fprintf(&sb, "* %v\n", err)
+		}
+		return gc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, sb.String()))
+	}
+
+	return nil
+}
+
+func parseIssueRef(issue, defaultOrg, defaultRepo string) (*IssueRef, error) {
+	// Handling single issue references
+	if num, err := strconv.Atoi(issue); err == nil {
+		return &IssueRef{Org: defaultOrg, Repo: defaultRepo, Num: num}, nil
+	}
+
+	if !strings.Contains(issue, "/") {
+		return nil, fmt.Errorf("unrecognized issue reference: %s", issue)
+	}
+
+	// Handling issue references in format org/repo#issue-number
+	parts := strings.Split(issue, "#")
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("invalid issue ref: %s", issue)
+	}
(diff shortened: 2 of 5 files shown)


pkg/plugins/issue-management/issue_management.go
```diff
@@ -0,0 +1,95 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
stmcginnis, line 2: Just FYI, current guidance is to not include the year for these copyright headers. I don't think it matters, just pointing it out to help spread awareness.

pkg/plugins/issue-management/issue_management.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+const pluginName = "issue-management"
+
+var (
+	linkIssueRegex   = regexp.MustCompile(`(?mi)^/link-issue((?: +(?:\d+|[a-zA-Z0-9_.-]+/[a-zA-Z0-9_.-]+#\d+))+)\b`)
+	unlinkIssueRegex = regexp.MustCompile(`(?mi)^/unlink-issue((?: +(?:\d+|[a-zA-Z0-9_.-]+/[a-zA-Z0-9_.-]+#\d+))+)\b`)
+)
+
+type githubClient interface {
```
stmcginnis, line 38: I don't believe there is a reason for defining this interface unless you are [going to mock it for unit testing](https://github.com/kubernetes-sigs/prow/blob/e3ae8cf22a59dce06665a4eec8ed70d14c276c77/pkg/plugins/cherrypickapproved/cherrypickapproved.go#L77). Is that planned as a follow up? Might be good to have that included in this PR.

Amulyam24: Yes, I was planning that as a follow up. Sure, I'll add it to this PR.

Amulyam24:
/hold 

WIP: Adding UT

pkg/plugins/issue-management/issue_management.go
Amulyam24: Hi @stmcginnis, I explored a couple of ways on adding the UT.

At the end I feel having this interface and using the existing fake client is convenient than mocking the `githubClient` or using the `plugins.PluginGitHubClient` directly instead of `githubClient` as that would need mocking as well.

PTAL and let me know your thoughts!

petr-muller: Having a small local  interface is somewhat established pattern in GH-facing plugins ([retitle](https://github.com/kubernetes-sigs/prow/blob/e3ae8cf22a59dce06665a4eec8ed70d14c276c77/pkg/plugins/retitle/retitle.go#L91-L97), [blunderbuss](https://github.com/kubernetes-sigs/prow/blob/e3ae8cf22a59dce06665a4eec8ed70d14c276c77/pkg/plugins/blunderbuss/blunderbuss.go#L125-L131)), the GH client interface surface is massive and it is often useful to have an idea of what GH operations the plugin is limited to (besides the mentioned test mock use case). I'd prefer to keep it.

pkg/plugins/issue-management/issue_management.go
```diff
@@ -0,0 +1,95 @@
...
+
+func handleIssues(gc githubClient, log *logrus.Entry, e github.GenericCommentEvent) error {
+
+	switch {
+	case linkIssueRegex.MatchString(e.Body):
+		log.WithFields(logrus.Fields{
+			"org":    e.Repo.Owner.Login,
+			"repo":   e.Repo.Name,
+			"number": e.Number,
+			"user":   e.User.Login,
```
petr-muller: let's make this `logrus.Fieds` instance  in `handleGenericComment` and pass it to here with `pc.Logger.WithFields`, then this method already gets a populated logger and can just `log.Info` the necessary  message

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+	matches := regex.FindStringSubmatch(e.Body)
+	if len(matches) == 0 {
+		return nil
+	}
+
+	issues := strings.Fields(matches[1])
+	if len(issues) == 0 {
+		log.Info("No issue references provided in the comment.")
+		return nil
+	}
```
petr-muller: so we first do a `MatchString` at the callsite (in `handleIssues`), set a boolean based on that, then we use that boolean to select the regex again and match it again with `FindStringSubmatch`

feels like we can do all this at the callsite (in `handleIssues`) and just pass `issues` to this method (in addition to the link/unlink boolean)

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func parseIssueRef(issue, defaultOrg, defaultRepo string) (*IssueRef, error) {
+	// Handling single issue references
+	if num, err := strconv.Atoi(issue); err == nil {
+		return &IssueRef{Org: defaultOrg, Repo: defaultRepo, Num: num}, nil
+	}
+
+	// Handling issue references in format org/repo#issue-number
+	if strings.Contains(issue, "/") {
```
petr-muller: nit: would slightly prefer making this a standard guard clause
```suggestion
	if !strings.Contains(issue, "/") {
	  return nil, fmt.Errorf("unrecognized issue reference: %s", issue)
	}
	
	parts := strings.Split(issue, "#")
	...
	return &IssueRef{Org: orgRepo[0], Repo: orgRepo[1], Num: num}, nil
```

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	pr, err := gc.GetPullRequest(org, repo, number)
+	if err != nil {
+		return fmt.Errorf("failed to get PR: %w", err)
+	}
+
+	newBody := updateFixesLine(pr.Body, issueRefs, linkIssue)
+	if newBody == pr.Body {
+		log.Info("PR body is already up-to-date. No changes needed.")
```
petr-muller: pls downgrade to debug

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil
+	}
+
+	var issueRefs []string
+	for _, issue := range issues {
+		issueRef, err := parseIssueRef(issue, org, repo)
+		if err != nil {
+			log.Warnf("Skipping invalid issue: %s", issue)
+			continue
+		}
```
petr-muller, line 108: this is an invalid user input, it does not make sense to log it as warning - whoever is looking at hook logs is unlikely to care or do something about it. if anything log it as `debug`.

In general, ideally any problem with what the user provided should be surfaced to them through a response comment (like you do with "linked issue is actually a PR" case). There can be more issues and they can fail differently, so I think we need to accumulate all errors of this type and if non-empty, surface it to the user through a comment with bullet point list or something.

pkg/plugins/issue-management/issue_management.go
```diff
@@ -0,0 +1,95 @@
...
+		}).Info("Handling link issue command")
+		return handleLinkIssue(gc, log, e, true)
+	case unlinkIssueRegex.MatchString(e.Body):
+		log.WithFields(logrus.Fields{
+			"org":    e.Repo.Owner.Login,
+			"repo":   e.Repo.Name,
+			"number": e.Number,
+			"user":   e.User.Login,
+		}).Info("Handling unlink issue command")
+		return handleLinkIssue(gc, log, e, false)
```
petr-muller: hmmm this means the user can either say `/link` or say `/unlink` but not both; if they do both, then only the `/link` will count. But these operations are sometimes symmetrical, let's say you make a typo and link something wrong, then you'd probably want to fix it like this:

```
/unlink wrong
/link right
```

So I think we should have something like a method that is given a comment body and it produces issues to add links to, and issues to remove links to, and then we'd pass both to `handleLinkIssue` instead of the `add` boolean`:

```
toLink, to unlink := parseCommentForLinkCommands(...)
if len(toLink) == 0 && len(toUnlink) == 0 {
  return nothing to do
}
return handleLinkIssue(..., toLink, toUnlink)
```

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+
+		// If repo in issue reference is different from the PR, check if it exists
+		if repo != issueRef.Repo {
+			if _, err := gc.GetRepo(issueRef.Org, issueRef.Repo); err != nil {
+				return fmt.Errorf("failed to get repo: %w", err)
```
petr-muller: collect errors and respond with summary comment

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+		}
+
+		// Verify if the issue exists
+		fetchedIssue, err := gc.GetIssue(issueRef.Org, issueRef.Repo, issueRef.Num)
+		if err != nil {
+			return fmt.Errorf("failed to get issue: %w", err)
```
petr-muller: collect errors and respond with summary comment

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+		if fetchedIssue.IsPullRequest() {
+			response := fmt.Sprintf("Skipping #%d of repo **%s** and org **%s** as it is a *pull request*.", fetchedIssue.Number, issueRef.Repo, issueRef.Org)
+			if err := gc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, response)); err != nil {
```
petr-muller: collect errors and respond with summary comment

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func formatIssueRef(ref *IssueRef, defaultOrg, defaultRepo string) string {
+	if ref.Org == defaultOrg && ref.Repo == defaultRepo {
+		return fmt.Sprintf("#%d", ref.Num)
+	}
+	return fmt.Sprintf("%s/%s#%d", ref.Org, ref.Repo, ref.Num)
+}
+
+func updateFixesLine(body string, issueRefs []string, add bool) string {
```
petr-muller: now if this is changed to the toLink / toUnlink concept I propsed above, then I think we can use sets to compute the resulting issues that the pr should have links to: `existing.Difference(toUnlink).Union(toLink)`

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,216 @@
...
+	var issueRefs []string
+	for _, issue := range issues {
+		issueRef, err := parseIssueRef(issue, org, repo)
+		if err != nil {
+			log.Warnf("Skipping invalid issue: %s", issue)
+			continue
+		}
+
+		// If repo in issue reference is different from the PR, check if it exists
+		if repo != issueRef.Repo {
```
petr-muller:
```suggestion
		if org != issueRef.Org || repo != issueRef.Repo {
```

I *think* we need to do the same if the repo name is the same but org differs (like in a fork)

petr-muller:
I think this is pretty close, I appreciate the test coverage. Left some comment about the code structure.

Amulyam24:
Hi @petr-muller, was out for year end vacation. Most of the review comments are in place, will be working on enhancing the logic as mentioned above and push the changes this week, Thanks!

pkg/plugins/issue-management/link-issue.go
Amulyam24: Initially, I went with the design assumption that we can skip the invalid issue, it makes sense to let the user know about it. 

@petr-muller, one question here, if a user request to link issues X, Y & Z and say X & Y have some errors associated with them, do we want to proceed with successfully linking Z and then add a comment on what went wrong with X & Y for the user to retry?
IMO, we should link whatever is correct and provide user details on what went wrong.

Amulyam24:
hi @petr-muller, I have updated the logic to handle both link and unlink and tested possible scenarios. UT has been added accordingly. The PR is ready for review, PTAL, Thanks!

pkg/plugins/issue-management/link-issue.go
Amulyam24: ~I added minimal logic as of now, let me know if a custom set implementation is recommended, Thanks!~
Nvm, implemented this using the existing sets utility in k8s.io/apimachinery, PTAL!

Amulyam24:
Hi @petr-muller, can you PTAL? Thanks!

Amulyam24:
Hi @petr-muller @matthyx 
Looking forward to hearing your thoughts, PTAL!
Thanks :)

pkg/plugins/issue-management/issue_management.go
```diff
@@ -0,0 +1,105 @@
...
+	GetIssue(org, repo string, number int) (*github.Issue, error)
+	GetPullRequest(org, repo string, number int) (*github.PullRequest, error)
+	GetRepo(org, name string) (github.FullRepo, error)
+	IsMember(org, user string) (bool, error)
+	UpdatePullRequest(org, repo string, number int, title, body *string, open *bool, branch *string, canModify *bool) error
+}
+
+func helpProvider(_ *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	pluginHelp := &pluginhelp.PluginHelp{
+		Description: "The issue management plugin provides commands for linking and unlinking issues to a PR.",
```
petr-muller, line 49:
```suggestion
		Description: "The issue management plugin provides commands for managing issues in a repository",
```

nit: `issue-management` aims to be a bit more general thing so it's description should not be that specific

pkg/plugins/issue-management/issue_management.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+	if len(toLink) > 0 || len(toUnlink) > 0 {
+		return handleLinkIssue(gc, log, e, toLink, toUnlink)
+	}
+
+	return nil
+}
+
+func parseCommentForLinkCommands(commentBody string) ([]string, []string, error) {
```
petr-muller, line 91:
```suggestion
func parseCommentForLinkCommands(commentBody string) ([]string, []string) {
```

nit: this never returns any error, so no need to have that and then even the callsite can be simpler

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+			e.Body, e.HTMLURL, user, "You must be an org member to use this command."))
+	}
+
+	// Verify if issues priovided in the comment are valid and format them accoordingly.
+	validateIssues := func(issues []string) []string {
+		var issueRefs []string
+		for _, issue := range issues {
+			issueRef, err := parseIssueRef(issue, org, repo)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Invalid format for issue **%s**. Supported formats are **issue-number** and **organization/repository#issue-number**", issue))
```
petr-muller, line 74: I'm not sure about the value of wrapping these into errors because we don't really check these for being nil or something anywhere later, we just write them out as a list of strings into the user-facing comment. 

Accumulating the messages into something like `var errMessages []string` would serve the same purpose and lead to a simpler code

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+
+var (
+	fixesRegex = regexp.MustCompile(`(?i)^fixes\s+(.*)$`)
+)
+
+type IssueRef struct {
+	Org  string
+	Repo string
+	Num  int
+}
```
petr-muller, line 41: does not seem necessary to be public

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+	isMember, err := gc.IsMember(org, user)
+	if err != nil {
+		return fmt.Errorf("unable to fetch if %s is an org member of %s: %w", user, org, err)
+	}
+	if !isMember {
+		return gc.CreateComment(org, repo, number, plugins.FormatResponseRaw(
+			e.Body, e.HTMLURL, user, "You must be an org member to use this command."))
+	}
+
+	// Verify if issues priovided in the comment are valid and format them accoordingly.
```
petr-muller, line 68:
```suggestion
	// Verify if issues provided in the comment are valid and format them accordingly.
```

nit

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+			issueRef, err := parseIssueRef(issue, org, repo)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Invalid format for issue **%s**. Supported formats are **issue-number** and **organization/repository#issue-number**", issue))
+				continue
+			}
+
+			// If repo or org of the issue reference is different from the one in which the PR is created, check if it exists
+			if org != issueRef.Org || repo != issueRef.Repo {
+				if _, err := gc.GetRepo(issueRef.Org, issueRef.Repo); err != nil {
+					errs = append(errs, fmt.Errorf("Failed to get repository details with name **%s**", issueRef.Repo))
```
petr-muller, line 81: this ends up in a user-facing message so framing it as "failed to get repository details" is not optimal (also we only mention repo name here but the *org* may be the part that causes trouble)

Something about "Repository org/repo does not exist" would be better.

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+					continue
+				}
+			}
+			// Verify if the issue exists
+			fetchedIssue, err := gc.GetIssue(issueRef.Org, issueRef.Repo, issueRef.Num)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Failed to get issue **#%d** from **%s/%s**", issueRef.Num, issueRef.Org, issueRef.Repo))
```
petr-muller, line 88: Same here, something more user facing like "Issue X in repository Y/Z does not exist" would be better

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+				continue
+			}
+			// Skip linking the issue if the provided issue number is a pull request
+			if fetchedIssue.IsPullRequest() {
+				errs = append(errs, fmt.Errorf("Skipping #%d of **%s/%s** as it is a ***pull request***.", fetchedIssue.Number, issueRef.Repo, issueRef.Org))
```
petr-muller, line 93: nit: Other messages are in past tense, so "Skipped" and ideally "Skipped issue 1234 of ..."

also Org and Repo args order are swapped

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+				continue
+			}
+			issueRefs = append(issueRefs, formatIssueRef(issueRef, org, repo))
+		}
+		return issueRefs
+	}
+
+	toLink := validateIssues(linkIssues)
+	toUnlink := validateIssues(unlinkIssues)
```
petr-muller, line 102: Do the existence checks (org/repo exists, issue exists, is a PR) make sense for unlink commands? Someone could use /unlink to remove a bad reference that someone put in a description manually, no?

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+
+			log.Infof("Successfully updated the PR body by linking issues: %s and unlinking issues: %s", toLink, toUnlink)
+			sb.WriteString("Updated the `Fixes` line in the pull request body.\nHowever, ")
+		} else {
+			log.Debug("PR body is already up-to-date. No changes needed.")
+		}
+	}
+
+	if len(errs) > 0 {
+		sb.WriteString("There are one or more errors with issue references provided, please cross check and retry.\n")
```
petr-muller, line 124:
```suggestion
		sb.WriteString("there are one or more errors with issue references provided, please cross check and retry:\n")
```

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+	repo := e.Repo.Name
+	number := e.Number
+	user := e.User.Login
+
+	var (
+		errs []error
+		sb   strings.Builder
+	)
+
+	if !e.IsPR || e.Action != github.GenericCommentActionCreated {
```
petr-muller, line 54: Posting a comment on non- `GenericCommentActionCreated` events does not sound right, I think we should check for that first and simply return on unrelated events.

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+		return fmt.Sprintf("#%d", ref.Num)
+	}
+	return fmt.Sprintf("%s/%s#%d", ref.Org, ref.Repo, ref.Num)
+}
+
+func updateFixesLine(body string, toAdd []string, toRemove []string) string {
+	lines := strings.Split(body, "\n")
+	var fixesLine string
+	fixesIndex := -1
+	issueSet := sets.NewString()
```
petr-muller, line 173:
```suggestion
	issueSet := sets.New[string]()
```

nit (deprecated), also for other `NewString`

pkg/plugins/issue-management/link-issue.go
```diff
@@ -0,0 +1,209 @@
...
+		if newBody != pr.Body {
+			if err := gc.UpdatePullRequest(org, repo, number, nil, &newBody, nil, nil, nil); err != nil {
+				return fmt.Errorf("failed to update PR body: %w", err)
+			}
+
+			log.Infof("Successfully updated the PR body by linking issues: %s and unlinking issues: %s", toLink, toUnlink)
+			sb.WriteString("Updated the `Fixes` line in the pull request body.\nHowever, ")
+		} else {
+			log.Debug("PR body is already up-to-date. No changes needed.")
+		}
```
petr-muller, line 120: I think that we should respond with a comment *always* after a user uses a slash command, to not confuse them into thinking the command did not work. Even if we have not actually change the description because it is already in the desired state, it is good to reassure the user that the comment  was seen and acted on. Ideally we assemble one comment that says both what was done and what failed.

Amulyam24:
> Given this is a new plugin and merging won't break anyone, I think we can merge this and the review can be addressed in a followup PR, WDYT?


Sure @petr-muller, I will go ahead and unhold the PR

Will raise a follow up PR addressing the review comments, Thank you!

PR #560, opened 2025-12-03
Update gimme and fix codegen determinism

saschagrunert:
Updated gimme to support Go 1.24.x and added `-modtime 1` flag to go-bindata for deterministic timestamps.

Go 1.25.x has a regression in `go/doc` package that breaks godoc comment extraction. The old gimme couldn't install Go 1.24.0 from `.go-version`, causing CI to use Go 1.25.3 and generate files without comments.


saschagrunert:
PTAL @petr-muller @matthyx @droslean @krzyzacy @smg247

stmcginnis:
Looks correct to me.

/lgtm

PR #563, opened 2025-12-04
Fix race condition when re-triggering GitHub Actions

### LINKED ISSUE #337: Tide merges PR when retesting GitHub action
It happens from time to time that tide merges a PR when re-triggering a GitHub action. This results in the following contextual situation:

screenshot

We can see that `e2e-fedora` has not passed but also not yet started. Example PR:
- https://github.com/kubernetes-sigs/security-profiles-operator/pull/2595

Tide still merged the PR, I assume because of a race like outlined in:
https://github.com/kubernetes-sigs/prow/blob/f9126b38a43149c74fc526822d4550368efc60ce/pkg/tide/status.go#L478-L492

saschagrunert:
When a GitHub Action is re-triggered, GitHub temporarily removes the old check status before the new run starts. This causes a race where Tide may merge the PR during the brief window when the check is missing.

This fix tracks previously seen contexts per PR/commit. When a required context disappears, it's treated as PENDING to prevent premature merging.

Fixes #337

### DIFF
--- a/pkg/tide/status.go
+++ b/pkg/tide/status.go
@@ -62,6 +62,66 @@ type storedState struct {
 	PreviousQuery string
 }
 
+// contextHistory tracks previously seen contexts for a PR/commit combination
+// to detect when required contexts disappear (e.g., when GitHub Actions are re-triggered)
+type contextHistory struct {
+	prCommitContexts map[string]sets.Set[string]
+	sync.Mutex
+}
+
+func newContextHistory() *contextHistory {
+	return &contextHistory{
+		prCommitContexts: make(map[string]sets.Set[string]),
+	}
+}
+
+// checkAndUpdate returns contexts that disappeared since last check, then updates history.
+// This detects the race condition when GitHub Actions are re-triggered and the check
+// status temporarily disappears before the new run starts.
+//
+// Note: Uses PR number only (not commit SHA) to track across force pushes. This means
+// if a force push genuinely removes a context from the workflow, it will be detected
+// as "disappeared" until the next sync. This is acceptable as it errs on the side of
+// preventing premature merges.
+func (ch *contextHistory) checkAndUpdate(pr *CodeReviewCommon, currentContexts []string) []string {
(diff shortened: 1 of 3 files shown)


petr-muller:
Wow that GH spam is annoying, need to fix that automation to not link to PRs / issues this way.

I have a bit of a hard time wrapping my head around this fix:

1. If a **known required** (by Tide) GH check is not passing (by not being momentarily present on a PR), how does it lead to a spurious merge?
2. If for some reason ^^^ has some reasonable explanation, why do we need to track previously seen checks instead of relying on configuration-time knowledge that a certain check is required?

*Maybe* I understand wrong and the behavior fixed here is actually Tide not knowing anything about specific GH checks, but doing some kind of "unknown non-Prow results must PASS when present" behavior. Tide would not be configured to require a certain non-Prow check, but would be configured to require non-Prow checks to be passing, and the temporary absence of a certain non-Prow result would pass that criteria leading to a merge. Can you confirm?

I have not yet looked at the code closely, but I wonder if we can have a situation where a check goes away legitimately (when a job is removed between when it last runs on a commit and when we evaluate later). Can that happen? Skimming the change it seems we only expect checks that were present on the same commit in the past. At least status contexts cannot be removed from a commit so this looks good but I'm not sure about all the Check behaviors.

saschagrunert:
@petr-muller it happens pretty often when a retrigger a flaking GH action, for example in: https://github.com/kubernetes-sigs/cri-tools/pull/1973 most recently:

- The action failed, tide is complaining and won't merge.
- I retrigger the GH action
- Tide merges during the time where the context is not visible / available to it.

This leaves the PR in a state where the CI is still running but it got already merged, see:
screenshot

BenTheElder:
> Tide merges during the time where the context is not visible / available to it.

I think if you configure it as `Required` then github will always have a status for it? (will switch to pending)

You can do that via branch-protector config.

I spot checked https://github.com/kubernetes-sigs/cri-tools/pull/1997 and it looks like none of these are marked Required. That may be the necessary fix.

Not sure which ones you want so the docs are:
https://docs.prow.k8s.io/docs/components/optional/branchprotector/

specifically `required_status_checks`

PR #567, opened 2025-12-07
Update  base images

prad9192:
Update prow base images

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -4,49 +4,49 @@
 # git-custom-k8s-auth which the same alpine image with kubernetes auth clients installed for AWS/GCP
 # git-custom-k8s-auth should only be used for components that talk to Kubernetes Clusters.
 baseImageOverrides:
-  sigs.k8s.io/prow/cmd/branchprotector: gcr.io/k8s-staging-test-infra/alpine:v20240719-47a381b1df
-  sigs.k8s.io/prow/cmd/checkconfig: gcr.io/k8s-prow/git:v20240729-4f255edb07
-  sigs.k8s.io/prow/cmd/clonerefs: gcr.io/k8s-prow/git:v20240729-4f255edb07
-  sigs.k8s.io/prow/cmd/config-bootstrapper: gcr.io/k8s-staging-test-infra/git-custom-k8s-auth:v20240719-47a381b1df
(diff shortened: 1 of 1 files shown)


site/go.mod
```diff
@@ -1,8 +1,3 @@
 module github.com/kubernetes-sigs/prow
 
-go 1.24
-
-require (
-	github.com/google/docsy v0.12.0 // indirect
-	github.com/google/docsy/dependencies v0.7.2 // indirect
-)
```
petr-muller, line 8: Why is this removed?

prad9192: Oh ```go mod tidy``` might have removed them, let me check

prad9192: Good catch on the removed dependencies. 
I updated the PR to address your feedback

stmcginnis:
Not trying to derail anything, but curious if it would be better to just move to 1.25.

petr-muller:
/retitle Update Go to 1.24.2 and base images
/test all

petr-muller:
Unless we're targeting a specific CVE that we think is really important for Prow to have fixed, I see no reason why we should call out CVE (I guess this was only done for the PR to sound more "important"? not sure).

Anyway, bumping base images and bumping Go version are two separate changes that may have potentially big impact (especially the base image bumps) and I'd prefer them to be done in separate commits at minimum, but even better in separate PRs. And Sean is right, there's Go 1.25 since August, if we're bumping Go we may as well take the effort to bump to that.

prad9192:
> Unless we're targeting a specific CVE that we think is really important for Prow to have fixed, I see no reason why we should call out CVE (I guess this was only done for the PR to sound more "important"? not sure).
> 
> Anyway, bumping base images and bumping Go version are two separate changes that may have potentially big impact (especially the base image bumps) and I'd prefer them to be done in separate commits at minimum, but even better in separate PRs. And Sean is right, there's Go 1.25 since August, if we're bumping Go we may as well take the effort to bump to that.

Thanks for your feedback, I agree I should have been more specific about the CVE and separated the changes.

The base image updates is to resolve CVEs which appeared in our internal security scans.
https://github.com/libexpat/libexpat/issues/888

I have updated this PR to focus only on the base image updates.

The Go version update (for https://pkg.go.dev/vuln/GO-2025-3563) will be submitted as a separate PR to keep changes separate and easier to review.

prad9192:
Updating go module to latest here - https://github.com/kubernetes-sigs/prow/pull/580

Issue #544, opened 2025-11-10
cherrypicker plugin does not process all cherrypick requests if one fails.


VannTen:
When typing a comment such as
```
/cherrypick release-X
/cherrypick release-Y
/cherrypick release-Z
```

Prow will stop doing anything if the cherrypick on release-Z fails (because apparently the cherrypick are processed in inverse order).

Example of the behavior : https://github.com/kubernetes-sigs/kubespray/pull/12682#issuecomment-3511421409

petr-muller:
I think this is handled differently when the cherrypick comment is issued on an unmerged and merged PR. In the former case, the cherrypicker eventually reacts on merge event and seems to correctly try all cherrypicks, iterating over all and accumulating errors: https://github.com/kubernetes-sigs/prow/blob/98a0c9d48d175803e1c616b856f39362011abef3/cmd/external-plugins/cherrypicker/server.go#L452-L455

In the latter case, like in https://github.com/kubernetes-sigs/kubespray/pull/12682#issuecomment-3511417842, it starts cherrypicking directly on the comment event and bails out on the first error: https://github.com/kubernetes-sigs/prow/blob/98a0c9d48d175803e1c616b856f39362011abef3/cmd/external-plugins/cherrypicker/server.go#L304-L306

/label help-wanted

petr-muller:
/label good-first-issue

I think https://github.com/kubernetes-sigs/prow/issues/544#issuecomment-3514502849 may be sufficient for a new contributor to be able to pick this up

valen-mascarenhas14:
I've added the fix in the following [PR](https://github.com/kubernetes-sigs/prow/pull/570)
Changes worked fine locally, cherrypick was able to proceed with the remaining targets even after encountering a failed targed.

Here's the key code change
```
var errs []error

for _, targetBranch := range sets.List(sets.KeySet(commands)) {
    if err := s.handle(...); err != nil {
        errs = append(errs, err)
        continue // do not abort remaining cherry-picks
    }
}

if len(errs) > 0 {
    return log, utilerrors.NewAggregate(errs)
}
```

Sharing the snapshot with the successful build for reference

<img width="1060" height="1036" alt="Image" src="https://github.com/user-attachments/assets/97cbbd43-31c0-4b6f-8e5b-e0824929fff6" />

valen-mascarenhas14:
> Prow will stop doing anything if the cherrypick on release-Z fails (because apparently the cherrypick are processed in inverse order).

[...]

PR #570, opened 2025-12-09
cherrypick: do not abort on first failure

valen-mascarenhas14:
Fix for #544 
This change fixes the behavior in the cherrypicker plugin where processing would stop after the first cherry-pick failure, even if additional valid targets were requested in the same comment.
The logic is updated to continue processing remaining branches while aggregating and reporting all errors at the end.

**Key change**
```
var errs []error

for _, targetBranch := range sets.List(sets.KeySet(commands)) {
    if err := s.handle(...); err != nil {
        errs = append(errs, err)
        continue 
    }
}

if len(errs) > 0 {
    return log, utilerrors.NewAggregate(errs)
}
```
This ensures that:
* Successful cherry-picks are not skipped due to unrelated failures
* Failures are still properly surfaced to the user

### DIFF
--- a/cmd/external-plugins/cherrypicker/server.go
+++ b/cmd/external-plugins/cherrypicker/server.go
@@ -305,17 +305,12 @@ func (s *Server) handleIssueComment(log logrus.FieldLogger, ic github.IssueComme
 		branchLog.Debug("Cherrypick request.")
 
 		if err := s.handle(branchLog, ic.Comment.User.Login, &ic.Comment, org, repo, targetBranch, baseBranch, commands[targetBranch], title, body, num); err != nil {
-			branchLog.WithError(err).Error("Cherrypick failed")
 			errs = append(errs, fmt.Errorf("failed to handle cherrypick for %s: %w", targetBranch, err))
-        	continue 
+			continue
 		}
 	}
 
-	if len(errs) > 0 {
-    return log, utilerrors.NewAggregate(errs) 
-	}
-
-	return log, nil
+	return log, utilerrors.NewAggregate(errs)
 }
 
 type cherrypickCommands map[string][]string
--- a/cmd/external-plugins/cherrypicker/server_test.go
+++ b/cmd/external-plugins/cherrypicker/server_test.go
@@ -20,9 +20,9 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"testing"
-	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


valen-mascarenhas14:
@droslean @BenTheElder @petr-muller Any inputs on this ?

matthyx:
@valen-mascarenhas14 sorry I forgot about this... let me know if after few days you don't hear from them I'll approve

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -302,10 +305,16 @@ func (s *Server) handleIssueComment(log logrus.FieldLogger, ic github.IssueComme
...
 
 		if err := s.handle(branchLog, ic.Comment.User.Login, &ic.Comment, org, repo, targetBranch, baseBranch, commands[targetBranch], title, body, num); err != nil {
-			return log, fmt.Errorf("failed to handle cherrypick for %s: %w", targetBranch, err)
+			branchLog.WithError(err).Error("Cherrypick failed")
+			errs = append(errs, fmt.Errorf("failed to handle cherrypick for %s: %w", targetBranch, err))
+        	continue 
 		}
 	}
 
+	if len(errs) > 0 {
```
droslean: You don't need this check. You can just `return log, utilerrors.NewAggregate(errs) `

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -302,10 +305,16 @@ func (s *Server) handleIssueComment(log logrus.FieldLogger, ic github.IssueComme
 		branchLog.Debug("Cherrypick request.")
 
 		if err := s.handle(branchLog, ic.Comment.User.Login, &ic.Comment, org, repo, targetBranch, baseBranch, commands[targetBranch], title, body, num); err != nil {
-			return log, fmt.Errorf("failed to handle cherrypick for %s: %w", targetBranch, err)
+			branchLog.WithError(err).Error("Cherrypick failed")
```
droslean: In the end all of the errors should be logged since you will return them.

cmd/external-plugins/cherrypicker/server_test.go
```diff
@@ -295,6 +296,96 @@ func TestCherryPickICV2(t *testing.T) {
 	testCherryPickIC(localgit.NewV2, t)
 }
 
+func TestCherryPickMultipleBranches_OneFailsOthersProceed(t *testing.T) {
```
droslean: You shouldn't need a new test function just for that specific case. Just ensure that you will add this test case to the existing tests.

valen-mascarenhas14: Noted, have made the changes

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: Noted

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: Noted

Issue #328, opened 2024-11-20
Allow Option for Ingress to Reach pods through SSL


NiJuFirenzia:
Currently Pods such as Deck and Hook rely on the SSL connection being terminated after reaching Ingress. This only allows for a http backend protocol. We would like for the option to also allow Prow pods to talk to the ingress without terminating ssl so that ingresses can have an https backend protocol. This will allow for a more secure set up.

BenTheElder:
/kind feature

Have you considered something like mTLS with istio / linkerd / ...?

I don't think we have a use case for the kubernetes project to add this complexity and we haven't seen this request from any other user so far, it's very common to use a loadbalancer or ingress for TLS termination to users.

NiJuFirenzia:
Hi Ben, we had a requirement come in from our organization that all ingresses must use an https backend protocol. I was thinking offering this option as well if anyone else using prow might run into similar requirements. I was able to get this working in my testing by modifying deck and hook to use the interrupts.ListenAndServeTLS() method. I plan to open a PR once everything is validated in the near future.

BenTheElder:
Please understand that prow has very limited maintainer bandwidth at the moment (I am not active as one, but I am keeping an eye on the repo as a lead of the sponsoring SIG), and the functionality we already have is essential to running the Kubernetes project itself.

NiJuFirenzia:
No worries, I would be happy to contribute this change and verify that it doesnt brake any current functionalities

NiJuFirenzia:
Hi team, I am planning on working on this and will have a PR up soon. I plan on having tests to ensure that the standard features of Prow is not affected by this change

PR #573, opened 2025-12-10
Adding option to enable Back End HTTPS for Prow Ingress

NiJuFirenzia:
Prow currently requires SSL connections to be terminated before connecting to the deck and hook pods. This does not enable users to run prow with backend https enabled on their ingress. This PR adds the option to enable backend https on ingresses by passing in specific flags to the hook and deck pods. The argument flags will look like as below: 

hook:
```yaml
- --enable-ssl=true
- --server-cert-file=/etc/prow/tls.crt
- --server-key-file=/etc/prow/tls.key
```

deck:
```yaml
- --enable-ssl=true
- --server-cert-file=/etc/prow/tls.crt
- --server-key-file=/etc/prow/tls.key
- --client-cert-file=etc/prow/deck-client.crt
```

In the example above, a secret containing the cert file and key file was mounted to the deck and hook pods. Additionally this has been tested to continue working as is with ingress backend to be set to just HTTP if the flags are not passed in.


NiJuFirenzia:
This PR addresses Issue https://github.com/kubernetes-sigs/prow/issues/328

matthyx:
@NiJuFirenzia thanks for your PR, can you check the CLA and sign it?

cmd/hook/main.go
```diff
@@ -72,6 +73,10 @@ type options struct {
 	instrumentationOptions prowflagutil.InstrumentationOptions
 	jira                   prowflagutil.JiraOptions
 
+	enableSSL bool
+	certFile  string
+	keyFile   string
```
petr-muller: let's group this into a small struct that we can embed at both places, we can share validation etc. similarly to how the `ThingOptions` bits above

cmd/hook/main.go
```diff
@@ -270,9 +286,16 @@ func main() {
...
-	httpServer := &http.Server{Addr: ":" + strconv.Itoa(o.port), Handler: hookMux}
-
 	health.ServeReady()
 
-	interrupts.ListenAndServe(httpServer, o.gracePeriod)
+	if o.enableSSL {
+		httpServer := &http.Server{
+			Addr:    ":" + strconv.Itoa(o.port),
+			Handler: hookMux,
+		}
```
petr-muller: this seems identical in both branches, so can be extracted before the condition

cmd/deck/pluginhelp.go
```diff
@@ -29,18 +33,21 @@ import (
 // cacheLife is the time that we keep a pluginhelp.Help struct before considering it stale.
 // We consider help valid for a minute to prevent excessive calls to hook.
 const cacheLife = time.Minute
+const tlsEnabledScehma = "https"
```
petr-muller:
```suggestion
const tlsEnabledSchema = "https"
```

cmd/deck/pluginhelp.go
```diff
@@ -52,6 +59,27 @@ func (ha *helpAgent) getHelp() (*pluginhelp.Help, error) {
...
+			return nil, fmt.Errorf("error decoding cert file: %w", err)
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+		//Error is eaten here for testing purposes. There should be no reason the transport is not of type HTTP
+		if transport, ok := http.DefaultTransport.(*http.Transport); ok {
+			transport.TLSClientConfig = &tls.Config{
+				RootCAs: caCertPool,
+			}
+			http.DefaultTransport = transport
```
petr-muller: helpAgent should get a custom client member with this transport instead of modifying the global default transport (this would affect anything else in deck that is a http client and relies on the default)

cmd/deck/pluginhelp.go
```diff
@@ -52,6 +59,27 @@ func (ha *helpAgent) getHelp() (*pluginhelp.Help, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("error parsing hook path: %w", err)
+	}
+	if hookPath.Scheme == tlsEnabledScehma {
+		caCert, err := os.ReadFile(ha.cert)
+		if err != nil {
+			return nil, fmt.Errorf("error decoding cert file: %w", err)
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
```
petr-muller: I am not entirely sure I like this coupling. This essentially reuses Deck's server cert as a CA cert to establish trust with Hook's server cert for this behavior where Deck is a client to hook's pluginhelp endpoint? That's a bit hacky and suprising.

petr-muller (changes requested):

cmd/deck/pluginhelp.go
NiJuFirenzia: That would require having to add another option flag to take in a client cert right? I had been trying to limit the options that we would need

NiJuFirenzia:
All change requests have been addressed and this PR is ready for a re-review

pkg/flagutil/ssl_enablement.go
```diff
@@ -0,0 +1,51 @@
...
+func (o *SSLEnablementOptions) AddFlags(fs *flag.FlagSet) {
+	fs.BoolVar(&o.EnableSSL, "enable-ssl", false, "Enable SSL to support Ingress backend HTTPS")
+	fs.StringVar(&o.ServerCertFile, "server-cert-file", "", "Location of the server cert file for hosting TLS call. This must be set if SSL is enabled.")
+	fs.StringVar(&o.ServerKeyFile, "server-key-file", "", "Location of the key file for hosting TLS call. This must be set if SSL is enabled.")
+}
+
+func (o *SSLEnablementOptions) Validate(_ bool) error {
+	if o.EnableSSL {
+		if o.ServerCertFile == "" {
+			return errors.New("flag --enable-ssl was set to true but required flag --cert-file was not set")
```
petr-muller:
```suggestion
			return errors.New("flag --enable-ssl was set to true but required flag --server-cert-file was not set")
```

actual flags are different

pkg/flagutil/ssl_enablement.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+		if o.ServerKeyFile == "" {
+			return errors.New("flag --enable-ssl was set to true but required flag --key-file was not set")
```
petr-muller:
```suggestion
			return errors.New("flag --enable-ssl was set to true but required flag --server-key-file was not set")
```

pkg/flagutil/ssl_enablement.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (o *SSLEnablementOptions) AddFlags(fs *flag.FlagSet) {
+	fs.BoolVar(&o.EnableSSL, "enable-ssl", false, "Enable SSL to support Ingress backend HTTPS")
+	fs.StringVar(&o.ServerCertFile, "server-cert-file", "", "Location of the server cert file for hosting TLS call. This must be set if SSL is enabled.")
+	fs.StringVar(&o.ServerKeyFile, "server-key-file", "", "Location of the key file for hosting TLS call. This must be set if SSL is enabled.")
+}
+
+func (o *SSLEnablementOptions) Validate(_ bool) error {
+	if o.EnableSSL {
```
petr-muller: We should either check also the opposite case (enabled is false but cert was passed, possibly confusing the admin they are using SSL when they are not), or make this an implied field from the presence of at least one file flags (if at least one is set, assume the user wants to have it enabled and validate both are set).

cmd/hook/main.go
```diff
@@ -71,13 +71,14 @@ type options struct {
 	bugzilla               prowflagutil.BugzillaOptions
 	instrumentationOptions prowflagutil.InstrumentationOptions
 	jira                   prowflagutil.JiraOptions
+	sslEnablement          prowflagutil.SSLEnablementOptions
```
petr-muller:
```suggestion
	ssl          prowflagutil.SSLEnablementOptions
```

pkg/flagutil/ssl_enablement.go
```diff
@@ -0,0 +1,51 @@
...
+	"errors"
+	"flag"
+)
+
+// SSLEnablementOptions provides flags for enabling running deck and
+// hook with SSL enabled. This will allow for the Prow Ingress to run with
+// HTTPS backend. If enableSSL is true, both certFile and keyFile must be set
+// to the location of the cert and key files respectively.
+
+type SSLEnablementOptions struct {
```
petr-muller: tbh I'd call this just `SSLOptions`  or `SSLServerOptions` and the package `flagutil/ssl`

pkg/flagutil/ssl_enablement.go
```diff
@@ -0,0 +1,51 @@
...
+	EnableSSL      bool
+	ServerCertFile string
+	ServerKeyFile  string
```
petr-muller:
```suggestion
	Enabled      bool
	CertFile string
	KeyFile  string
```

the instance of this struct will typically called `ssl` or something so meaning is typically provided by that context

petr-muller (commented):
There's a flag name mismatch between error message and actual names. I have pointed out some opportunities to reduce stuttering (`sslEnablement.EnableSSL` and verbosity `sslEnablement` vs `ssl`) but treat them as non-blocking nits.

pkg/flagutil/ssl_enablement.go
NiJuFirenzia: I agree with the name changes. I'll have updates to this PR out next week

cmd/deck/main.go
```diff
@@ -161,6 +163,20 @@ func (o *options) Validate() error {
...
+		if err != nil {
+			return err
+		}
+		if sslEnabled && o.clientCertFile == "" {
+			return errors.New("flag --hook-url was set to an https url but required flag --client-cert-file was not set")
+		}
+	}
+
+	if o.ssl.CertFile != "" && o.clientCertFile == "" {
+		return errors.New("flag --server-cert-file and --server-key-file was set to true but required flag --client-cert-file was not set")
```
ivankatliarchuk, line 178: "flag --server-cert-file and --server-key-file was set to true" — these are string flags, not booleans. Should be "were set" (plural) and drop "to true"

cmd/deck/pluginhelp.go
```diff
@@ -29,19 +33,47 @@ import (
 // cacheLife is the time that we keep a pluginhelp.Help struct before considering it stale.
 // We consider help valid for a minute to prevent excessive calls to hook.
 const cacheLife = time.Minute
+const sslEnabledSchema = "https"
 
 type helpAgent struct {
-	path string
+	path   string
+	cert   string
```
ivankatliarchuk, line 40: cert field is a dead code. The field is set in newHelpAgent but never read after construction — the TLS client is fully initialized at construction time

cmd/deck/pluginhelp.go
```diff
@@ -29,19 +33,47 @@ import (
 // cacheLife is the time that we keep a pluginhelp.Help struct before considering it stale.
 // We consider help valid for a minute to prevent excessive calls to hook.
 const cacheLife = time.Minute
+const sslEnabledSchema = "https"
```
ivankatliarchuk, line 36:
```suggestion
const sslEnabledScheme = "https"
```

should be sslEnabledScheme ("scheme" not "schema")

cmd/deck/main.go
```diff
@@ -161,6 +163,20 @@ func (o *options) Validate() error {
...
+		sslEnabled, err := isHttpsPath(o.hookURL)
+		if err != nil {
+			return err
+		}
+		if sslEnabled && o.clientCertFile == "" {
+			return errors.New("flag --hook-url was set to an https url but required flag --client-cert-file was not set")
+		}
+	}
+
+	if o.ssl.CertFile != "" && o.clientCertFile == "" {
```
ivankatliarchuk, line 177: I'm not sure about this requirement. Requiring --client-cert-file whenever --server-cert-file is set does not sound correct - these are independent. Deck can serve HTTPS to the ingress without needing to call hook over HTTPS. Maybe the check should only trigger if --hook-url starts with https://

pkg/flagutil/ssl.go
```diff
@@ -0,0 +1,49 @@
...
+	KeyFile  string
+}
+
+func (o *SSLOptions) AddFlags(fs *flag.FlagSet) {
+	fs.StringVar(&o.CertFile, "server-cert-file", "", "Location of the server cert file for hosting TLS call. This must be set if SSL is enabled.")
+	fs.StringVar(&o.KeyFile, "server-key-file", "", "Location of the key file for hosting TLS call. This must be set if SSL is enabled.")
+}
+
+func (o *SSLOptions) Validate(_ bool) error {
+	if o.CertFile != "" || o.KeyFile != "" {
```
ivankatliarchuk, line 40: Worth to check for path exist and where or not file is missing.

 Both kubernetes_cluster_clients.goand k8s_client.go use os.Stat inside their Validate methods to check kubeconfig paths.

Probably something like 

```go
if _, err := os.Stat(o.CertFile); err != nil {
       return fmt.Errorf("--server-cert-file: %w", err)
}
if _, err := os.Stat(o.KeyFile); err != nil {
    return fmt.Errorf("--server-key-file: %w", err)
}
```

The benefit is that you get a clear error at startup rather than a cryptic TLS failure when ListenAndServeTLS is called.

ivankatliarchuk (changes requested):
I'm not too sure. Maybe missing something. What is definately missing is a documentation, like operator docs how to, then it could be easier to understand the intent.

Misleading flag name suggests mTLS was intended but not delivered, aka `--client-cert-file name` -> it's a CA cert, not a client cert from first look.

I'll try to explain

 CA certificate (what the code actually uses) When deck calls hook over HTTPS, it needs to verify that hook's server certificate is trustworthy. If hook uses a self-signed or internal CA cert, the OS trust store won't know about it. So you load that CA cert into a x509.CertPool and put it in RootCAs:

```go
  caCertPool.AppendCertsFromPEM(caCert)
  client = &http.Client{
      Transport: &http.Transport{
          TLSClientConfig: &tls.Config{
              RootCAs: caCertPool,  // "trust this CA when verifying the server"
          },
      },
  }
```

  This is one-way TLS: deck verifies hook's identity, but hook doesn't verify deck's identity.

 Client certificate (what the name implies)

  A client cert is deck's own certificate that it presents to hook during the handshake, so hook can verify deck's identity. That's mutual TLS (mTLS). It would look like:

```go
  clientCert, _ := tls.LoadX509KeyPair(certFile, keyFile)
  TLSClientConfig: &tls.Config{
      Certificates: []tls.Certificate{clientCert},  // "here's who I am"
      RootCAs: caCertPool,
  }
```

The problem
  
An operator reading --client-cert-file will naturally think "this is the cert deck presents to authenticate itself to hook." But the code uses it purely as a CA cert to trust hook's server cert. These are completely different files with different formats and different roles.

 A better name would be --hook-ca-cert-file or --hook-tls-ca-file, which makes it clear: "the CA cert used to verify hook's TLS certificate."

Another problem/concert from security perspective

The solution seems like half implemented. 
- One-way TLS only (deck verifies hook, hook doesn't verify deck)
- No cert rotation -> not too clear, missing documentation describing that
-

ivankatliarchuk:
For this PR, I would strongly recommend to create a documentation page describing this setup for operators and having in place describing alternatives. The honest answer without offence is: this in Application TLS is probably not worth the complexity it adds, this is opinionated answer, but some security best practice should be documented, and use cases with tradeoffs.

Something like a page to consider other approaches, before diving in APP tls setup and certificate lifecycle world. As an operator, I would like to know why I'm now managing certs, handling rotation, and adding TLS overhead on connections that are already encrypted at the network layer(as example).

Alternative approaches and tradeoff:

 Just use this PR
  - Makes sense if: no service mesh, small cluster, you control the CA, Prow is the only thing needing this
  - Low complexity, self-contained
  - The manual cert rotation problem is real though -> nothing in this PR handles cert rotation, and no description, no metric emitted with cert expiry

Service mesh (Istio, Linkerd, etc)
- mTLS between pods is automatic and transparent — no code changes, no cert management in the app
- Identity is based on SPIFFE/SPIRE workload identity, not manually provisioned certs
- You get observability, retries, circuit breaking for free
- Downside: operational complexity to run the mesh, but it worth depends on client/use-case
- If you already have a mesh, this APP TLS is unnecessary overhead

Zero trust / SPIFFE
- Similar to service mesh but more explicit — each workload gets a cryptographic identity
- Works well across clusters and cloud boundaries
- Downside: operational complexity to run the mesh, but it worth depends on client/use-case. More portable then mesh
- Better long-term if you're building toward zero trust across multiple environments

Prow running within a private network / network perimeter
- All traffic between nodes/pods travels over an encrypted tunnel
- The ingress→pod HTTP gap this PR is trying to close is already encrypted at the network layer
- Gap in: application-layer identity (any pod on the VPN can talk to any other pod), Compromised node and  Compliance requirements

NiJuFirenzia:
> For this PR, I would strongly recommend to create a documentation page describing this setup for operators and having in place describing alternatives. The honest answer without offence is: this in Application TLS is probably not worth the complexity it adds, this is opinionated answer, but some security best practice should be documented, and use cases with tradeoffs.
> 
> Something like a page to consider other approaches, before diving in APP tls setup and certificate lifecycle world. As an operator, I would like to know why I'm now managing certs, handling rotation, and adding TLS overhead on connections that are already encrypted at the network layer(as example).
> 
> Alternative approaches and tradeoff:
> 
> Just use this PR
> 
> * Makes sense if: no service mesh, small cluster, you control the CA, Prow is the only thing needing this
> * Low complexity, self-contained
> * The manual cert rotation problem is real though -> nothing in this PR handles cert rotation, and no description, no metric emitted with cert expiry
> 
> Service mesh (Istio, Linkerd, etc)
> 
> * mTLS between pods is automatic and transparent — no code changes, no cert management in the app
> * Identity is based on SPIFFE/SPIRE workload identity, not manually provisioned certs
> * You get observability, retries, circuit breaking for free
> * Downside: operational complexity to run the mesh, but it worth depends on client/use-case
> * If you already have a mesh, this APP TLS is unnecessary overhead
> 
> Zero trust / SPIFFE
> 
> * Similar to service mesh but more explicit — each workload gets a cryptographic identity
> * Works well across clusters and cloud boundaries
> * Downside: operational complexity to run the mesh, but it worth depends on client/use-case. More portable then mesh
> * Better long-term if you're building toward zero trust across multiple environments
> 
> Prow running within a private network / network perimeter
> 
> * All traffic between nodes/pods travels over an encrypted tunnel
> * The ingress→pod HTTP gap this PR is trying to close is already encrypted at the network layer
> * Gap in: application-layer identity (any pod on the VPN can talk to any other pod), Compromised node and  Compliance requirements

Hi Ivan, thanks for the review. I will make the changes and add the documentation. To answer your question regarding why the TLS encryption is being added to the application layer is because of a policy that our company, Fidelity Investments, implemented forcing us to enable backend https for all our ingresses running in a cluster. When that feature is enabled, the ingress cannot connect to the deck and hook pods if they are hosting over http. And so, I wanted to add this option for anyone else running prow in a similar situation.

PR #576, opened 2025-12-12
deck: add base path plumbing and helper APIs

### LINKED ISSUE #376: Cannot run prow at a subpath
I cannot find a way to have prow deck works when using a subpath.

Let's say i have the following ingress
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /$1
  name: prow-deck
  namespace: prow
spec:
  ingressClassName: nginx
  rules:
  - host: infra.example.com
    http:
      paths:
      - backend:
          service:
            name: prow-deck
            port:
              name: http
        path: /prow/?(.*)
```

I would expect prow to be available at `https://infra.example.com/prow` which is the case, but the static are not working and the prowjobs.js api are not working as they would be tried from `https://infra.example.com`

Reading the code, I cannot find a way to set a prefix

tsj-30:
This change teaches Deck about a configurable base path so it can be served from a sub‑URL without breaking static assets, APIs, or Spyglass. It introduces the --base-path flag, wraps the HTTP mux so requests are stripped/prefixed appropriately, extends template helpers to emit base-aware links, and adds deckURL utilities for the front-end. 
This fixes kubernetes-sigs/prow#376.


tsj-30:
CC: @petr-muller , @matthyx 
As suggested by Petr, I’ll break this down and raise smaller PRs to fix #376.

tsj-30:
Hi @petr-muller ,
Can you PTAL?
Thanks 😄

tsj-30:
cc: @loispostula

loispostula:
@tsj-30 I'm not sure how to test this, I'll try to get a local build working and check

tsj-30:
Hi @loispostula, 
Did you get a chance to test this, I have completed some additional code changes and looking forward to submit a follow PR if this looks good and merged.
If you need any futher help from my side please let me know.
Thanks!

cmd/deck/basepath.go
```diff
@@ -0,0 +1,113 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
kishen-v: Nit: Please update the year.  
Also, CNCF does not recommend mentioning the `year`.  https://github.com/cncf/foundation/blob/main/copyright-notices.md#copyright-notices.

cmd/deck/main.go
```diff
@@ -166,6 +168,7 @@ func (o *options) Validate() error {
 
 func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	var o options
+	o.basePath = "/"
```
petr-muller, line 171: nit: this is redundant with setting a default on L192

cmd/deck/main.go
```diff
@@ -140,9 +140,11 @@ type options struct {
 	controllerManager     prowflagutil.ControllerManagerOptions
 	dryRun                bool
 	tenantIDs             prowflagutil.Strings
+	basePath              string
 }
 
 func (o *options) Validate() error {
+	o.basePath = normalizeBasePath(o.basePath)
```
petr-muller, line 147: seems redundant with doing the same in `gatherOptions`

petr-muller:
/cc
/hold

I have an in-flight review with some feedback, will post hopefully tomorrow

cmd/deck/basepath.go
```diff
@@ -0,0 +1,113 @@
...
+	if basePath == "." {
+		return "/"
+	}
+	return basePath
+}
+
+// withBasePath ensures that only requests that begin with the configured base path are handled
+// and that downstream handlers only see paths relative to that base path.
+func withBasePath(basePath string, handler http.Handler) http.Handler {
+	basePath = normalizeBasePath(basePath)
```
petr-muller, line 45: another `normalizeBasePath` call

cmd/deck/templates.go
```diff
@@ -74,6 +74,8 @@ func prepareBaseTemplate(o options, cfg config.Getter, csrfToken string, t *temp
 		"deckVersion":      func() string { return version.Version },
 		"googleAnalytics":  func() string { return cfg().Deck.GoogleAnalytics },
 		"csrfToken":        func() string { return csrfToken },
+		"deckBasePath":     func() string { return o.basePath },
+		"deckPath":         func(p string) string { return o.absolutePath(p) },
```
petr-muller, line 78: How are these consumed?

cmd/deck/basepath.go
```diff
@@ -0,0 +1,113 @@
...
+		pathPart = "/" + pathPart
+	}
+	if o.basePath == "/" {
+		return pathPart + suffix
+	}
+	if pathPart == "/" {
+		return o.basePath + suffix
+	}
+	return o.basePath + pathPart + suffix
+}
```
petr-muller, line 106: It's not easy to reason about all path manipulation possibilities, I'd like to have unit tests (also for `stripBasePath` and `normalizeBasePath` both to make sure they are correct and to serve as examples of what gets transformed to what.

cmd/deck/basepath.go
```diff
@@ -0,0 +1,113 @@
...
+			return "/", true
+		}
+		return trimmed, true
+	}
+	return "", false
+}
+
+// absolutePath returns the absolute URL path for a given relative path by prepending the
+// configured base path. The input is expected to be either "/" or begin with a slash.
+func (o options) absolutePath(rel string) string {
```
petr-muller, line 82: This technically only needs to `basePath` from `options`, could the *field* be a custom string type and the methods would be defined on that? It would make the interface more narrow.

petr-muller (commented):
My main concern is how well this was tested on the current default case (path is "/")? If the new subpath functionality is not 100% then I don't mind because we can fix it as we go, but the one thing we need to prevent is breaking existing deployments.

Some comments inline

/hold cancel

cmd/deck/main.go
tsj-30: Will modify this.

cmd/deck/templates.go
tsj-30: deckPath is used throughout the templates to prefix links/static asset URLs and nav links.
deckBasePath is used in base.html to set window.prowBasePath. That gets consumed in the frontend JS helpers in urls.ts

cmd/deck/basepath.go
tsj-30: Makes sense — I’ll add unit tests for stripBasePath and normalizeBasePath (including tricky slash + ./.. cases) so we verify correctness and document expected transformations. Will push the code as part of this PR.

cmd/deck/basepath.go
tsj-30: Yes, that’s possible. I can refactor this so it only depends on options.basePath: introduce a basePath custom string type and move absolutePath (and related helpers) onto it. I’ll push the change as well.

tsj-30:
Fix type error in templates_test.go by adding proper fakeConfigAgent
- Fix template function call in base.html (remove printf wrapper)
- Update test expectations to match actual output

Fixes linting failure in pull-prow-verify-lint check.

PR #577, opened 2025-12-17
updateconfig: Add ACL support for ConfigMap updates

midnightconman:
This enhancement adds Access Control List (ACL) functionality to the updateconfig plugin, allowing administrators to control which organizations and repositories can trigger ConfigMap updates.

## New Features

- **AllowedRepos**: A list of org or org/repo entries that are permitted to update a configmap. If specified, only PRs from these sources will trigger updates. When empty, all repos are allowed (unless explicitly denied).

- **DeniedRepos**: A list of org or org/repo entries that are blocked from updating a configmap. Deny rules take precedence over allow rules.

## Configuration Format

Entries can be specified in two formats:
- Org-level: "kubernetes" (matches all repos in the kubernetes org)
- Repo-level: "kubernetes/test-infra" (matches only this specific repo)

Example configuration:
```yaml
config_updater:
  maps:
    prow/config.yaml:
      name: config
      allowed_repos:
        - kubernetes           # Allow entire org
        - istio/istio          # Allow specific repo
      denied_repos:
        - kubernetes/fork      # Block specific repo within allowed org
```

## ACL Precedence Rules

1. If repo is in denied list (exact match) → denied
2. If org is in denied list → denied
3. If allow lists exist, repo must match an allowed entry
4. If no allow lists specified → allowed (unless denied above)

## Implementation Details

- Added `AllowedRepos` and `DeniedRepos` fields to `ConfigMapSpec`
- Added `IsAllowed(repo string) bool` method to `ConfigMapSpec`
- Updated `FilterChanges` to accept `repo` parameter and filter updates
- Updated `config-bootstrapper` to pass empty repo (bypasses ACL checks)
- Uses `config.SplitRepoName` for org/repo parsing

## Testing

- Added `TestConfigMapSpecIsAllowed` with 14 test cases covering:
  - No ACLs configured
  - Org-level and repo-level allow/deny entries
  - Mixed org and org/repo entries
  - Deny precedence over allow
  - Empty repo handling for config-bootstrapper

- Added `TestFilterChangesWithACLs` with 7 test cases covering:
  - End-to-end filtering with various ACL configurations
  - Multiple configmaps with different ACLs

### DIFF
--- a/cmd/checkconfig/main.go
+++ b/cmd/checkconfig/main.go
@@ -123,6 +123,7 @@ const (
 	validateLabelWarning                           = "validate-label"
 	requiredJobAnnotationsWarning                  = "required-job-annotations"
 	periodicDefaultCloneWarning                    = "periodic-default-clone-config"
+	validateConfigUpdaterACLsWarning               = "validate-config-updater-acls"
 
 	defaultHourlyTokens = 3000
 	defaultAllowedBurst = 100
@@ -148,6 +149,7 @@ var defaultWarnings = []string{
 	validateLabelWarning,
 	requiredJobAnnotationsWarning,
 	periodicDefaultCloneWarning,
+	validateConfigUpdaterACLsWarning,
 }
 
 var expensiveWarnings = []string{
@@ -439,6 +441,12 @@ func validate(o options) error {
 
 	if pcfg != nil && o.warningEnabled(validateLabelWarning) {
 		if err := verifyLabelPlugin(pcfg.Label); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if pcfg != nil && o.warningEnabled(validateConfigUpdaterACLsWarning) {
+		if err := validateConfigUpdaterACLs(pcfg); err != nil {
 			errs = append(errs, err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


Prucek:
I would like this not to be a breaking change, so I would like to see a passing test, when the `config-updater` uses the old version of the config, without `AllowedRepos` and `DeniedRepos`.

midnightconman:
@Prucek Sounds good to me 👍 Can you add `/ok-to-test`, so we can test that?

midnightconman:
@Prucek I built the checkconfig binary image locally and tested it against a config without any `updateconfig` ACLs, it didn't return any errors:

```
$ docker run -it --rm -v $PWD:/src ko.local/checkconfig:ko-v20260105-63cbfe640 --job-config-path jobs --strict --expensive-checks ; echo $?

{"component":"checkconfig","file":"sigs.k8s.io/prow/cmd/checkconfig/main.go:254","func":"main.main","level":"info","msg":"checkconfig passes without any error!","severity":"info","time":"2026-01-05T21:15:23Z"}
0
```

Anything else we should check?

midnightconman:
@petr-muller / @cjwagner Do y'all see anything else we should do prior to reviewing / approving this?

Happy New Year by the way! 🍾

matthyx (approved):
thanks @midnightconman !

PR #578, opened 2025-12-18
Clarify untrusted owners message

lentzi90:
I would like to clarify this message, since it wasn't clear to me that the untrusted user was the reason that the OWNERS file was considered invalid.


matthyx:
fair enough
/ok-to-test

matthyx (approved):
/lgtm

PR #579, opened 2025-12-19
tide: add configurable GitHub merge blocks enforcement

### LINKED ISSUE #575: PR #537 prevents `tide` from merging PRs in repos which `Restrict updates` via Github Rulesets
Unfortunately #537 breaks tide for most of our repositories.

We use a Github Ruleset to prevent anyone but tide from merging PRs.
The Ruleset [restrict updates](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/available-rules-for-rulesets#restrict-updates) to the branch and lists the tide Github App on the `Bypass list` as the documentation suggests. This was working until tide was updated today to the latest version.

Since then tide shows red lights for all PRs showing status:
`PR is blocked from merging by GitHub (check branch protection, required reviews, or rulesets)`

After we deactivated the Ruleset, tide was able to merge again.
After we rolled back to the previous version of tide, we are able to activate the Ruleset again.

Prucek:
Adds a new `github_merge_blocks_policy` configuration option that allows                                                                                                                                                            
Tide to check GitHub's `mergeStateStatus` to prevent merging PRs that are                                                                                                                                                              
blocked by GitHub due to branch protection rules, rulesets, required                                                                                                                                                                 
reviews, etc.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                     
The configuration can be set globally with '*' or per-org/repo level.                                                                                                                                                                
Defaults to "permit".
 Valid values:
   - "ignore": Ignore BLOCKED status entirely, attempt to merge anyway (risky)
   - "permit": Allow merging but log warnings and show "In merge pool (despite BLOCKED)"
               status for monitoring. Use this to identify repos needing BP/ruleset fixes.
   - "block": Respect BLOCKED status and prevent Tide from merging (safest)
    
Fixes https://github.com/kubernetes-sigs/prow/issues/575, https://github.com/kubernetes-sigs/prow/issues/269
Previous attempt: https://github.com/kubernetes-sigs/prow/pull/537

🤖 Assisted by Claude.

### DIFF
--- a/pkg/config/prow-config-documented.yaml
+++ b/pkg/config/prow-config-documented.yaml
@@ -1560,6 +1560,13 @@ tide:
               org: ' '
               repos:
                 - ""
+    # GitHubMergeBlocksPolicyMap configures on org or org/repo level how Tide should handle
+    # GitHub's mergeStateStatus (BLOCKED state from branch protection rules, rulesets,
+    # required reviews, etc.).
+    # Use '*' as key to set this globally. Defaults to "permit" which allows merging but logs warnings.
+    # Valid values: "ignore", "permit", "block"
+    github_merge_blocks_policy:
+        "": ""
     # A key/value pair of an org/repo as the key and Go template to override
     # the default merge commit title and/or message. Template is passed the
     # PullRequest struct (prow/github/types.go#PullRequest)
--- a/pkg/config/tide.go
+++ b/pkg/config/tide.go
@@ -39,6 +39,25 @@ import (
 // TideQueries is a TideQuery slice.
 type TideQueries []TideQuery
 
+// GitHubMergeBlocksPolicy describes how Tide should handle GitHub's merge blocking conditions.
+type GitHubMergeBlocksPolicy string
+
+const (
+	// GitHubMergeBlocksIgnore ignores GitHub's BLOCKED status entirely in merge decisions.
+	// Tide will attempt to merge PRs even if GitHub reports them as blocked.
+	GitHubMergeBlocksIgnore GitHubMergeBlocksPolicy = "ignore"
+
+	// GitHubMergeBlocksPermit allows merging of BLOCKED PRs but logs warnings and may
+	// surface the blocked state in PR status messages. This is useful for monitoring
+	// repos that need branch protection or ruleset fixes.
+	GitHubMergeBlocksPermit GitHubMergeBlocksPolicy = "permit"
+
+	// GitHubMergeBlocksBlock respects GitHub's BLOCKED status and prevents Tide from
+	// attempting to merge PRs that are blocked by branch protection rules, rulesets,
+	// required reviews, or other GitHub-side blocking conditions.
+	GitHubMergeBlocksBlock GitHubMergeBlocksPolicy = "block"
+)
+
 type TideBranchMergeType struct {
 	MergeType types.PullRequestMergeType
 	Regexpr   *regexp.Regexp
@@ -227,6 +246,12 @@ type Tide struct {
 	// starting a new one requires to start new instances of all tests.
 	// Use '*' as key to set this globally. Defaults to true.
 	PrioritizeExistingBatchesMap map[string]bool `json:"prioritize_existing_batches,omitempty"`
+	// GitHubMergeBlocksPolicyMap configures on org or org/repo level how Tide should handle
+	// GitHub's mergeStateStatus (BLOCKED state from branch protection rules, rulesets,
+	// required reviews, etc.).
+	// Use '*' as key to set this globally. Defaults to "permit" which allows merging but logs warnings.
+	// Valid values: "ignore", "permit", "block"
+	GitHubMergeBlocksPolicyMap map[string]GitHubMergeBlocksPolicy `json:"github_merge_blocks_policy,omitempty"`
 
 	TideGitHubConfig `json:",inline"`
 }
@@ -368,6 +393,22 @@ func (t *Tide) BatchSizeLimit(repo OrgRepo) int {
 		return limit
 	}
 	return t.BatchSizeLimitMap["*"]
+}
+
+// GitHubMergeBlocksPolicy returns the policy for handling GitHub's merge blocking conditions.
+// The default is "permit" which allows merging but logs warnings for monitoring.
+func (t *Tide) GitHubMergeBlocksPolicy(repo OrgRepo) GitHubMergeBlocksPolicy {
+	if val, set := t.GitHubMergeBlocksPolicyMap[repo.String()]; set {
+		return val
+	}
+	if val, set := t.GitHubMergeBlocksPolicyMap[repo.Org]; set {
+		return val
+	}
+	if val, set := t.GitHubMergeBlocksPolicyMap["*"]; set {
+		return val
+	}
+	// Default to "permit" to allow merging while surfacing the blocked state for monitoring
+	return GitHubMergeBlocksPermit
 }
 
 // MergeMethod returns the merge method to use for a repo. The default of merge is
--- a/pkg/tide/github.go
+++ b/pkg/tide/github.go
@@ -632,6 +632,18 @@ func (m *mergeChecker) isAllowedToMerge(crc *CodeReviewCommon) (string, error) {
 	} else if !allowed {
 		return fmt.Sprintf("Merge type %q disallowed by repo settings", *mergeMethod), nil
 	}
+	if pr.MergeStateStatus == MergeStateStatusBlocked {
+		switch policy := m.config().Tide.GitHubMergeBlocksPolicy(orgRepo); policy {
+		case config.GitHubMergeBlocksBlock:
+			return "PR is blocked from merging by GitHub (check branch protection, required reviews, or rulesets)", nil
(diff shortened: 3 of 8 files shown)


petr-muller:
Thanks for following up on this! Opt-in config seems like a reasonable short-term way forward. Some thoughts (have not yet looked at the code):

I'm not sure I feel comfortable leaving default-configured (and even explicitly configured) instances vulnerable to the class of #269 -like bugs. Some repos may have policies that help them not getting affected by some instances of this problem but not necessarily to all, and the "tide silently fails to merge and never surfaces the problem" error is really annoying to leave some instances open to it. So to me it seems we'd still need to invent some way to propagate the actual merge failure, post-fact, to PR author _somehow_, I don't think we could consider #269 -like issues fully fixed :thinking: 

Configuration booleans are always a bit of a smell, for similar reasons like [API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) states:

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

For this specific case, I think we may prefer treating this as a three-value for starters:

- ignore: ignore the `BLOCKED` entirely in decisions 
- permit: merge `BLOCKED` PRs but make some noise about it, possibly in the log, possibly through some PR-exposed message like "In Merge Pool (despite BLOCKED)". This would be a good default to set us up for eventual flip the default to enforcing, because we'd help Prow and/or repo admins to monitor repos that either need BP/Ruleset fixes or explicit ignore/permit config
- block: respect `BLOCKED`

We'd start with `permit` as a default and possibly eventually flip the default to `block`?

Prucek:
Thanks @petr-muller for the guidance! That is definitely a good point. I updated the PR based on your comments.

pkg/config/prow-config-documented.yaml
```diff
@@ -1560,6 +1560,15 @@ tide:
               org: ' '
               repos:
                 - ""
+    # GitHubMergeBlocksPolicyMap configures on org or org/repo level how Tide should handle
+    # GitHub's mergeStateStatus (BLOCKED state from branch protection rules, rulesets,
+    # required reviews, etc.).
+    # Use '*' as key to set this globally. Defaults to "permit" which allows merging but logs warnings.
+    # Valid values: "ignore", "permit", "block"
+    # Note: "ignore" may cause issues with repos using GitHub Rulesets that restrict updates
```
petr-muller: What kind of issues may `ignore` cause? "Restrict updates" in what sense?

pkg/config/tide.go
```diff
@@ -227,6 +246,14 @@ type Tide struct {
...
 	// Use '*' as key to set this globally. Defaults to true.
 	PrioritizeExistingBatchesMap map[string]bool `json:"prioritize_existing_batches,omitempty"`
+	// GitHubMergeBlocksPolicyMap configures on org or org/repo level how Tide should handle
+	// GitHub's mergeStateStatus (BLOCKED state from branch protection rules, rulesets,
+	// required reviews, etc.).
+	// Use '*' as key to set this globally. Defaults to "permit" which allows merging but logs warnings.
+	// Valid values: "ignore", "permit", "block"
+	// Note: "ignore" may cause issues with repos using GitHub Rulesets that restrict updates
+	// but have Tide on the bypass list.
+	GitHubMergeBlocksPolicyMap map[string]GitHubMergeBlocksPolicy `json:"github_merge_blocks_policy,omitempty"`
```
petr-muller, line 254: I think config load and `checkconfig` should validate the value is one of the three

pkg/tide/github.go
```diff
@@ -632,6 +632,20 @@ func (m *mergeChecker) isAllowedToMerge(crc *CodeReviewCommon) (string, error) {
...
+	if pr.MergeStateStatus == "BLOCKED" {
+		policy := m.config().Tide.GitHubMergeBlocksPolicy(orgRepo)
+		switch policy {
+		case config.GitHubMergeBlocksBlock:
+			return "PR is blocked from merging by GitHub (check branch protection, required reviews, or rulesets)", nil
+		case config.GitHubMergeBlocksPermit:
+			// Allow merge but the warning will be surfaced in PR status by requirementDiff
+		case config.GitHubMergeBlocksIgnore:
+			// Ignore BLOCKED status entirely
+		}
```
petr-muller, line 645: Do something loud in `default:`, meaning an unexpected value somehow slipped through (programmer error or botched validation)

pkg/tide/status.go
```diff
@@ -260,6 +261,27 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
...
+	// This is controlled by the github_merge_blocks_policy configuration.
+	if tide != nil && pr.MergeStateStatus == "BLOCKED" {
+		orgRepo := config.OrgRepo{
+			Org:  string(pr.Repository.Owner.Login),
+			Repo: string(pr.Repository.Name),
+		}
+		policy := tide.GitHubMergeBlocksPolicy(orgRepo)
+		switch policy {
+		case config.GitHubMergeBlocksBlock:
+			// Block merge by adding to diff
```
petr-muller: this is misleading, this is status controller - does not actually block merges

pkg/tide/status.go
```diff
@@ -260,6 +261,27 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
...
+			diff += 100
+			if desc == "" {
+				desc = " Blocked by GitHub (branch rulesets or protection)"
```
petr-muller: Not something that should be addressed right away in this PR, but we'll want to investigate if GH gives the actual reason somewhere in the API. It gives Tide a fairly informative message on an actual denied merge attempt, like

```
At least 2 approving reviews are required by reviewers with write access.
```

It would be great if Tide showed that.

pkg/tide/status.go
```diff
@@ -260,6 +261,27 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker) (s
...
 		}
 	}
+	// Check GitHub's mergeStateStatus which reflects all GitHub-side blocking conditions
+	// This is controlled by the github_merge_blocks_policy configuration.
+	if tide != nil && pr.MergeStateStatus == "BLOCKED" {
+		orgRepo := config.OrgRepo{
+			Org:  string(pr.Repository.Owner.Login),
+			Repo: string(pr.Repository.Name),
+		}
+		policy := tide.GitHubMergeBlocksPolicy(orgRepo)
```
petr-muller: `requirementDiff` only needs the policy, but we give it the full tide config and we even need to reconstruct an `OrgRepo` to call the method to get the policy. The `requirementDiff` can receive just the policy, and the callsite (`expectedStatus` on L340) even has the necessary `OrgRepo` struct so ti can easily do

```go
mergeBlocksPolicy := sc.config().Tide.GitHubMergeBlocksPolicy(orgRepo)
for _, q := range queryMap.ForRepo(repo) {
  diff, diffCount := requirementDiff(pr, &q, cc, mergeBlocksPolicy)
```

pkg/tide/status.go
```diff
@@ -357,7 +379,30 @@ func (sc *statusController) expectedStatus(log *logrus.Entry, queryMap *config.Q
...
 		return github.StatePending, retestingStatus(diff), nil
 	}
-	return github.StatusSuccess, statusInPool, nil
+	return github.StatusSuccess, poolStatus(pr, &sc.config().Tide, log), nil
+}
+
+// poolStatus returns the appropriate status message for a PR that is in the merge pool.
+// If the PR has BLOCKED merge state and the policy is "permit", it returns a warning message.
+// This also logs a warning for monitoring purposes.
+func poolStatus(pr *PullRequest, tide *config.Tide, log *logrus.Entry) string {
```
petr-muller: this also only needs the policy, not the full tide config, and both callsites have access to the policy computed to call `requirementDiff` (we'd just need to compute the policy earlier, before the conditional on L316

pkg/tide/github.go
```diff
@@ -632,6 +632,20 @@ func (m *mergeChecker) isAllowedToMerge(crc *CodeReviewCommon) (string, error) {
 	} else if !allowed {
 		return fmt.Sprintf("Merge type %q disallowed by repo settings", *mergeMethod), nil
 	}
+	// Check GitHub's mergeStateStatus which reflects all GitHub-side merge blocking conditions
+	// including branch protection rules, rulesets, required reviews, status checks, etc.
+	// This is controlled by the github_merge_blocks_policy configuration.
+	if pr.MergeStateStatus == "BLOCKED" {
```
petr-muller: `"BLOCKED"` should be a constant somewhere, maybe close to `PullRequest` type

pkg/tide/status.go
Prucek: I think there is no API to get it before the merge. We can maybe add the status after it tries to merge, from the mergeError.

petr-muller (approved):

Prucek:
🚀 ❤️

PR #580, opened 2025-12-22
Update go to latest

prad9192:
Update go to latest

### DIFF
--- a/pkg/genyaml/genyaml.go
+++ b/pkg/genyaml/genyaml.go
@@ -76,6 +76,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"unicode"
 
 	"github.com/clarketm/json"
 	yaml3 "gopkg.in/yaml.v3"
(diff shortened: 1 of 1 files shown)


pkg/config/prow-config-documented.yaml
```diff
@@ -1,255 +1,157 @@
 branch-protection:
-    # AllowDeletions allows deletion of the protected branch by anyone with write access to the repository.
```
stmcginnis, line 2: I don't think we want to remove the documentation from the "documented" example file.

prad9192: Ack, the comments were preserved with 1.24, I will review and fix forward whats causing this in 1.25.

prad9192:
@stmcginnis @petr-muller - I have addressed the feedback, could I please get couple reviews?

stmcginnis:
Thanks for figuring out the annotations. Looks like things are working now, and we need to handle it correctly with the newer Go version anyway, so this looks good to me.

/lgtm

PR #581, opened 2025-12-23
Fix message when PR needs ok-to-test

Michcioperz:
spotted this reordering in https://github.com/kubernetes/dns/pull/737#issuecomment-3686585465


petr-muller (approved):
LGTM thanks!

Issue #545, opened 2025-11-12
An unknown state job break job-bar on "Prow Status" page


liangxia:
A good page,
<img width="3440" height="322" alt="Image" src="https://github.com/user-attachments/assets/b8939a57-38f8-40ae-8d96-8f79c744d9cc" />

A page that with unknown state job,
<img width="3440" height="322" alt="Image" src="https://github.com/user-attachments/assets/62dd39a8-0e39-4bfc-9a32-5eb30b229ffa" />

petr-muller:
/area deck
/kind bug

@liangxia when you say "unknown state job" were you able to find the actual single ProwJob resource that caused this? Any chance you have some features of that job saved somewhere?

liangxia:
Not sure how to locate that "unknown" job. But you can see it here, https://prow.ci.openshift.org/

petr-muller:
The culprit was a (likely manually?) created ProwJob without any `status`, for some reason I cannot attach it here (`<!-- Failed to upload "hypershift-images.yaml" -->`). I was also seeing and empty item in the state filter dropdown in Deck. The frontend code should probably filter out anything that does not have `.status` or is unexpectedly missing the structs it reads from it.

/label help-needed

Qqkyu:
@petr-muller would this be a good first issue? If so, I'm down to give it a try as I wanted to dive into prow code a bit anyways

petr-muller:
@Qqkyu yeah, feel free to give this a shot if still interested. I _think_ this should not be too tricky, seems like what we need is to filter out garbage Prowjobs (like the one without any status, mentioned in my comment above) from the set used to compute the visuals

petr-muller:
It would probably be nice if `sinker` removed the garbage Prowjob eventually 🤔

Qqkyu:
Fixed in https://github.com/kubernetes-sigs/prow/pull/582

PR #582, opened 2025-12-26
Fix `An unknown state job break job-bar on "Prow Status" page`

Qqkyu:
Fix: https://github.com/kubernetes-sigs/prow/issues/545

This should fix the deck bug, but I wasn't 100% sure on the safe logic for Sinker cleanup (https://github.com/kubernetes-sigs/prow/issues/545#issuecomment-3687445345) - I assume some time should pass to make sure it's abandoned/garbage job and not just a new one, but I'm not sure what that required time should be. I'm happy to add this logic here if I get confirmation on this approach and a tip on the required time amount.


Qqkyu:
cc: @petr-muller

PR #583, opened 2026-01-05
`peribolos`:  Add Repository Fork Management Support

hoxhaeris:
## Summary

This PR adds support for managing GitHub repository forks in Peribolos. It allows users to declaratively configure forks alongside standard repositories, handling creation, metadata configuration, and edge cases like naming conflicts.

## Key Features

- **Declarative Fork Management:** Define forks using a `fork` config block with a `from` field.
- **Full Metadata Support:** Configure description, homepage, privacy, and other settings for forks just like regular repos.
- **Fork Name Enforcement:** The config key becomes the fork name via GitHub's `name` API parameter.
- **Race Condition Handling:** Intelligently waits for asynchronous fork creation on GitHub to complete before attempting to apply settings.

### Basic Fork
Create a fork using the config key as the fork name.

```yaml
orgs:
  my-org:
    repos:
      # Will create my-org/kubernetes as a fork of kubernetes/kubernetes
      kubernetes:
        fork:
          from: kubernetes/kubernetes
```

### Customized Fork (Hard Fork / Mirror)
Create a fork with a custom name, description, and specific settings (e.g., disabling issues).

```yaml
orgs:
  my-org:
    repos:
      # Create 'internal-k8s' as a fork of 'kubernetes/kubernetes'
      internal-k8s:
        fork:
          from: kubernetes/kubernetes
          default_branch_only: true

        # Metadata is applied immediately after creation
        description: "Internal mirror of Kubernetes - DO NOT OPEN PRS HERE"
        homepage: "https://internal.docs/k8s"

        # Configure repo settings
        has_issues: false  # Disable issues to force users upstream
        has_projects: false
        has_wiki: false
        allow_merge_commit: true
        allow_squash_merge: false

        # Configure access
        collaborators:
          ci-bot: admin
```
## Usage

Fork management is enabled automatically when you fix repositories.

```bash
# Standard run - manages org, teams, and repos (including forks)
peribolos --config-path org.yaml --fix-org --fix-teams --fix-repos

# Target only forks specifically (if needed)
peribolos --config-path org.yaml --fix-forks

# Explicitly exclude forks
peribolos --config-path org.yaml --fix-org --fix-teams --fix-repos --fix-forks=false
```

## Behavior

| Scenario | Behavior |
|----------|----------|
| Repo doesn't exist | Creates fork from upstream, waits for readiness, then applies metadata |
| Repo exists as correct fork | Updates metadata (idempotent) |
| Repo exists but not a fork | Error |
| Repo exists as fork of different upstream | Error (safe failure) |
| Multiple config entries fork from the same upstream | Error (GitHub only allows one fork per upstream per org) |
| Invalid `fork.from` format | Error |
| Upstream repo doesn't exist | Error |
| Some forks fail, others succeed | Logs errors for failed forks, continues with successful ones, exits non-zero |

> **Note on "repo already exists but is not a fork":** This can only happen if someone creates a repository manually on GitHub with the same name as a fork defined in your config. It cannot occur if all repositories are managed through Peribolos, since YAML does not allow duplicate keys. To resolve, either rename/delete the manually-created repository, or choose a different name for the fork in your config.

## Metadata Inheritance

Forks inherit metadata from the upstream repository:
- Fields explicitly set in config **override** the inherited/current value
- Fields omitted (nil) **retain** the fork's current value
- Metadata restrictions vary by GitHub edition; restricted changes result in API errors

## Implementation Details

- **Ordering:** Fork configuration runs *before* repository configuration to ensure forks exist before metadata is applied.
- **Fork Naming:** The config key is passed to GitHub's fork API `name` parameter, enforcing the exact fork name. If a repository with this name already exists, the operation fails with an error.
- **Error Propagation:** Fork errors are deferred (other subsystems still run) but propagated before `configureOrg` returns, so peribolos exits non-zero.
- **API Efficiency:** `GetRepos` is only called when fork configs actually exist. Validation (format, duplicate upstream) runs before any API calls.
- **Race Condition Handling:** `configureForks` polls GitHub every 10 seconds until the fork is fully created (up to 5 minutes) before proceeding. Only retries on 404 (fork not ready); other errors fail immediately.
- **Refactoring:** Includes `RepoMetadata` struct extraction (separate commit) and `ForkConfig` struct to group fork-related fields.

## API Reference

- [GitHub Forks API](https://docs.github.com/en/rest/repos/forks)
- [GitHub Repos API](https://docs.github.com/en/rest/repos/repos#update-a-repository)


Prucek (approved):
/ok-to-test

hoxhaeris:
/hold
I want to watch the rollout.

bradmwilliams:
I have reviewed this PR and would like to recommend it for addition to the codebase.  The changes aligns well with existing patterns, have adequate test coverage, and are protected by the `--fix-forks` flag. 
Thanks for your consideration!

matthyx:
hi, the PR seems well written indeed, just wondering why we need the flag `--fix-forks`

hoxhaeris:
> hi, the PR seems well written indeed, just wondering why we need the flag `--fix-forks`

Mostly to have more granular control (although most users would want them both, `--fix-forks` and `--fix-repos`). A fork is still a repo, and removing the flag would offer a simpler design. Also, since this is a new feature, gating it made sense at that time, but thinking about it more carefully, I believe we can drop the flag.

cmd/peribolos/main_test.go
```diff
@@ -3982,3 +3992,336 @@ func TestConfigureCollaborators_PermissionMatrix_PendingInvitationUpdates(t *tes
...
+	}
+}
+
+func boolPtr(b bool) *bool {
+	return &b
+}
+
+func strPtr(s string) *string {
+	return &s
+}
```
petr-muller: Don't bother respinning the PR for this but there's generic `ptr.To` now: https://pkg.go.dev/k8s.io/utils/ptr#To (also useful in other codebases that import k8s utils ;) )

cmd/peribolos/main.go
```diff
@@ -1139,6 +1160,107 @@ func configureRepos(opt options, client repoClient, orgName string, orgConfig or
...
+			allErrors = append(allErrors, err)
+			continue
+		}
+
+		// Note: GitHub may name the fork differently if there's a naming conflict
+		if createdName != repoName {
+			repoLogger.WithField("created_name", createdName).Warn("fork was created with a different name than expected")
+		} else {
+			repoLogger.Info("fork created successfully")
+		}
```
petr-muller: I'm not sure if I correctly understand how this is idempotent. We config a fork to exist in our subject org, after first run it is created under an arbitrary name. How do the subsequent runs avoid trying to recreate a fork if the existing one has an arbitrary name not present in the config?

petr-muller (approved):
Just want to clarify some things, so I'm approving with a hold.
/hold

```
    my-fork:
      fork_from: kubernetes/kubernetes
      default_branch_only: true
      description: Our fork of **Kubernetes**
```

Is this snippet configuring a `my-org/my-fork` to exist, or `my-org/kubernetes`? IIUC it is the former, right?

petr-muller:
> Also, since this is a new feature, gating it made sense at that time, but thinking about it more carefully, I believe we can drop the flag.

I'm leaning towards keeping it, maybe make it inherit whatever `--fix-repos` is unless `--fix-forks` is explicitly passed? It's a pattern in peribolos and I think new larger reconciled GH entities with sufficiently custom logic benefit from a flag. But I do not really have strong opinion on this, either case is fine.

petr-muller:
/reopen

dammit why are the buttons so close to each other

cmd/peribolos/main.go
hoxhaeris: Ah, good point, this is a bug.

hoxhaeris: The current code assumes the fork will be named exactly as specified in the config, and only looks up existing repos by that config name.

hoxhaeris:
> I'm leaning towards keeping it, maybe make it inherit whatever `--fix-repos` is unless `--fix-forks` is explicitly passed? It's a pattern in peribolos and I think new larger reconciled GH entities with sufficiently custom logic benefit from a flag. But I do not really have strong opinion on this, either case is fine.

I'm fine either way as well. The middle ground sounds good -- having `--fix-forks` inherit from `--fix-repos` unless explicitly passed. This keeps the pattern consistent while providing sensible defaults.

Users who want repos managed but not forks can still do `--fix-repos --fix-forks=false`.

I'll update the implementation to follow this approach.

cmd/peribolos/main.go
hoxhaeris: I also need to implement a wait loop here. CreateForkInOrg returns a 202 Accepted immediately because fork creation is asynchronous on GitHub. Currently, we proceed to the next step without verifying the repository exists/is ready. This creates a race condition where subsequent configuration steps (like applying permissions) may fail with a 404 if the fork isn't fully ready yet.

cmd/peribolos/main_test.go
hoxhaeris: Since I had to respin the PR for a few other fixes anyway, I went ahead and made this change as well. Done in [04ef97729](https://github.com/kubernetes-sigs/prow/pull/583/commits/04ef977299df26b209887d231c06f07e068e9d7f)

hoxhaeris:
> Just want to clarify some things, so I'm approving with a hold. /hold
> 
> ```
>     my-fork:
>       fork_from: kubernetes/kubernetes
>       default_branch_only: true
>       description: Our fork of **Kubernetes**
> ```
> 
> Is this snippet configuring a `my-org/my-fork` to exist, or `my-org/kubernetes`? IIUC it is the former, right?

cc @petr-muller 
The config key (`my-fork`) is a YAML identifier, but GitHub determines the actual fork name. By default, GitHub names forks after the upstream repo, so this would create `my-org/kubernetes` (if a repo with the same name does not exist on `my-org`, in that case it would likely be `kubernetes-1`), not my-org/my-fork.

The config key is useful for YAML uniqueness - for example, if we already have a repo named kubernetes and want to fork kubernetes/kubernetes, we can't use duplicate keys in YAML.

Current behavior (admittedly weird):
  - Fork is created with whatever name GitHub assigns (typically the upstream repo name)
  - Collaborators are applied correctly (we track the actual name internally)
  - Metadata fields (description, private, etc.) are only applied if the config key happens to match the actual fork name - otherwise they're silently ignored
  - Even when metadata is applied, some fields will fail due to GitHub restrictions (e.g., "public forks can't be made private")

Planned fix: I'm leaning towards supporting non-restrictive metadata changes on forks (e.g., description, has_issues, has_wiki) since those are valid use cases. However, restrictive fields like private should be rejected or warned about, since GitHub doesn't allow making public forks private (although that might not be true for Enterprise, and that is an argument to actually allow all metadata modification, and let the tool fail if the modification is restricted, since trying to anticipate what GitHub will reject is fragile).

I'll also need to fix the current issue where metadata is only applied when the config key matches the actual fork name - it should work regardless of what GitHub names the fork.

Wanted to get your opinion: does supporting non-restrictive metadata on forks make sense, or should we keep forks minimal (fork_from, default_branch_only, collaborators only)?

We can also add support for the name parameter in GitHub's fork API, which would let users enforce a specific fork name via the config key (I am also leaning towards this, more explicit, follows the current behaviour of normal repos).

>(Note: I’ve already implemented what I mentioned above, but I kept the commits split in case we need to revert any.)

petr-muller:
> Wanted to get your opinion: does supporting non-restrictive metadata on forks make sense, or should we keep forks minimal (fork_from, default_branch_only, collaborators only)?

Thanks for pointing this out, that's an interesting angle. If we can support non-restrictive metadata in some way then I'd prefer to do so. Seems like a valid use case to me but also not something that we need to get 100% right here in the initial PR, so I do not worry about that too much.

> We can also add support for the name parameter in GitHub's fork API, which would let users enforce a specific fork name via the config key (I am also leaning towards this, more explicit, follows the current behaviour of normal repos).

This is what my original "Is this snippet configuring a my-org/my-fork to exist, or my-org/kubernetes?" comment was concerned about. I think that if possible, the declarative peribolos config should be as explicit as possible, ideally specifying 'I want a fork of upstream/repo and I want it named `repo`'. The state where we'd only specify "I want a fork" and the fork would be named in an arbitrary, GitHub-side selected way is IMO worse than requiring the config author to make a decision (or fail if they rely on the default which cannot be fulfilled)

I wonder if there's a more convenient config surface for this, like making `fork_from` a key:

```
forks:
  kubernetes/kubernetes:
    default_branch_only: true
    description: Our fork of **Kubernetes**
    optional_custom_name: k8s
```

The `optional_custom_name` would be optional, but if it was not used, we'd assume the same (`kubernetes`) name client-side and therefore always use the specific name in the API call (basically never rely on GH-side name selection).

I have not yet reviewed the code increment, hope I get to it this week.

cmd/peribolos/main.go
```diff
@@ -1137,6 +1206,162 @@ func configureRepos(opt options, client repoClient, orgName string, orgConfig or
...
+		time.Sleep(interval)
+	}
+
+	return fmt.Errorf("timeout waiting for fork %s/%s to become available after %v", org, repo, timeout)
+}
+
+// configureForks creates repository forks from upstream repositories as specified in the config.
+// This function only creates forks - it does not delete existing forks that are not in the config.
+// Returns a mapping of config repo names to actual GitHub repo names (for forks that were renamed).
+func configureForks(client forkClient, orgName string, orgConfig org.Config) (map[string]string, error) {
```
ivankatliarchuk, line 1354: fork detection by upstream is expensive and potentially incorrect.

The function fetches every repo in the org and calls GetRepo on each to find which ones are forks and what their parent is. This is O(n) API calls where n = number of repos in the org. For large orgs (100s of repos) this will be very slow and may hit GitHub rate limits. A better approach: only GetRepo on repos whose names match config entries, or check IsForked/Fork field from the GetRepos list response

https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-organization-repositories

cmd/peribolos/main.go
```diff
@@ -1137,6 +1206,162 @@ func configureRepos(opt options, client repoClient, orgName string, orgConfig or
...
+}
+
+// waitForFork polls until the fork repository is available.
+// GitHub's fork API returns HTTP 202 (accepted) and creates the fork asynchronously.
+// This function polls GetRepo until the fork exists or the timeout is reached.
+func waitForFork(client forkClient, org, repo string, timeout, interval time.Duration) error {
+	deadline := time.Now().Add(timeout)
+	logger := logrus.WithFields(logrus.Fields{"org": org, "repo": repo})
+
+	for time.Now().Before(deadline) {
```
ivankatliarchuk, line 1313: unbounded retry on transient errors

The function returns an error only on timeout, but doesn't distinguish between "fork not ready yet" (404) and other transient API errors. If there's a persistent API error, it will silently retry for 5 minutes before failing. It should surface non-404 errors immediately.

cmd/peribolos/main.go
hoxhaeris: Thanks for flagging this. I've reworked the detection logic so it no longer scans all org repos.

The current implementation uses `GetRepos` (single list call) to build a name lookup map, then for each configured fork it only checks at most 2 candidates: the config key name and the upstream repo name. The `Fork` field from the list response is checked before any `GetRepo` call, so non-fork repos never trigger an API call.

Total cost: 1 `GetRepos` + at most 2 `GetRepo` calls per configured fork, not O(n).

cmd/peribolos/main.go
hoxhaeris: Good catch, thanks. `waitForFork` now distinguishes 404 from other errors.

hoxhaeris:
> > Wanted to get your opinion: does supporting non-restrictive metadata on forks make sense, or should we keep forks minimal (fork_from, default_branch_only, collaborators only)?
> 
> Thanks for pointing this out, that's an interesting angle. If we can support non-restrictive metadata in some way then I'd prefer to do so. Seems like a valid use case to me but also not something that we need to get 100% right here in the initial PR, so I do not worry about that too much.
> 
> > We can also add support for the name parameter in GitHub's fork API, which would let users enforce a specific fork name via the config key (I am also leaning towards this, more explicit, follows the current behaviour of normal repos).
> 
> This is what my original "Is this snippet configuring a my-org/my-fork to exist, or my-org/kubernetes?" comment was concerned about. I think that if possible, the declarative peribolos config should be as explicit as possible, ideally specifying 'I want a fork of upstream/repo and I want it named `repo`'. The state where we'd only specify "I want a fork" and the fork would be named in an arbitrary, GitHub-side selected way is IMO worse than requiring the config author to make a decision (or fail if they rely on the default which cannot be fulfilled)
> 
> I wonder if there's a more convenient config surface for this, like making `fork_from` a key:
> 
> ```
> forks:
>   kubernetes/kubernetes:
>     default_branch_only: true
>     description: Our fork of **Kubernetes**
>     optional_custom_name: k8s
> ```
> 
> The `optional_custom_name` would be optional, but if it was not used, we'd assume the same (`kubernetes`) name client-side and therefore always use the specific name in the API call (basically never rely on GH-side name selection).
> 
> I have not yet reviewed the code increment, hope I get to it this week.

Hey @petr-muller, apologies for the slow follow-up on this. And thank you for the feedback.

On the `fork_from`-as-key idea: I don't have a strong preference and I can go either way. That said, here's what I've been thinking about the trade-offs:

The current design (forks under `repos:`) has the config key serve as the explicit fork name. When you write:

```yaml
repos:
  kubernetes:
    fork_from: "kubernetes/kubernetes"
```

the repo name is `kubernetes`, that's what gets passed to GitHub's fork API `name` parameter. There's nothing implicit or GitHub-selected about the naming now. 

A separate `forks:` section with upstream-as-key is nice for readability, but it introduces some friction:

- `Team.Repos` references repos by name. A separate `forks:` section means team configs need to cross-reference fork names, which adds indirection.
- Fork repos need the same metadata and collaborator config as regular repos. Keeping them under `repos:` means all of that works without duplication.
- If `optional_custom_name` is omitted, the name is derived from the upstream key, which is actually more implicit than having the user type the desired name as the config key.

Also worth noting: we now support overriding metadata on forks (description, has_issues, private, etc.). After the fork is created, `configureRepos` picks it up, computes a delta against its current state, and applies any metadata fields that are explicitly set in the config. Fields left unset (nil) keep their current value (inherited from upstream). If a change isn't supported by GitHub (e.g., making a public fork private on github.com), the API will return an error and peribolos will report it, so invalid config fails explicitly rather than silently.

That said, I see the appeal of the upstream-as-key approach for discoverability, it puts the "what are we forking" front and center. Happy to discuss further or rework if you feel strongly about it.

pkg/config/org/org.go
```diff
@@ -67,16 +68,57 @@ type Repo struct {
...
+	//   - If a RepoMetadata field is set in config, it overrides the inherited/current value.
+	//   - If a RepoMetadata field is not set (nil), the fork keeps its current value.
+	//   - Some metadata changes may be restricted by GitHub (e.g., public forks cannot be
+	//     made private on GitHub.com). Such restrictions vary by GitHub edition (Enterprise
+	//     may allow more). Restricted changes will result in an API error.
+	ForkFrom *string `json:"fork_from,omitempty"`
+
+	// DefaultBranchOnly specifies whether to fork only the default branch.
+	// Only applicable when ForkFrom is set.
+	DefaultBranchOnly *bool `json:"default_branch_only,omitempty"`
```
petr-muller: If the fork-related fields are coupled like this, can we couple them in a struct, like `ForkConfig` or something?

```
fork:
  from: org/repo # invalid to be empty
  default_branch_only: false
```

pkg/config/org/org.go
```diff
@@ -67,16 +68,57 @@ type Repo struct {
...
+	Archived                 *bool   `json:"archived,omitempty"`
+}
 
-	DefaultBranch *string `json:"default_branch,omitempty"`
-	Archived      *bool   `json:"archived,omitempty"`
+// Repo declares a repository and its configuration.
+type Repo struct {
+	// RepoMetadata contains fields that map directly to the GitHub Repos API.
+	// See https://docs.github.com/en/rest/repos/repos#update-a-repository
+	RepoMetadata
```
petr-muller, line 99: I am a bit worried that this will break external Go tools managing (even in OpenShift peribolos config is actually machine-produced, right?) peribolos configs but porting should be fairly mechanical and I hope there aren't that many.

cmd/peribolos/main.go
```diff
@@ -890,10 +912,27 @@ func configureOrg(opt options, client github.Client, orgName string, orgConfig o
...
+	// forkNames maps config repo name -> actual GitHub repo name (for renamed forks)
+	var forkNames map[string]string
+	if !opt.fixForks {
+		logrus.Info("Skipping repository forks configuration")
+		forkNames = make(map[string]string)
+	} else {
+		var err error
+		forkNames, err = configureForks(client, orgName, orgConfig)
+		if err != nil {
+			logrus.WithError(err).Error("errors configuring some forks, continuing with partial results")
```
petr-muller: not sure if simply swallowing the error is the right thing to do; it is fine to continue but eventually the failure should propagate and I'd expect peribolos to exit with non-zero in that case

cmd/peribolos/main.go
```diff
@@ -890,10 +912,27 @@ func configureOrg(opt options, client github.Client, orgName string, orgConfig o
...
+		}
+		if forkNames == nil {
+			forkNames = make(map[string]string)
+		}
```
petr-muller: this (also L920) is not necessary, both consumers only read from the map, which is safe to do even for `nil` maps

cmd/peribolos/main.go
```diff
@@ -1138,6 +1209,216 @@ func configureRepos(opt options, client repoClient, orgName string, orgConfig or
...
+	for repoName, repoCfg := range orgConfig.Repos {
+		if repoCfg.ForkFrom == nil || *repoCfg.ForkFrom == "" {
+			continue
+		}
+		upstream := strings.ToLower(*repoCfg.ForkFrom)
+		if existing, ok := upstreamToConfig[upstream]; ok {
+			return nil, fmt.Errorf("multiple config entries (%s, %s) fork from the same upstream %s — only one fork per upstream is allowed per org", existing, repoName, *repoCfg.ForkFrom)
+		}
+		upstreamToConfig[upstream] = repoName
+	}
```
petr-muller: can we do the L1253 GetRepos only after this block, and gate it by an early return on `len(upstreamToConfig) == 0`? 

if we fail validation, or do not actually have fork configs, we would save a `GetRepos` API call, potentially paginated. we infer `--fix-forks` from `--fix-repos` so even runs that do not explicitly opt-in for fork management and do not contain any fork-related items in their peribolos config would neeed to pay this API tax

cmd/peribolos/main.go
```diff
@@ -1138,6 +1209,216 @@ func configureRepos(opt options, client repoClient, orgName string, orgConfig or
...
+		var configNameResult *parentResult
+		found := false
+		for _, candidate := range candidates {
+			repo, exists := byName[candidate]
+			if !exists || !repo.Fork {
+				continue
+			}
+			fullRepo, err := client.GetRepo(orgName, repo.Name)
+			if candidate == strings.ToLower(repoName) {
+				configNameResult = &parentResult{fullName: fullRepo.Parent.FullName, err: err}
```
petr-muller: `fullRepo.Parent.FullName` will likely be empty if on error (`err != nil`), should we fallback to some other name in that case (I'm not sure how this is exactly consumed later)?

cmd/peribolos/main.go
```diff
@@ -1138,6 +1209,216 @@ func configureRepos(opt options, client repoClient, orgName string, orgConfig or
...
+				found = true
+				break
+			}
+		}
+		if found {
+			continue
+		}
+
+		// Check if a repo with the config name already exists but is not a fork of our upstream
+		existingRepo, exists := byName[strings.ToLower(repoName)]
```
petr-muller: the amount of `ToLower` and `EqualFold` usage I see makes me a bit uncomfortable - it is unclear what actually does need this treatment and what doesn't, and I also feel like it makes it very easy for anyone touching this code to make a mistake

not sure how exactly to solve this and feel free to ignore, just an observation

cmd/peribolos/main.go
```diff
@@ -1138,6 +1209,216 @@ func configureRepos(opt options, client repoClient, orgName string, orgConfig or
...
+					}
+				}
+			} else {
+				// It's not a fork at all
+				err := fmt.Errorf("repo %s already exists but is not a fork", repoName)
+				repoLogger.WithError(err).Error("cannot create fork - repo exists")
+				allErrors = append(allErrors, err)
+			}
+			continue
+		}
```
petr-muller: consider, not blocking: this section is hard to follow with this many nested conditionals... maybe consider extracting it to a method? not sure.

petr-muller:
@hoxhaeris thanks for the response on the fork configuration shape! what you say makes sense - no good reason to change the config shape that significantly, let's go with yours. I have done one more walkthrough, have a bunch of mostly code-oriented feedback, nothing significant. Main two points would be coupling the fork-related config fields into a struct and gating a `GetRepos` API call only after we make sure there are actually forks to be managed.

hoxhaeris:
> @hoxhaeris thanks for the response on the fork configuration shape! what you say makes sense - no good reason to change the config shape that significantly, let's go with yours. I have done one more walkthrough, have a bunch of mostly code-oriented feedback, nothing significant. Main two points would be coupling the fork-related config fields into a struct and gating a `GetRepos` API call only after we make sure there are actually forks to be managed.

Thank you, @petr-muller! I'll start addressing those.

pkg/config/org/org.go
hoxhaeris: The `RepoMetadata` extraction is backwards compatible at the serialization layer: YAML/JSON keys are unchanged because Go struct embedding flattens the fields. The only consumers that would break are Go programs that construct `org.Repo` struct literals directly, and they would get a clear compile error (not a silent behavior change, not a runtime error).

Within the prow repo, `cmd/peribolos` is the only importer of `pkg/config/org`. External consumers are unknown, but as you noted, porting is mechanical: change `org.Repo{Description: ...}` to `org.Repo{RepoMetadata: org.RepoMetadata{Description: ...}}`.

I kept this as a separate commit so it can be reverted independently if needed. That said, do you consider the `org` package's Go API stable for external importers, or is it effectively internal to peribolos?

pkg/config/org/org.go
hoxhaeris: Done in 2479e7c74. Introduced `ForkConfig` struct with `From string` and `DefaultBranchOnly bool`. The config surface is now:

```yaml
repos:
  my-fork:
    fork:
      from: owner/repo
      default_branch_only: true
```

`Fork *ForkConfig` on `Repo` means `nil` = "not a fork", which simplifies all the nil/empty-string checks throughout.

cmd/peribolos/main.go
hoxhaeris: Done in a513e7421. Fork errors are now deferred: `configureOrg` saves the error, continues with repos/collaborators/teams, and returns the fork error at the end if nothing else failed first. This way fork failures don't block other subsystems, but peribolos still exits non-zero.

cmd/peribolos/main.go
hoxhaeris: Removed in a513e7421. Both `make(map[string]string)` guards are gone. Downstream consumers (`configureRepos`, `configureCollaborators`) only read from the map, so a nil map is safe.

cmd/peribolos/main.go
hoxhaeris: Done in 9c74482e9. Validation (format checks + duplicate upstream detection) now runs before `GetRepos`. If no fork configs exist (`len(upstreamToConfig) == 0`), the function returns early with no API calls. This also means format validation errors surface before any network activity.

cmd/peribolos/main.go
hoxhaeris: Fixed in af38ab180. `fullName` is now only captured when the `GetRepo` call succeeds:

```go
pr := &parentResult{err: err}
if err == nil {
    pr.fullName = fullRepo.Parent.FullName
}
configNameResult = pr
```

The downstream code already checked `configNameResult.err != nil` before using `fullName`, so this was safe before, but now the intent is clearer.

cmd/peribolos/main.go
hoxhaeris: Addressed in af38ab180. Repo names and upstream identifiers are now normalized to lowercase once at the top of each loop iteration (`repoNameLower`, `upstreamRepoNameLower`, `expectedUpstream`). The normalized values are used consistently for all map lookups and comparisons, replacing the scattered `strings.ToLower()` and `strings.EqualFold()` calls. The original-case `repoName` is kept for log/error messages and the `forkNames` map key.

cmd/peribolos/main.go
hoxhaeris: Extracted in af38ab180. The nested collision-check block is now a separate `checkExistingRepoConflict` function that returns an error describing the conflict (wrong upstream, not a fork) or nil.

Issue #585, opened 2026-01-05
Remove unneeded prow plugins/cmds


upodroid:
Prow has many plugins/cmds that we don't need, or GitHub implemented them as features since we wrote them:

Examples:
 - golint, it uses `golang.org/x/lint` which is deprecated, repositories should be running golangci-lint as a verify/unit test job
 - dco. GitHub offers native DCO checking as a repository setting. https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository
 - jenkins: We don't use this anymore
 - buildifier: We don't use bazel anymore. Also, it's a linter that should run in a verify/lint job
 
Most of these are pulling large/unmaintained dependencies that we should drop from prow.

rajibmitra:
I’d like to take a stab at it @upodroid

upodroid:
Try removing the dco one

petr-muller:
I agree we should shed this stuff eventually but I'd like to do it in a somewhat controlled way: 

1. announce their deprecation and eventual removal (not having a good release process makes this a bit harder 😅) with a specific date in the future
2. make a change in the code first that logs, as a warning, that a Prow instance uses an eventually deprecated plugin
3. only at the date actually remove the functionality, and _preferably_ in a way that does not _break_ a hook instance that still consumes config for the plugin (=fail in checkconfig but do not fail on hook startup)

valen-mascarenhas14:
@upodroid @petr-muller 
> Try removing the dco one

My understanding is that GitHub’s native DCO enforcement only applies to commits created via the [GitHub web UI](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository#:~:text=Compulsory%20commit%20signoffs%20only%20apply%20to%20commits%20made%20via%20the%20web%20interface).
Does it also enforce sign-offs for commits created via CLI/IDEs or bot-generated commits?
If not, how would we ensure equivalent DCO coverage without the Prow DCO plugin?

upodroid:
It also applies to commits made by git client. It says the following:

```
For commits made via the Git command line interface, the commit author must sign off on the commit using the --signoff option. For more information, see the [Git documentation](https://git-scm.com/docs/git-commit).
```

valen-mascarenhas14:
> It also applies to commits made by git client. It says the following:
> 
> ```
[...]

PR #586, opened 2026-01-07
Remove DCO from prow 

rajibmitra:
See the main issue here https://github.com/kubernetes-sigs/prow/issues/585#issue-3781597724 
This is a child task to remove the DCO from the Prow.


petr-muller:
/hold

I agree we should shed this but I'd like to deprecate -> announce -> wait -> remove, and take a bit of care to not recklessly break Prow instance that still has this enabled. Ideally we decide how exactly to proceed for all removals in the parent issue ( https://github.com/kubernetes-sigs/prow/issues/585#issuecomment-3718929014 )

PR #590, opened 2026-01-12
fix(deck): update MDL CDN to use cdnjs

wuhuizuo:
Fix #588 
Switch from code.getmdl.io to cdnjs for Material Design Lite assets

The whole `code.getmdl.io` site was abnormal with response 403 status code.

PR #593, opened 2026-01-16
fix(spyglass): Regenerate bindata.go with cdnjs URL

cardil:
Follow up #591 and #590

## Problem

After commits 782492312ad0696115819ab023e7bda6606d3a17 and cc03e3b4fe3f1a8874c304b38a862233e6041521 updated the Material Design Lite CDN from `code.getmdl.io` to `cdnjs.cloudflare.com`, the old URL was still appearing in spyglass iframes on the live prow.k8s.io site.

Example: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-prow-test-verify/2010981049969938432

Browser console showed 403 errors:
```
Failed to load resource: the server responded with a status of 403 () @ https://code.getmdl.io/...
```

## Root Cause

The [`pkg/spyglass/lenses/common/bindata.go`](pkg/spyglass/lenses/common/bindata.go:81) file contains a gzip-compressed, embedded copy of `spyglass-lens.html` that is compiled into the deck binary at build time. The previous commits only updated the source HTML files but did not regenerate `bindata.go`, so the deployed binary was still serving the old embedded content with the `code.getmdl.io` URL.

## Solution

- Regenerated `bindata.go` using go-bindata to embed the updated HTML with the cdnjs URL
- Added `bindata_test.go` to verify the embedded content contains the correct URL and prevent future regressions
- Test confirms the embedded HTML now contains `https://cdnjs.cloudflare.com/ajax/libs/material-design-lite/1.3.0/material.indigo-pink.min.css`

Once deployed, spyglass iframes will load Material Design Lite from the working cdnjs CDN instead of the defunct code.getmdl.io.

## Testing

```bash
$ go test -v ./pkg/spyglass/lenses/common -run TestBindataContainsCdnjsURL
=== RUN   TestBindataContainsCdnjsURL
--- PASS: TestBindataContainsCdnjsURL (0.00s)
PASS
ok  	sigs.k8s.io/prow/pkg/spyglass/lenses/common	0.017s
```


smg247:
It looks like the `make generate` target was commented out some time ago, I am not sure if that would have caught this in the first place though. What command did you use to generate the file? I would like to see the unit test verify that the generated file matches what is expected in general, rather than just checking that a couple of strings do or don't exist.

cardil:
> @smg247 What command did you use to generate the file?

I manually regenerated it like this:
```
$ GOBIN=/tmp go install github.com/go-bindata/go-bindata/v3/go-bindata@latest
$ cd pkg/spyglass/lenses/common/ && /tmp/go-bindata -modtime 1 -pkg=common static/ && gofmt -s -w ./
```

From here: https://github.com/kubernetes-sigs/prow/blob/898377d8b2b69e233c141cb8595c3669e3a09c16/hack/make-rules/update/codegen.sh#L158-L164

> @smg247 I would like to see the unit test verify that the generated file matches what is expected in general, rather than just checking that a couple of strings do or don't exist.

I can remove the unit test if you like. In general, there probably should be some kind of build test that would verify the generated code is properly updated.

smg247:
> I can remove the unit test if you like. In general, there probably should be some kind of build test that would verify the generated code is properly updated.

Yes, that is what I would like to see here as well.

cardil:
> @smg247 :Yes, that is what I would like to see here as well.

In Knative we used to have simple verify scripts that ran as part of the build test to verify there are no pending changes after the codegen run: https://github.com/knative/client/blob/e46abb4d1df866343859796fa545abd088c5177e/hack/verify-codegen.sh#L24-L35

cardil:
@smg247 There looks to be a similar verify script here already: https://github.com/kubernetes-sigs/prow/blob/898377d8b2b69e233c141cb8595c3669e3a09c16/Makefile#L90-L92

It looks like it is not being executed on the Prow tests...

smg247:
> @smg247 There looks to be a similar verify script here already:
> 
> https://github.com/kubernetes-sigs/prow/blob/898377d8b2b69e233c141cb8595c3669e3a09c16/Makefile#L90-L92
> 
> It looks like it is not being executed on the Prow tests...

Ah yes, I thought I remembered there being a make target... that does appear to make the change to bindata when running locally on `main`. So we likely need to get that into a presubmit. For now, remove your unit test, and make sure you generate the bindata using that command. We can get this merged and then add/modify the presubmit after.

cardil:
Unit test is deleted already.

I ran `make verify-codegen` successfully:

```
diffing /home/ksuszyns/git/kubernetes-sigs-prow against freshly generated codegen
/home/ksuszyns/git/kubernetes-sigs-prow up to date.
```

cardil:
We also need to approve the autobump so the changes are in effect: https://github.com/kubernetes/k8s.io/pull/8955

PR #594, opened 2026-01-16
Verify that the spyglass bindata is generated as well

smg247:
https://github.com/kubernetes-sigs/prow/pull/593 fixed the `spyglass` bindata, but the fact that it was able to get outdated in the first place points to a gap in the presubmits. We need to add the `spyglass` dir to those that we check with `make verify`.


smg247:
Still looking into a solution to the failing `verify-lint` presubmit here, but posting this here for transparency (and so I don't forget). Claude thinks it has discovered the issue:

  The issue is caused by a Go 1.25.x regression in doc comment handling that affects how go-bindata generates code:

  The Problem Chain:

  1. go-bindata v3.1.3 (built with Go 1.25.5) generates code with a leading space before the comment:
   //Package common generated by go-bindata.// sources:
  2. First gofmt run removes the leading space but doesn't fix the comment spacing:
  //Package common generated by go-bindata.// sources:  // Missing space after //
  3. Second gofmt run would add the required space:
  // Package common generated by go-bindata.// sources:  // Correct!

  The gen-spyglass-bindata() function in hack/make-rules/update/codegen.sh:158-164 only runs gofmt once, leaving the file in the incorrect state.

  Why CI Fails But Local Passes:

  - Locally: make update-codegen generates the incorrect format (//Package)
  - In CI: The generated code somehow ends up with the correct format (// Package)
  - When CI runs verify-codegen, it compares the committed (wrong) version against freshly generated (correct) version and fails

  Solutions:

  Option 1 - Run gofmt twice (simplest fix):
  gen-spyglass-bindata() {
    cd pkg/spyglass/lenses/common/
    echo "Generating spyglass bindata..." >&2
    $go_bindata -modtime 1 -pkg=common static/
    gofmt -s -w ./
    gofmt -s -w ./  # Run twice to fix go-bindata + Go 1.25 issue
    cd - >/dev/null
  }

  Option 2 - Upgrade/fork go-bindata to fix the Go 1.25 compatibility issue

  Option 3 - Build go-bindata with Go 1.24 instead of Go 1.25

krzyzacy:
shall we open an upstream issue as well?

cardil:
We also need to approve the autobump: https://github.com/kubernetes/k8s.io/pull/8955

PR #598, opened 2026-01-23
Jenkins: Check building before accepting result

### LINKED ISSUE #142: Wrong prow status during Jenkins parallel pipeline test
We are having an issue where prow Jenkins operator is not supporting Jenkins parallel pipelines. Since migrating from ghprb, we have noticed instances where tests triggered from prow reported as passed on GitHub, while some of the parallel stages had failed/aborted on Jenkins.

The issue reproduced in the following test PR where it has parallel Jenkins job was running: [Test PR](https://github.com/metal3-io/cluster-api-provider-metal3/pull/1641) . Here we can see `metal3-centos-e2e-feature-test-main — Jenkins job succeeded.` status for parallel run even though only one test was passing and two other tests were aborted/failed. [Jenkins Reproduction 1](https://jenkins.nordix.org/blue/organizations/jenkins/metal3-centos-e2e-feature-test-main/detail/metal3-centos-e2e-feature-test-main/27/pipeline) , [Jenkins Reproduction 2](https://jenkins.nordix.org/blue/organizations/jenkins/metal3-centos-e2e-feature-test-main/detail/metal3-centos-e2e-feature-test-main/22/pipeline)

Also, adding the link to the parallel test [pipeline](https://github.com/metal3-io/project-infra/blob/main/jenkins/jobs/parallel_e2e_features_test.pipeline) code which was working properly before the prow migration from ghprb.

We expect to have correct prow response in case of any failure in the parallel test pipeline. In the above test cases prow response status should have` Failure `status instead of `Success`.

lentzi90:
In some cases, jenkins may report a result for a job while it is still building. This can happen in matrix pipelines when one branch finishes before the rest. To avoid false results, we need to check if the job is still building before we accept the result. Otherwise we may mark a job as successful after only one branch has succeeded.

This commit uncludes the building status in the API query and checks it as part of IsRunning. It also adds unit tests for IsRunning, IsSuccess, IsFailure and IsAborted.

Fixes #142

### DIFF
--- a/pkg/jenkins/jenkins.go
+++ b/pkg/jenkins/jenkins.go
@@ -96,6 +96,7 @@ type Build struct {
 	} `json:"task"`
 	Number   int     `json:"number"`
 	Result   *string `json:"result"`
+	Building bool    `json:"building"`
 	enqueued bool
 }
 
@@ -122,8 +123,11 @@ type JobInfo struct {
 }
 
 // IsRunning means the job started but has not finished.
+// We check Building first to handle Jenkins matrix builds where
+// result can be set to SUCCESS while the build is still running
+// (when one matrix cell completes before others).
 func (jb *Build) IsRunning() bool {
-	return jb.Result == nil && !jb.enqueued
+	return jb.Building || (jb.Result == nil && !jb.enqueued)
 }
 
 // IsSuccess means the job passed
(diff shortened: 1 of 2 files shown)


lentzi90:
Looks like we have a non-deterministic unit test. I'll push a separate PR for that.
/retest

lentzi90:
Flaky test fixed in #599

Issue #527, opened 2025-10-13
Docs: Add subsection for running Prow outside GKE (e.g., local/Kind)


tsj-30:
While setting up Prow for local exploration and development, we noticed that the current Deploying Prow
 guide is primarily written with Google Kubernetes Engine (GKE) in mind. Although the document states that “Prow runs in any Kubernetes cluster,” most configuration examples and environment assumptions are Google-specific.

**Problem**

- This GKE-centric approach can make it challenging for new contributors or developers who want to:
- Experiment with Prow locally (for learning, testing, or extending plugins/UI).
- Run Prow on lightweight environments such as Kind.
- Validate PRs, configs, or plugin changes without relying on cloud infrastructure.

Currently, there is no dedicated subsection that explains how to set up Prow in non-GKE environments. This limits readability and creates a barrier for those exploring Prow in alternative setups.

**Proposed Improvement**

- Add a subsection (for example, under “Deploying Prow”) that describes:
- Minimal local setup instructions using Kind:
- Create the cluster.
- Apply starter manifests.
- Configure a simple GitHub App + HMAC secret.
- Call out common pitfalls during setup.
- Explicitly clarify that this setup is intended for exploration/learning, not production-like use cases.

Broaden the doc’s scope to make it less GKE-centric by acknowledging other possible environments where contributors may want to test.

Why This Matters

- Lowers the barrier for contributors who want to test PRs or changes locally.
- Makes the docs more accessible for developers working outside GKE.

BenTheElder:
> Configure a simple GitHub App + HMAC secret.

This isn't going to work well with a local cluster ... the HMAC secret is for incoming webhooks, so the cluster has to be exposed to the public internet, which we do not generally recommend with `kind` for security reasons (also doing so portably is a complex topic because it involves exposing your local machine to the public internet as well).

These are the parts that are generally GKE centric, because that is where Kubernetes runs prow.

For something like running a prowjob locally on `kind`, there are already tools / scripts / docs in the repo for that.

https://github.com/kubernetes-sigs/prow/issues/283 already covers moving the starter config, which was left in test-infra when prow moved. IMHO it should be moved here and updated. PRs welcome.


Also, FYI if you're running prow outside of the Kubernetes project ... #113 at some point we will need to cut off access to pulling from staging.

Amulyam24:
> > Configure a simple GitHub App + HMAC secret.
> 
[...]

PR #611, opened 2026-02-04
Doc: Add a local hook development guide

linolayani:
This PR adds a local development guide for Hook and its plugins. It uses a webhook delivery service, smee, to allow Hook to receive GitHub webhooks while remaining privately exposed.
It contains a step by step guide, as well as boilerplate configuration files, to get started easily. Local changes to these files are ignored by git.

**Purpose** 
Makes it easier for newcomers to run Hook and its plugins locally, ultimately lowering the barrier to local development

**Testing**
Followed the guide

**Assumptions**
- the `config/local` folder is a good place for boilerplate and local configuration files
- the guide should live in its own file (separate from other getting-started content)


ivankatliarchuk:
relates #527

The approach make sense. What worth to review - execute withing docker-compose smee + prow. Worth to scope it, aka hook tutorial, UI development and etc. 

Location `config/local/hook` is confising - someone make decide is a pre-commit hook. 

Security concerns:
  - config/local/hook/hmac-secret contains a hardcoded secret value (abcde12345). Even though it's a placeholder, committing credential-like content to the repo is a bad pattern. This should instead be a plain comment file (like the .pem file) instructing the user to place their secret here - not an example value.
  - The README says "sensible credentials" — this is a typo, should be "sensitive credentials".

Typos and grammar in getting-started-hook.md:
  - "Form the root" → "From the root" (Step 1 under "Run hook and smee")
  - "smee forwards webhooks" (extra space before "forwards")
  - "We're now going to" — sentence is incomplete; missing the rest (e.g. "run the two processes.")
  - "This plugin listen for" → "This plugin listens for"
  - "3. The local smee-client receive" → "receives"
  - "Hook receives the webhook, parse" → "parses"
  - "eg." → "e.g." (twice)
  - "adding a comment the issue" → "adding a comment to the issue"

CLI flags formatting issue:
  The go run command block has misaligned flags - the last two flags (--github-app-private-key-path and --github-app-id) are not escaped properly with `\` continuation. As written, the command would fail. The `\ after the penultimate line is missing:

  --github-app-private-key-path=config/local/hook/prow-sandbox.private-key.pem \
  --github-app-id=$GITHUB_APP_ID

Some minor structural
- The App registration steps suggest http://irrelevant as a Homepage URL, which may not pass GitHub's validation. Better to suggest http://localhost
-The guide doesn't mention what port Hook defaults to until Step 2 of "Run hook and smee". Consider mentioning it earlier (e.g., in the smee setup step).
- The .gitignore comment says "Make sure to not commit sensitive credentials" - this is a double negative; consider "Do not commit sensitive credentials."

I've not yet tried this guide, but quite close to how I would do it.

Issue #468, opened 2025-05-22
Cherrypick behaviour is different depending on PR author and whether you use a label or a comment


craigbox:
We have had intermittent problems in Istio where setting a cherry-pick label on an issue would not cause the cherrypicker to run, but the same person leaving a comment would.

PR that shows failure: https://github.com/istio/istio.io/pull/16501 (first comment ignored, setting label ignored, second comment worked)
PR that shows success: https://github.com/istio/istio.io/pull/16505

We have today narrowed it down to this set of circumstances:

* Only org members are allowed to trigger Prow actions 
* Prow treats a cherrypick label as a comment _from the PR author_, not from the person who set the label
* The people who set the cherrypick label are almost always org members

Thus, one of the following things might be an improvement:

* The cherrypick label is treated as a comment _from the person who set the label_, not the PR author
* (if the above is somehow insecure) A comment is left saying that the user doesn't have access to interact with Prow and that no action will be taken, so they can know to make a comment (or add/remove the label?)

craigbox:
Thanks @smg247; we did see that but we probably want to keep instructions to the bot restricted to org members. (The bar to becoming one is low.)

petr-muller:
> It does seem like it would be somewhat trivial to leave a comment in the instance that a label was added that will be ignored due to the PR author not being an org member though.

This direction sounds easy enough for a good first issue, labeling as such

petr-muller:
I'm not that familiar with the label-trigger workflow of this plugin, but it should also be possible to make it work for the case where the label is added after the PR is merged, where the event author is available.

valen-mascarenhas14:
@petr-muller @smg247 
Have raised a PR for this. Here's a snapshot of the comment generated when the cherrypick label is added on a PR authored by a non-org member 

<img width="1158" height="231" alt="Image" src="https://github.com/user-attachments/assets/4eee191c-65d0-4cad-aeda-61d19a465705" />

valen-mascarenhas14:
@craigbox As the PR for the above functionality is merged, can this issue be closed ?

PetrMc:
opened #943 for this, uses the issue events api to find who added the label so no state needs to be stored

PR #614, opened 2026-02-07
external-plugins/cherrypicker: comment when label is ignored for non-org member

valen-mascarenhas14:
Fix for https://github.com/kubernetes-sigs/prow/issues/468

## Summary

This PR introduces an improvement to the cherry-pick handling logic:

**It Adds an explicit feedback when a cherry-pick label is ignored due to org membership restrictions.**

Previously, when a cherry-pick label was added to a PR authored by someone who is not a member of the org, the request was silently ignored. This could be confusing as there was no visible indication that the label had been rejected.

With this change:
- If a cherry-pick label is detected
- And the PR author is not an org member
- The system posts a GitHub comment explaining that the label-based cherry-pick request is being ignored due to membership restrictions.

---


valen-mascarenhas14:
@matthyx @smg247 Any review comments ?

matthyx:
isn't it a bit spam-y?

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -424,6 +424,16 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
 				return requester == member.Login
 			})
 			if !isMember {
+				s.ghc.CreateComment(
```
petr-muller: `CreateComment` can fail and should be logged like in other places

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -424,6 +424,16 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
...
+    				org,
+    				repo,
+    				num,
+    				fmt.Sprintf(
+        				"The cherrypick label was added, but it will be ignored because the PR author (@%s) is not a member of the %s org.",
```
petr-muller: There's a similar message on L49 that is more actionable, suggesting the requestor to make sure their org membership is public. Once we make sure the change really responds only to the event where the new label is added, the comment can also suggest to whoever added the label to use a comment instead.

petr-muller:
I don't think this is *too* spammy but it can definitely be improved: if I read the code right, this would post a comment for 1) each disallowed requestor, including ones that *commented* earlier on 2) each relevant post-merge (because of L338 gate) - actual merge, cherry-pick command, label added. Which is excessive but not the end of the world.

My concern is that the comment is not applied when someone adds the label before merge - only when the PR is later processed post-merge. This is handled correctly for comments because it goes through the issue commented event path.

I believe to fix this right we'd need a similar flow for when the relevant label is added like we have for comments: when the label is added, first check if the requestor (for labels it is author) is eligible to do a cherry pick and respond accordingly right away; only proceed to an actual cherry pick afterwards if the PR is already merged (this is what `handleIssueComment` does. 

This would mean splitting the current `handlePullRequest` path to "label added" path that would partially mimic the comment path describe above, and the "PR merged" path to handle all instructions (comment & labels) put on the PR before merge.

We'd also need adequate test coverage.

valen-mascarenhas14:
@petr-muller 
Thanks, that makes sense. Just to confirm my understanding of the intended end functionality:

* When a relevant cherry-pick label is added (even before merge), we should immediately validate the requestor (PR author).
* If the requestor is not eligible, the plugin should post a comment right away indicating that the cherry-pick request is not allowed.
* If the requestor is eligible, we record the intent but do not perform the cherry-pick until the PR is merged.
* Once the PR is merged, the plugin should then proceed to raise the cherry-pick PR .
Does this align with what you had in mind for the final behavior?

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: Done

valen-mascarenhas14:
@petr-muller  Please  take a look. I've updated the description with info on the changes I made

valen-mascarenhas14:
@matthyx @smg247 @petr-muller Any review comments/ suggestions on this ?

petr-muller:
> Does this align with what you had in mind for the final behavior?

Yeah I think your comment summarizes it pretty well. I'm not sure if I'll get to the PR this week but I'll try.

valen-mascarenhas14:
@petr-muller Did you get a chance to review this PR ?

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -419,51 +482,35 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
...
 				continue
 			}
-			handledBranches[targetBranch] = true
 			branchLog := log.WithFields(logrus.Fields{
 				"requester":     requester,
 				"target_branch": targetBranch,
 			})
 			branchLog.Debug("Cherrypick request.")
-			err := s.handle(branchLog, requester, ic, org, repo, targetBranch, baseBranch, targetBranchToChainBranches[targetBranch], title, body, num)
+			err := s.handle(branchLog, requester, nil, org, repo, targetBranch, baseBranch, nil, title, body, num)
```
petr-muller: Changes to pass `nil` to the two parameters is suspicious. This is the merge event handler, and the logic iterates over past comments (and labels but that does not seem relevant). Previous logic passed the individual acted-on comments to handle (I think this is needed for "this is a response to this command" part of response comments - losing this woudl not be too serious but there's no reason to). Losing correct passing of the `chainBranches` param would mean the pre-merge comments would not be processed correctly though - and inconsistently with post-merge comments (that would be processed through the "commented" path which processess the `chainBranches` correctly.

cmd/external-plugins/cherrypicker/server_test.go
```diff
@@ -604,12 +604,6 @@ func testCherryPickPR(clients localgit.Clients, t *testing.T) {
 	expectedFn := func(branch string) string {
 		expectedTitle := fmt.Sprintf("[%s] This is a fix for Y", branch)
 		expectedBody := fmt.Sprintf("This is an automated cherry-pick of #%d", prNumber)
-		if branch == "release-1.3" {
-			expectedBody = fmt.Sprintf("%s\n\n/cherrypick release-1.2", expectedBody)
-		}
-		if branch == "release-1.12" {
-			expectedBody = fmt.Sprintf("%s\n\n/cherrypick release-1.11 release-1.10 release-1.9", expectedBody)
-		}
```
petr-muller, line 612: I think this removal shows the problem with losing `chainBranches I mention above.

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -329,12 +329,90 @@ func parseComment(comment github.IssueComment) cherrypickCommands {
...
 func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullRequestEvent) (logrus.FieldLogger, error) {
-	// Only consider newly merged PRs
-	if pre.Action != github.PullRequestActionClosed && pre.Action != github.PullRequestActionLabeled && pre.Action != github.PullRequestActionOpened {
+	switch pre.Action {
+	case github.PullRequestActionLabeled:
+		return s.handlePullRequestLabelAdded(log, pre)
+	case github.PullRequestActionClosed:
+		return s.handlePullRequestMerged(log, pre)
+	default:
+		return log, nil
```
petr-muller, line 338: Previous code also handled `Opened` - where did that one go?

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -372,43 +448,30 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
...
-			}
-			requesterToComments[comment.User.Login][targetBranch] = &comment
-
-			if len(chainedBranches) > 0 {
-				targetBranchToChainBranches[targetBranch] = chainedBranches
+		for targetBranch := range commands {
+			if requesterToBranches[comment.User.Login] == nil {
+				requesterToBranches[comment.User.Login] = sets.New[string]()
 			}
+			requesterToBranches[comment.User.Login].Insert(targetBranch)
```
petr-muller: I do not understand these changes - how are they related to the PR goal? I would expect handling an added label code from this method, but this seems to go beyond that. Is this somehow a refactoring or do these changes have a functional change goal?

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -419,51 +482,35 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
 		if err != nil {
 			return log, err
 		}
-		for requester := range requesterToComments {
+		for requester := range requesterToBranches {
 			isMember := slices.ContainsFunc(members, func(member github.TeamMember) bool {
```
petr-muller, line 480: Hmm in your new code you use `IsMember` for this which seems more efficient, we wont be checking more than a couple names. Could be an opportunity for a followup - but pls not in this PR, it's already a bit invoved to review.

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -419,51 +482,35 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
...
 	}
 
-	// Handle multiple comments serially. Make sure to filter out
-	// comments targeting the same branch.
-	handledBranches := make(map[string]bool)
 	var errs []error
-	for requester, branches := range requesterToComments {
-		for targetBranch, ic := range branches {
-			if handledBranches[targetBranch] {
-				// Branch already handled. Skip.
```
petr-muller, line 439: This comment feels useful, why remove it?

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -419,51 +482,35 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
...
 				continue
 			}
+			handledBranches.Insert(targetBranch)
+
 			if targetBranch == baseBranch {
-				resp := fmt.Sprintf("base branch (%s) needs to differ from target branch (%s)", baseBranch, targetBranch)
-				log.Info(resp)
-				if err := s.createComment(log, org, repo, num, ic, resp); err != nil {
-					log.WithError(err).WithField("response", resp).Error("Failed to create comment.")
-				}
```
petr-muller, line 447: So previously we warned about this and now we don't and just silently skip? Does not seem right.

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: Hey, Thanks for pointing out.
While reviewing the merge handler, the call to `handle()` made it look like the essential inputs were the requester & the target branch used to create the cherrypick PR, while the comment & `chainBranches` seemed secondary. Based on that, I simplified the reconstruction logic to track only requester → target branches & passed `nil` for those parameters.

However, I see that `chainBranches` is required to preserve the sequential cherry-pick behavior. 

I’ll revert this part to keep the current behavior intact and move the refactoring into a follow-up PR.

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: @petr-muller 
As per my understanding, In the previous implementation `PullRequestActionOpened` passed the initial filter, but was ignored later as the handler immediately returned if the PR wasn’t merged 
https://github.com/valen-mascarenhas14/prow/blob/8c5d6f1cbaa91cc3fd06f35abc2078e5a352592d/cmd/external-plugins/cherrypicker/server.go#L338

Since an opened PR is never merged, that path would never result in any cherry-pick handling.

So I added it in the default case as it would keep the behavior the same.

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: @petr-muller `IsMember` was already referenced in the original code https://github.com/kubernetes-sigs/prow/blob/af120b0c1a42eaae8f57a5376079801d2470fd77/cmd/external-plugins/cherrypicker/server.go#L423
I've just used the same code

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: Noted. Have added it back. Thanks

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: Have added it back

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: This was a misunderstanding on my end. The earlier changes were made based on my interpretation of the behavior for chained cherry-pick branches, which resulted in passing `nil` for those parameters in handle() method.

valen-mascarenhas14:
@petr-muller I've addressed the review comments. Did you get a chance to check ?

cmd/external-plugins/cherrypicker/server.go
petr-muller: Interesting. I looked into the commit history and it was apparently added long time ago to allow authors to have cherry-pick commands right in the initial PR descriptions, as if the author commented. It is true that it actually did not trigger any behavior by itself, but the commands would still probably apply after merge. 

Interesting. If we wanted to fix this we'd probably need to wire the comment handler behavior to the PR opened event (no need to do it here in this PR).

cmd/external-plugins/cherrypicker/server.go
```diff
@@ -383,35 +454,21 @@ func (s *Server) handlePullRequest(log logrus.FieldLogger, pre github.PullReques
...
-	foundCherryPickLabels := false
 	for _, label := range labels {
 		if strings.HasPrefix(label.Name, s.labelPrefix) {
 			requesterToComments[pr.User.Login][label.Name[len(s.labelPrefix):]] = nil // leave this nil which indicates a label-initiated cherry-pick
-			foundCherryPickLabels = true
 		}
 	}
-
-	if !foundCherryPickComments && !foundCherryPickLabels {
+	if len(requesterToComments) == 0 {
```
petr-muller, line 469: This is never true because we always add an item on L462 

This would make us continue down on all merges, consuming API tokens even when there's nothing to do. I think it should be enough to move the map creation to just before L466 that adds to it, and we should be good to go.

cmd/external-plugins/cherrypicker/server.go
petr-muller: looks good now, thanks!

cmd/external-plugins/cherrypicker/server.go
petr-muller: fair enough. could be worth a followup

petr-muller:
noticed one last thing then I think this is good to go 

thanks for the work and patience!

cmd/external-plugins/cherrypicker/server.go
valen-mascarenhas14: Thanks for pointing out @petr-muller it was an oversight on my end. Have made the change

valen-mascarenhas14:
@petr-muller Did you get a chance to look at this ?

valen-mascarenhas14:
@petr-muller whenever free, could you take a look at this too please ?

stmcginnis:
Looks good, and I think all comments have been addressed.

/lgtm

valen-mascarenhas14:
Thanks @stmcginnis .
@petr-muller Do you have any inputs / review comments on this ?

valen-mascarenhas14:
@petr-muller Could you take a look at this ?

petr-muller:
/approve

Thank you!

PR #615, opened 2026-02-07
deck: drop stale non-blocking presubmit contexts from PR status

### LINKED ISSUE #36399: /skip does not fully clear stale non-blocking failures in prow bot comment (especially after presubmit job rename)
**What happened**:

On some PRs, the prow bot comment lists multiple non-blocking job failures that appear stale or unrelated to the current commit.

Running `/skip` (and sometimes `/clear`) may ignore certain CI statuses, but stale non-blocking failures can remain visible in the bot comment.  
In addition, `/skip` may not behave as expected when a presubmit job has been renamed, leaving outdated results in the comment.

This creates confusing CI signal for reviewers even when all required GitHub checks are passing.

One of the following:

- `/skip` should fully clear or reset stale non-blocking failures shown in prow bot comments, **or**
- There should be a documented and reliable way to reset the bot comment state, including cases where presubmit job names change.

1. Open a PR where the prow bot comment lists non-blocking failed jobs.
2. Confirm required GitHub checks are passing and merge is not blocked.
3. Issue `/skip` (and optionally `/clear`) on the PR.
4. Observe that:
   - Required checks remain correct, but  
   - Stale non-blocking failures may still remain visible in the bot comment, or  
   - Behavior may differ when presubmit job names were renamed.

- https://github.com/kubernetes/kubernetes/pull/134827

This behavior was discussed in SIG Testing Slack, where it was noted that:

- `/skip` may not fully clear stale bot comment results.
- The issue may be worse when presubmit job names are renamed.
- Prow comment state appears to be delta-updated rather than fully recomputed.

ghost:
### What I did
- Updated Deck PR status rendering to build the context list strictly from current ProwJobs.
- GitHub contexts are now only used to enrich existing ProwJob entries.
- GitHub-only contexts without a corresponding ProwJob are dropped, preventing stale or renamed presubmit jobs from appearing in the PR UI.

### Related issue
Fixes: [#36399](https://github.com/kubernetes/test-infra/issues/36399)

### What still needs to be done
- Redefine how Prow comments are rebuilt, so they are fully recomputed instead of delta-updated.
- Handle presubmit job renames explicitly, ensuring old job entries are removed from comments.
- Align expected `/skip` behavior with comment state reset semantics.

This was not addressed here because it is a larger behavioral change and needs agreement from the Prow team before implementation.

### DIFF
--- a/cmd/deck/static/pr/pr.ts
+++ b/cmd/deck/static/pr/pr.ts
@@ -328,53 +328,48 @@ function createSearchCard(): HTMLElement {
 }
 
 /**
- * GetFullPRContexts gathers build jobs and pr contexts. It firstly takes
- * all pr contexts and only replaces contexts that have existing Prow Jobs. Tide
- * context will be omitted from the list.
+ * GetFullPRContexts gathers build jobs and PR contexts.
+ * ProwJobs are the source of truth; GitHub contexts only enrich existing jobs.
+ * Stale contexts without a corresponding ProwJob are dropped.
  */
 function getFullPRContext(builds: ProwJob[], contexts: Context[]): UnifiedContext[] {
   const contextMap: Map<string, UnifiedContext> = new Map();
-  if (contexts) {
-    for (const context of contexts) {
-      if (context.Context === "tide") {
-        continue;
-      }
-      contextMap.set(context.Context, {
-        context: context.Context,
-        description: context.Description,
-        discrepancy: null,
-        state: context.State.toLowerCase() as UnifiedState,
-      });
-    }
-  }
 
+  // Build contexts strictly from current ProwJobs
   for (const build of builds) {
     const {
-      spec: {
-        context = "",
-      },
-      status: {
-        url = "", description = "", state = "",
-      },
+      spec: { context = "" },
+      status: { url = "", description = "", state = "" },
     } = build;
 
-    let discrepancy = null;
-    // If GitHub context exits, check if mismatch or not.
-    if (contextMap.has(context)) {
-      const githubContext = contextMap.get(context)!;
-      // TODO (qhuynh96): ProwJob's states and GitHub contexts states
-      // are not equivalent in some states.
-      if (githubContext.state !== state) {
-        discrepancy = "GitHub context and Prow Job states mismatch";
-      }
-    }
     contextMap.set(context, {
       context,
       description,
-      discrepancy,
+      discrepancy: null,
       state,
       url,
     });
+  }
+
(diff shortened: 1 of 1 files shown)


kkauy:
Thanks for linking this to #36399 and working on the Deck-side cleanup.

Just to confirm: this addresses the stale non-blocking contexts in the PR UI, while the separate issues around prow comment delta-updates and `/skip` reset behavior remain open follow-ups, correct?

ghost:
> Thanks for linking this to #36399 and working on the Deck-side cleanup.
> 
> Just to confirm: this addresses the stale non-blocking contexts in the PR UI, while the separate issues around prow comment delta-updates and `/skip` reset behavior remain open follow-ups, correct?

This PR focuses specifically on the Deck-side cleanup by filtering out those stale contexts from the UI. The issues regarding Prow comment delta-updates and the /skip behavior are indeed separate and will remain as follow-ups :) 
But there are still some things that need to be done, as I mentioned in the description.

kkauy:
Thanks for the clarification.
Glad to see the Deck-side cleanup handled separately from the broader issues around Prow comment delta updates and /skip reset behavior. I’d be interested in helping with the follow-up work once the direction is clearer. Please let me know if there’s an existing discussion or preferred approach I should align with.

ghost:
> Thanks for the clarification.
> Glad to see the Deck-side cleanup handled separately from the broader issues around Prow comment delta updates and /skip reset behavior. I’d be interested in helping with the follow-up work once the direction is clearer. Please let me know if there’s an existing discussion or preferred approach I should align with.
> 

Sure @kkauy , as soon as this pull request is merged, I'll start the continuation and let you know there :)

kkauy:
Sounds good, thanks!

I’ll keep an eye on this and would be happy to help with the follow-up once it’s merged.

ghost:
> Sounds good, thanks!
> 
> I’ll keep an eye on this and would be happy to help with the follow-up once it’s merged.

Sure, thank you for that

ivankatliarchuk (commented):
Am I correctly understand this

The PR title says "non-blocking presubmit contexts" but the change drops all GitHub-only contexts. If a PR has required status checks from non-Prow CI (GitHub Actions,   external bots, etc.), they will silently disappear from Deck's PR status view.

 No test files exist for getFullPRContext and none are added. The behavioral change is non-trivial and this function deserves at minimum a few unit tests covering:
  - ProwJob-only context (no GitHub counterpart)
  - GitHub-only context (should now be dropped)
  - Matching pair with state mismatch → discrepancy populated
  - Matching pair with state match → no discrepancy

Issue #302, opened 2024-10-16
Add Environment Variable For SRC_BASE And SRC_HOST


midnightconman:
We have `REPO_NAME` and `REPO_OWNER`, but we don't have `SRC_BASE` (`GOPATH`+`/src`) and `SRC_HOST` (which is normally `github.com`, but it can be different for github enterprise). This would allow building extra_ref paths instead of using relative paths for executables.

midnightconman:
@petr-muller Can you reopen this issue, please? I'll make a PR for the changes 👍

PR #616, opened 2026-02-10
Added SRC_BASE And SRC_HOST PodUtils Environment Variables

midnightconman:
Fixes https://github.com/kubernetes-sigs/prow/issues/302

We have `REPO_NAME` and `REPO_OWNER`, but we don't have `SRC_BASE` (`GOPATH`+`/src`) and `SRC_HOST` (which is normally `github.com`, but it can be different for github enterprise). This would allow building extra_ref paths instead of using relative paths for executables.


midnightconman:
@Prucek PTAL please 😸 All tests passing! Thank you! ❤️

midnightconman:
/cc @krzyzacy 

Can I trouble you for a quick PR review, please? 😸

midnightconman:
Thank you! ❤️

PR #626, opened 2026-02-18
feat(clonerefs): add sparse-checkout mode

Prucek:
Adds `SparseCheckoutFiles` to Refs, which limits the working tree to specific files or directories without a full clone.                                                                                                               

  When set, clonerefs:
  - runs git sparse-checkout init + git fetch --filter=blob:none --no-tags --depth 1
  - applies git sparse-checkout set <paths> before checkout

  Benchmarked against kubernetes/kubernetes fetching a single Makefile:

| | time | RSS | files | disk |                                                                                                                                                                                                      
  |---|---|---|---|---|                                                                                                                                                                                                              
  | regular clone | 35s | 121 MB | 28,358 | 400 MB |                                                                                                                                                                                   
  | sparse checkout | 1.7s | 31 MB | 16 | 5 MB | 

  Usage:
  ```
  refs:
    org: kubernetes
    repo: kubernetes
    base_ref: main
    sparse_checkout_files:
      - Makefile
      - pkg/operator

### DIFF
--- a/pkg/apis/prowjobs/v1/types.go
+++ b/pkg/apis/prowjobs/v1/types.go
@@ -1232,6 +1232,18 @@ type Refs struct {
 	// using the --filter=blob:none flag. If unspecified, defaults to
 	// DecorationConfig.BloblessFetch.
 	BloblessFetch *bool `json:"blobless_fetch,omitempty"`
+	// SparseCheckoutFiles limits the working tree to only the listed paths.
+	// Accepts the same patterns as git sparse-checkout set: file names,
+	// directory names, and gitignore-style globs (e.g. "Makefile",
+	// "pkg/operator", "config/**/*.yaml").
+	//
+	// When set, clonerefs will:
+	//   1. run git sparse-checkout init to enable sparse mode
+	//   2. run git fetch with --depth 1 --filter=blob:none --no-tags
+	//   3. run git sparse-checkout set <paths> before checkout
+	//
+	// Only the blobs needed for the requested paths are downloaded.
+	SparseCheckoutFiles []string `json:"sparse_checkout_files,omitempty"`
 }
 
 func (r Refs) String() string {
--- a/pkg/pod-utils/clone/clone.go
+++ b/pkg/pod-utils/clone/clone.go
@@ -239,16 +239,26 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
 		commands = append(commands, g.gitCommand("config", "http.cookiefile", cookiePath))
 	}
 
+	sparseCheckoutSet := isSparseCheckoutSet(refs)
+	if sparseCheckoutSet {
+		commands = append(commands, g.gitCommand("sparse-checkout", "init"))
+	}
+
+	cloneDepth := refs.CloneDepth
+	if sparseCheckoutSet && cloneDepth == 0 {
+		cloneDepth = 1
+	}
 	var depthArgs []string
-	if d := refs.CloneDepth; d > 0 {
-		depthArgs = append(depthArgs, "--depth", strconv.Itoa(d))
+	if cloneDepth > 0 {
+		depthArgs = append(depthArgs, "--depth", strconv.Itoa(cloneDepth))
 	}
 	var filterArgs []string
-	if refs.BloblessFetch != nil && *refs.BloblessFetch {
+	if (refs.BloblessFetch != nil && *refs.BloblessFetch) || sparseCheckoutSet {
 		filterArgs = append(filterArgs, "--filter=blob:none")
 	}
 
-	if !refs.SkipFetchHead {
+	skipTagFetch := refs.SkipFetchHead || sparseCheckoutSet
+	if !skipTagFetch {
 		var fetchArgs []string
 		fetchArgs = append(fetchArgs, depthArgs...)
 		fetchArgs = append(fetchArgs, filterArgs...)
@@ -270,8 +280,16 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
 		var fetchArgs []string
 		fetchArgs = append(fetchArgs, depthArgs...)
 		fetchArgs = append(fetchArgs, filterArgs...)
+		if sparseCheckoutSet {
+			fetchArgs = append(fetchArgs, "--no-tags")
+		}
 		fetchArgs = append(fetchArgs, g.repositoryURI, fetchRef)
 		commands = append(commands, g.gitFetch(fetchArgs...))
+	}
+
+	if sparseCheckoutSet {
+		sparseArgs := append([]string{"sparse-checkout", "set"}, refs.SparseCheckoutFiles...)
+		commands = append(commands, g.gitCommand(sparseArgs...))
 	}
 
 	// we need to be "on" the target branch after the sync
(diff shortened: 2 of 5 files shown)


pkg/pod-utils/clone/clone.go
```diff
@@ -39,6 +43,113 @@ type runnable interface {
...
 }
 
+// apiClient handles GitHub Contents API requests
+type apiClient struct {
+	client  *http.Client
+	baseURL string
+	token   string
+}
+
+func newAPIClient(refs prowapi.Refs, token string) *apiClient {
```
droslean: I don't think that you should create a new client. Why can't you just use the existing one and expand the call that needs to be used?

pkg/pod-utils/clone/clone.go
```diff
@@ -39,6 +41,86 @@ type runnable interface {
...
+
+// fileGetterFunc is a function type that implements FileGetter
+type fileGetterFunc func(org, repo, filepath, commit string) ([]byte, error)
+
+func (f fileGetterFunc) GetFile(org, repo, filepath, commit string) ([]byte, error) {
+	return f(org, repo, filepath, commit)
+}
+
+// fetchFilesViaAPI downloads specific files using the GitHub Contents API
+func fetchFilesViaAPI(refs prowapi.Refs, dir string, client FileGetter) Record {
```
droslean: This looks like that needs to be a method of the github client interface.

pkg/apis/prowjobs/v1/types.go
```diff
@@ -1232,6 +1232,10 @@ type Refs struct {
 	// using the --filter=blob:none flag. If unspecified, defaults to
 	// DecorationConfig.BloblessFetch.
 	BloblessFetch *bool `json:"blobless_fetch,omitempty"`
+	// FetchFilesOnly specifies file paths to fetch via GitHub API instead of git clone.
+	// When set, clonerefs will use the GitHub Contents API to download only these files.
+	// Paths can be files or directories (e.g., "Dockerfile", "dockerfiles/").
+	FetchFilesOnly []string `json:"fetch_files_only,omitempty"`
```
droslean: Do we need to support regexing here?

droslean: One more thing. The feature seems to not support private repos. We need to document that or since you are doing that, you can add the private repo support as well. Check the decoration config.

petr-muller:
I'm not a big fan of adding GH API file fetches to clonerefs, feels a bit arbitrary. I see some utility in fetching only some files sometimes but using sparse checkout would be less added complexity.

/hold

pkg/pod-utils/clone/clone.go
```diff
@@ -240,15 +240,20 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
 	}
 
 	var depthArgs []string
+	if len(refs.SparseCheckoutFiles) > 0 {
```
petr-muller: We use the len check at multiple places for "sparse checkout requested" check, let's extract it to a bool (`sparseCheckout := len(...) > 0`) and use it in conditions for better readability

pkg/pod-utils/clone/clone.go
```diff
@@ -240,15 +240,20 @@ func (g *gitCtx) commandsForBaseRef(refs prowapi.Refs, gitUserName, gitUserEmail
...
+		refs.CloneDepth = 1
```
petr-muller: I don't like messing with `refs` - that's an input param and changing it is unexpected. If we have two factors that influence clonde depth just make a dedicated var for it.

Also, is the reset to one *always* desirable?

pkg/pod-utils/clone/clone.go
petr-muller, line 227: Only `commandsForBaseRef` is modified which makes sense for your use case. Have you thought about the interaction when PR refs (handled in `commandsForPullRefs`) are present?

pkg/pod-utils/clone/clone.go
Prucek: I wanted to be as efficient as possible. I think setting the depth to 1 when not specified otherwise and without a PR makes sense. When a pull is specified, another fetch is happening, and there we won't specify the depth.

pkg/pod-utils/clone/clone.go
Prucek: Good point. I added a unit test to check what command will run and also I ran a script on a real Prowjob refs data to verify, and it worked:
```
CLONEREFS=${1:-./clonerefs}
REPO_URI="https://github.com/openshift/release.git"
TARGET_DIR="Makefile"
BASE_SHA="7fc2f391a0e59fa32b4440b08b3983a1756e9abf"
PR_SHA="cb1ec7b45876bf11bc66d7e4037b70071c714a60"

die() { echo "ERROR: $*" >&2; exit 1; }
[ -x "$CLONEREFS" ] || die "clonerefs not found at $CLONEREFS"
command -v /usr/bin/time &>/dev/null || die "/usr/bin/time required"

WORKDIR=$(mktemp -d /tmp/clonerefs-bench-XXXXX)
trap 'rm -rf "$WORKDIR"' EXIT

run_once() {
  local mode=$1 opts_json=$2
  local clone_root log_file time_out
  clone_root=$(mktemp -d "$WORKDIR/${mode}-XXXXX")
  log_file="$WORKDIR/${mode}.log"

  local opts
  opts=$(printf '%s' "$opts_json" \
    | sed "s|__SRC__|$clone_root|g; s|__LOG__|$log_file|g")

  printf "  running %-8s ... " "$mode"

  time_out=$(CLONEREFS_OPTIONS="$opts" /usr/bin/time -v "$CLONEREFS" 2>&1 >/dev/null)

  local wall rss
  wall=$(printf '%s' "$time_out" | awk '/Elapsed.*wall/{match($0,/[0-9]+:[0-9]+\.[0-9]+/); print substr($0,RSTART,RLENGTH)}')
  rss=$(printf  '%s' "$time_out" | awk '/Maximum resident/{print $NF}')
  local secs
  secs=$(printf '%s' "$wall" | awk -F: '{printf "%.2f", $1*60+$2}')

  # checked-out file count and disk (excluding .git)
  local files kb
  files=$(find "$clone_root" -type f ! -path '*/.git/*' 2>/dev/null | wc -l)
  kb=$(du -sk "$clone_root" 2>/dev/null | cut -f1)

  printf "done\n"
  printf "    wall time : %ss\n"     "$secs"
  printf "    peak RSS  : %s kB\n"   "$rss"
  printf "    files out : %d\n"      "$files"
  printf "    disk used : %d kB\n"   "$kb"

  # stash for comparison
  printf '%s %s %s %s\n' "$secs" "$rss" "$files" "$kb" > "$WORKDIR/${mode}.result"
}

# ── CLONEREFS_OPTIONS payloads ────────────────────────────────────────────────

REGULAR_OPTS=$(cat <<EOF
{
  "src_root": "__SRC__",
  "log":      "__LOG__",
  "git_user_name":  "ci-robot",
  "git_user_email": "ci-robot@k8s.io",
  "refs": [{
    "org":      "openshift",
    "repo":     "release",
    "base_ref": "main",
    "base_sha": "$BASE_SHA",
    "clone_uri": "$REPO_URI",
    "pulls": [{
      "number": 75147,
      "sha":    "$PR_SHA"
    }]
  }]
}
EOF
)

SPARSE_OPTS=$(cat <<EOF
{
  "src_root": "__SRC__",
  "log":      "__LOG__",
  "git_user_name":  "ci-robot",
  "git_user_email": "ci-robot@k8s.io",
  "refs": [{
    "org":      "openshift",
    "repo":     "release",
    "base_ref": "main",
    "base_sha": "$BASE_SHA",
    "clone_uri": "$REPO_URI",
    "pulls": [{
      "number": 75147,
      "sha":    "$PR_SHA"
    }],
    "sparse_checkout_files": ["$TARGET_DIR"]
  }]
}
EOF
)

echo "Repo  : $REPO_URI"
echo "Target: $TARGET_DIR"
echo "Binary: $CLONEREFS"
echo

echo "── Regular clone ──"
run_once "regular" "$REGULAR_OPTS"
echo

echo "── Sparse checkout (Makefile only) ──"
run_once "sparse" "$SPARSE_OPTS"
echo

PR #627, opened 2026-02-19
feat(trigger): highlight org invite after 3 merged PRs

### LINKED ISSUE #180: Feature Request: Issue Org Invitations
Migrated from [the original issue](https://github.com/kubernetes/test-infra/issues/13371).

 **What would you like to be added**: It would be nice to have trigger include a bold, explicit invitation to join the org in its ok-to-test message after you've made a three PRs (that is, on your fourth), sample message:
 
 > **We noticed you've done this a few times! Consider [joining the Kubernetes org](https://github.com/kubernetes/community/blob/master/community-membership.md#member) to skip this step, as well as gain LGTM and other bot rights rights**. We recommend asking the approvers on your prior PRs to be your sponsors.
 
 We can do a query like `type:pr is:merged org:kubernetes author:$user` before making the comment to determine if someone is qualified - this graphql query has a flat cost of 1:
 
 ```graphql
 query {
   search(query: "type:pr is:merged org:kubernetes author:katharine", type: ISSUE) {
     issueCount
   }
 }
 ```
 
 Alternatively, we can just always comment and leave the user to figure out whether they count, but I think this diminishes the effectiveness over a positive assertion.

If there are issues with doing a count over the whole org, then doing it just on the current repository would be adequate.
 
 **Why is this needed**: Joining the org is pretty intimidating. On top of that, it is wildly unclear when people are eligible to join, and even if they clearly are, it often doesn't occur to them to actually do so.  An estimated 1/3 of our regular contributors are not org members.  While this doesn't help contributors who do not make PRs who would be eligible for org membership, it would be helpful to those who do. The real bar is much lower than non-members tend to think it is.
 
 We currently mention that "regular contributors should join the org to skip this step", but it's not at all noticeable and "regular contributors" is wildly open to interpretation.
 
 (shoutout to @liztio for the concept, and @Katharine for the original writeup)

/kind feature
/sig contributor-experience
/priority backlog
/help

miltalex:
- Added a merged-PR threshold check (>= 3) for untrusted PR authors, so on their 4th+ PR they get a bold explicit org-invitation message.
- Kept behavior safe on lookup failures: if search fails, trigger falls back to the existing generic “Regular contributors should join the org” guidance.
- Used org-scoped search query:
   - `type:pr is:merged org:<org> author:<author>`
   
Fixes #180

### DIFF
--- a/pkg/plugins/trigger/pull-request.go
+++ b/pkg/plugins/trigger/pull-request.go
@@ -92,7 +92,7 @@ func handlePR(c Client, trigger plugins.Trigger, pr github.PullRequestEvent) err
 			return buildAllButDrafts(c, &pr.PullRequest, pr.GUID, baseSHA, presubmits)
 		}
 		c.Logger.Infof("Welcome message to PR author %q.", author)
-		if err := welcomeMsg(c.GitHubClient, trigger, pr.PullRequest); err != nil {
+		if err := welcomeMsg(c, trigger, pr.PullRequest); err != nil {
 			return fmt.Errorf("could not welcome non-org member %q: %w", author, err)
 		}
 	case github.PullRequestActionReopened:
@@ -248,7 +248,7 @@ func buildAllIfTrusted(c Client, trigger plugins.Trigger, pr github.PullRequestE
 	return nil
 }
 
-func welcomeMsg(ghc githubClient, trigger plugins.Trigger, pr github.PullRequest) error {
+func welcomeMsg(c Client, trigger plugins.Trigger, pr github.PullRequest) error {
 	var errors []error
 	org, repo, a := orgRepoAuthor(pr)
 	author := string(a)
@@ -279,10 +279,12 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
 </details>
 `, author, encodedRepoFullName, plugins.AboutThisBotWithoutCommands)
 	} else {
-		membershipGuidance := orgInvitationGuidance(ghc, org, author, joinOrgURL)
+		membershipGuidance := orgInvitationGuidance(c, org, pr.User, joinOrgURL)
 		comment = fmt.Sprintf(`Hi @%s. Thanks for your PR.
 
-I'm waiting for a [%s](https://%s/orgs/%s/people) %smember to verify that this patch is reasonable to test. If it is, they should reply with `+"`/ok-to-test`"+` on its own line. Until that is done, I will not automatically test new commits in this PR, but the usual testing commands by org members will still work. %s
+I'm waiting for a [%s](https://%s/orgs/%s/people) %smember to verify that this patch is reasonable to test. If it is, they should reply with `+"`/ok-to-test`"+` on its own line. Until that is done, I will not automatically test new commits in this PR, but the usual testing commands by org members will still work.
+
+%s
 
 Once the patch is verified, the new status will be reflected by the `+"`%s`"+` label.
 
@@ -294,20 +296,20 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
 </details>
 `, author, org, github.DefaultHost, org, more, membershipGuidance, labels.OkToTest, encodedRepoFullName, plugins.AboutThisBotWithoutCommands)
 
-		l, err := ghc.GetIssueLabels(org, repo, pr.Number)
+		l, err := c.GitHubClient.GetIssueLabels(org, repo, pr.Number)
 		if err != nil {
 			errors = append(errors, err)
 		} else if !github.HasLabel(labels.OkToTest, l) {
 			// It is possible for bots and other automations to automatically
 			// add the ok-to-test label. If that's the case, then we will not
 			// add the needs-ok-to-test-label any more.
-			if err := ghc.AddLabel(org, repo, pr.Number, labels.NeedsOkToTest); err != nil {
+			if err := c.GitHubClient.AddLabel(org, repo, pr.Number, labels.NeedsOkToTest); err != nil {
 				errors = append(errors, err)
 			}
 		}
 	}
 
-	if err := ghc.CreateComment(org, repo, pr.Number, comment); err != nil {
+	if err := c.GitHubClient.CreateComment(org, repo, pr.Number, comment); err != nil {
 		errors = append(errors, err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/plugins/trigger/pull-request_test.go
```diff
@@ -631,3 +641,167 @@ func TestAbortAllJobs(t *testing.T) {
...
+				t.Fatalf("shouldHighlightJoinOrgMessage() returned error: %v", err)
+			}
+			if highlight != tc.expecting {
+				t.Fatalf("shouldHighlightJoinOrgMessage() = %t, want %t", highlight, tc.expecting)
+			}
+		})
+	}
+}
+
+func TestWelcomeMsgUsesProminentOrgInvitation(t *testing.T) {
```
Prucek: I don't think we need to be testing the welcomeMsg, you are basically testing the GH client.

pkg/plugins/trigger/pull-request.go
```diff
@@ -315,6 +317,38 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
...
+	for _, issue := range issues {
+		if !issue.IsPullRequest() || !issue.IsAuthor(author) || issue.State != github.PullRequestStateClosed {
+			continue
+		}
+
+		mergedPRCount++
+		if mergedPRCount >= mergedPRCountForProminentJoinOrgMessage {
+			return true, nil
+		}
+	}
```
Prucek: consider flipping it to a positive if, so it's easier to read

pkg/plugins/trigger/pull-request.go
```diff
@@ -315,6 +317,38 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
...
+		return defaultMessage
+	}
+
+	return fmt.Sprintf("**We noticed you've done this a few times! Consider [joining the org](%s) to skip this step and gain `/lgtm` and other bot rights.** We recommend asking approvers on your previous PRs to sponsor you.", joinOrgURL)
+}
+
+func shouldHighlightJoinOrgMessage(ghc githubClient, org, author string) (bool, error) {
+	query := fmt.Sprintf("type:pr is:merged org:%s author:%s", org, author)
+	issues, err := ghc.FindIssuesWithOrg(org, query, "", false)
+	if err != nil {
```
Prucek: I think you can make the function return just `bool` and write a comment here saying that we can eat the error here.

pkg/plugins/trigger/pull-request.go
```diff
@@ -315,6 +317,38 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
...
 }
 
+func orgInvitationGuidance(ghc githubClient, org, author, joinOrgURL string) string {
+	defaultMessage := fmt.Sprintf("Regular contributors should [join the org](%s) to skip this step.", joinOrgURL)
+	shouldHighlight, err := shouldHighlightJoinOrgMessage(ghc, org, author)
+	if err != nil || !shouldHighlight {
+		return defaultMessage
+	}
+
+	return fmt.Sprintf("**We noticed you've done this a few times! Consider [joining the org](%s) to skip this step and gain `/lgtm` and other bot rights.** We recommend asking approvers on your previous PRs to sponsor you.", joinOrgURL)
```
Prucek: I think the defaultMessage is not needed, and I would flip this `if` to a positive one

pkg/plugins/trigger/pull-request.go
```diff
@@ -315,6 +317,35 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
 	return nil
 }
 
+func orgInvitationGuidance(ghc githubClient, org, author, joinOrgURL string) string {
+	if shouldHighlightJoinOrgMessage(ghc, org, author) {
+		return fmt.Sprintf("**We noticed you've done this a few times! Consider [joining the org](%s) to skip this step and gain `/lgtm` and other bot rights.** We recommend asking approvers on your previous PRs to sponsor you.", joinOrgURL)
```
petr-muller: Can we make it so that the bold encouragement is placed on its own separate paragraph? Like:

>...
>commands by org members will still work.
>
>**We noticed you've done this a few times! Consider ...**

Otherwise the sentence is still a bit sunk in the paragraph.

Actually... maybe we'd even want to use a [callout](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)? They shoudl work in the comments I think:

```
>[!TIP]
>**We noticed you've done this a few times! Consider ...**
```

=>

>[!TIP]
>**We noticed you've done this a few times! Consider ...**

pkg/plugins/trigger/pull-request.go
```diff
@@ -315,6 +317,35 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
...
+	}
+
+	return fmt.Sprintf("Regular contributors should [join the org](%s) to skip this step.", joinOrgURL)
+}
+
+func shouldHighlightJoinOrgMessage(ghc githubClient, org, author string) bool {
+	query := fmt.Sprintf("type:pr is:merged org:%s author:%s", org, author)
+	issues, err := ghc.FindIssuesWithOrg(org, query, "", false)
```
petr-muller: My main concern is  that this call is potentially API-call expensive for contributors that are not members of the org but have many PRs, because it is paginated. I imagine it can happen for e.g. bot accounts?

pkg/plugins/trigger/pull-request.go
petr-muller: The errors should get logged

pkg/plugins/trigger/pull-request.go
```diff
@@ -315,6 +317,35 @@ I understand the commands that are listed [here](https://go.k8s.io/bot-commands?
...
+	if err != nil {
+		// Search failures should not block the welcome message; fall back to the default copy.
+		return false
+	}
+
+	mergedPRCount := 0
+	for _, issue := range issues {
+		if issue.IsPullRequest() && issue.IsAuthor(author) && issue.State == github.PullRequestStateClosed {
```
petr-muller: Aren't these redundant with the `type:pr is:merged author:...` query above?

petr-muller (commented):
I'm a bit concerned about API usage. If possible I'd leave out bot-authored PRs from the PR count check entirely (they receive the comment today: https://github.com/kubernetes-sigs/prow/pull/601#issuecomment-3795697102)

pkg/plugins/trigger/pull-request.go
miltalex: That is true, this was mostly there due to the `fakeclient` I have in the test file. I can remove it and adapt it accordingly

petr-muller:
/hold cancel

:shipit:

PR #628, opened 2026-02-20
fix(status-reconciler): fail closed on config load errors and add operational metrics

### LINKED ISSUE #540: status-reconciler started retiring whole world when its configuration became corrupted
A while ago the OpenShift Prow deployment encountered a situation where `status-reconciler` unexpectedly started retiring numerous existing job results on open pull requests, like:

```
ci/prow/images — Context retired without replacement.
```

In OpenShift, components are supplied configuration through a [git-sync sidecar](https://github.com/openshift/release/blob/master/clusters/app.ci/prow/03_deployment/statusreconciler.yaml#L41-L48), and we observed that sidecar to fail fetching the repository content:

```
{"file":"main.go","line":927},"msg":"error syncing repo, will retry","error":"Run(git fetch https://github.com/openshift/release.git master --verbose --no-progress --prune --no-auto-gc --depth 1): context deadline exceeded: { stdout: \"\", stderr: \"POST git-upload-pack (317 bytes)\\nPOST git-upload-pack (272 bytes)\" }","failCount":1}
{"logger":"","ts":"2025-10-14 02:46:52.486169","caller":{"file":"main.go","line":1350},"msg":"repo contains lock file","error":null,"path":"/tmp/git-sync/.git/shallow.lock"}
{"logger":"","ts":"2025-10-14 02:46:52.486222","caller":{"file":"main.go","line":1237},"level":0,"msg":"repo directory was empty or failed checks","path":"/tmp/git-sync"}
```

We also observed `status-reconciler` to fail loading its config from disk:

```
{"component":"status-reconciler","error":"stat /var/repo/release/ci-operator/jobs: no such file or directory","file":"sigs.k8s.io/prow/pkg/config/agent.go:371","func":"sigs.k8s.io/prow/pkg/config.(*Agent).Start.func1","jobConfig":"/var/repo/release/ci-operator/jobs","level":"error","msg":"Error loading config.","prowConfig":"/etc/config/config.yaml","severity":"error","time":"2025-10-14T10:11:43Z"}
```

Which did not stop it from doing its job, reconciling statuses on in-flight pull requests to the the set of jobs configured for the given org/repo/branch. And because it apparently saw empty job config because it was failing to load prow and job config from disk (see above), it was reconciling the world to the desired state of "no jobs exist":

```
{"client":"github","component":"status-reconciler","duration":"12.404288439s","file":"sigs.k8s.io/prow/pkg/github/client.go:806","func":"sigs.k8s.io/prow/pkg/github.(*client).log.func2","level":"debug","msg":"CreateStatus(opendatahub-io, odh-dashboard, 8e06c54e2bd62e79b070f1492271dc87d1503233, {success Context retired without replacement. ci/prow/odh-dashboard-pr-image-mirror}) finished","severity":"debug","time":"2025-10-14T10:11:43Z"}
```

This is obviously catastrophic, because retiring existing context overwrote results of the jobs on the open PRs with false passing signal, potentially allowing the PRs to merge (I am not entirely sure about Tide behavior when it encounters a mergeable Pr with a retired green status that matches a required existing job; Tide's retest stale result logic before merge may save the day by forcing falsely retired jobs to be re-run. I will open a separate issue to check this behavior and if Tide does not behave this then we may want to make it so).

Status reconciler should be fixed so never actuate unless it has a provably good config loaded. If its dynamic config loading is failing (like it was in the example above), it should cease reconciling, and strongly signal, through metrics at least, that it is not actuating right now. We may consider flipping its liveness probe endpoint to false, to signal kubernetes that the pod is not healthy and cause it to restart.

miltalex:
## What this PR changes

  - Makes `status-reconciler` fail closed on bad config/state loads:
    - corrupted saved status now fails `Load()`
    - config-agent load failures are propagated via error handlers
    - on load failure, controller is marked unhealthy and stops reconciling

  - Adds status-reconciler metrics:
    - `status_reconciler_loaded_presubmit_count` (Gauge)
    - `status_reconciler_contexts_retired_total{org,repo}` (Counter)

  - Refactors config-agent wiring for simplicity:
    - `ConfigOptions.ConfigAgent(...)` now uses options:
      - `WithErrorHandler(...)`
      - `WithAdditionals(...)`
      - `WithReuseAgent(...)`
    - migrates call sites in `status-reconciler` and `deck`

  - Keeps health wiring explicit with `ServeLive(...)`:
    - `status-reconciler` uses `ServeLive(c.Healthy)` for liveness checks
    - other binaries keep the existing explicit `ServeLive()` approach

  - Keeps status-reconciler liveness/readiness probes in starter manifests.

  ## Outcome

  - Status-reconciler no longer reconciles from failed/corrupted loads.
  - On load failures, liveness/metrics clearly signal degraded state and reconciliation is stopped.

  Fixes #540


pkg/statusreconciler/controller.go
```diff
@@ -183,6 +183,10 @@ func (c *Controller) Run(ctx context.Context) {
 }
 
 func (c *Controller) reconcile(delta config.Delta, log *logrus.Entry) error {
+	if countPresubmits(delta.Before.PresubmitsStatic) > 0 && countPresubmits(delta.After.PresubmitsStatic) == 0 {
+		return fmt.Errorf("refusing to reconcile config update because all presubmits disappeared")
+	}
+
```
Prucek: I think this is only one part of the story. Try to add the other recomendations from: https://github.com/kubernetes-sigs/prow/issues/540
> Status reconciler should be fixed so never actuate unless it has a provably good config loaded. If its dynamic config loading is failing (like it was in the example above), it should cease reconciling, and strongly signal, through metrics at least, that it is not actuating right now. We may consider flipping its liveness probe endpoint to false, to signal kubernetes that the pod is not healthy and cause it to restart.

miltalex: I pushed one commit to setup liveness/readiness for `status-reconciler`, and flipping it to false when the config Agent is failing. I am looking into pushing another commit to add metrics, since currently there are no metrics will try to cover only metrics related to this functionality. Happy to adjust/rework based on your feedback.

petr-muller:
@Prucek I'm leaving this one on you, feel free to ping me when this needs `approved`

config/prow/cluster/starter/starter-azure.yaml
```diff
@@ -573,6 +573,19 @@ spec:
...
+            port: 8081
+          initialDelaySeconds: 3
+          periodSeconds: 3
+        readinessProbe:
+          httpGet:
+            path: /healthz/ready
+            port: 8081
+          initialDelaySeconds: 10
+          periodSeconds: 3
+          timeoutSeconds: 600
```
Prucek: I think 600s is too long for a health check

pkg/pjutil/health.go
```diff
@@ -31,8 +32,13 @@ const healthPort = 8081
 // Health keeps a request multiplexer for health liveness and readiness endpoints
 type Health struct {
 	healthMux *http.ServeMux
+
+	livenessLock   sync.RWMutex
+	livenessChecks []LivenessCheck
```
Prucek: This is not needed

pkg/pjutil/health.go
```diff
@@ -42,17 +48,23 @@ func NewHealth() *Health {
...
-	healthMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) { fmt.Fprint(w, "OK") })
-	server := &http.Server{Addr: ":" + strconv.Itoa(port), Handler: healthMux}
+	h := &Health{healthMux: http.NewServeMux()}
+	h.healthMux.HandleFunc("/healthz", h.serveLive)
+	server := &http.Server{Addr: ":" + strconv.Itoa(port), Handler: h.healthMux}
 	interrupts.ListenAndServe(server, 5*time.Second)
-	return &Health{
-		healthMux: healthMux,
-	}
+	return h
```
Prucek: I wouldn't make changes here

pkg/pjutil/health.go
```diff
@@ -66,3 +78,17 @@ func (h *Health) ServeReady(readinessChecks ...ReadinessCheck) {
...
+	h.livenessLock.RUnlock()
+	for _, livenessCheck := range livenessChecks {
+		if !livenessCheck() {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			fmt.Fprint(w, "LivenessCheck failed")
+			return
+		}
+	}
+	fmt.Fprint(w, "OK")
+}
```
Prucek: This is correct, but I would reregister the `/healthz` endpoint in a single  `ServeLive` method and add the same logic as `ServeReady`, but with the `LivenessCheck`
Then this can be called like:
```
 // in a controller
  var healthy atomic.Bool
  healthy.Store(true)

  health := pjutil.NewHealth()
  health.ServeReady()
  health.ServeLive(healthy.Load)   // passes the method as the check func

  // anywhere in your code:
  healthy.Store(false)  // liveness probe now returns 503
  healthy.Store(true)   // recovers

File: pkg/pjutil/health_test.go
Comment (Prucek) [inline pkg/pjutil/health_test.go:1]: I think tests are not needed here.

File: pkg/statusreconciler/controller_health_test.go
Comment (Prucek) [inline pkg/statusreconciler/controller_health_test.go:1]: These tests are also not needed

File: pkg/statusreconciler/controller_test.go [hunk 6]
```diff
@@ -1274,6 +1274,47 @@ func TestControllerReconcile(t *testing.T) {
 	}
 }
 
+func TestControllerReconcileRefusesDroppingAllPresubmits(t *testing.T) {
```
Prucek: Check other tests in this file, and look at how we write unit tests. Usually, you don't want to write just a single test

pkg/statusreconciler/status.go
```diff
@@ -36,6 +36,7 @@ type storedState struct {
 type statusClient interface {
 	Load() (chan config.Delta, error)
 	Save() error
+	Healthy() bool
```
Prucek: This does not belong here. If you change the code according to the comment in `pkg/pjutil/health.go`, everything will be simplified.

Prucek (changes requested):
It is going in the right direction, but we need to improve:
 - the healthiness handling- I don't think the config agent should handle it. This is a tricky one: the error comes from the configAgent (`,"msg":"Error loading config."`), but it comes from a goroutine, so we can't handle it directly in the code. I think we could add a callback `onError func(error)`  to the `func (ca *Agent) Start` and then a `func (o *ConfigOptions) ConfigAgentWithErrorHandling(onError func(error), reuse ...*config.Agent)` and have a `onError` method in the controller, that will flip the healthiness to false. WDYT?
 - unit tests
- the metrics -  still missing
Also, could you please update the description to reflect what was done in this PR?

pkg/statusreconciler/controller.go
```diff
@@ -182,7 +196,53 @@ func (c *Controller) Run(ctx context.Context) {
...
+func boolToUint32(value bool) uint32 {
+	if value {
+		return 1
+	}
+	return 0
+}
+
 func (c *Controller) reconcile(delta config.Delta, log *logrus.Entry) error {
+	if countPresubmits(delta.Before.PresubmitsStatic) > 0 && countPresubmits(delta.After.PresubmitsStatic) == 0 {
```
stevekuznetsov: This seems like the wrong place - we should not transmit if we cannot load, or if the load is corrupted. You would risk not catching real deletes with this logic. The go-unavailable-on-failed-load logic mentioned below is a good idea, as well as root-causing why corrupted loads are sending deltas.

pkg/pjutil/health.go
miltalex: Thank you very much for the tip, Indeed this is much cleaner

pkg/statusreconciler/controller.go
miltalex: Thank you very much @stevekuznetsov . I moved the logic to the `Load` instead

pkg/config/agent.go
```diff
@@ -302,12 +308,25 @@ func watchConfigs(ca *Agent, prowConfig, jobConfig string, supplementalProwConfi
 // will log the failure message but continue attempting to load.
 // This function will replace Start in a future release.
 func (ca *Agent) StartWatch(prowConfig, jobConfig string, supplementalProwConfigDirs []string, supplementalProwConfigsFileNameSuffix string, additionals ...func(*Config) error) error {
+	return ca.StartWatchWithErrorHandler(nil, prowConfig, jobConfig, supplementalProwConfigDirs, supplementalProwConfigsFileNameSuffix, additionals...)
+}
+
+// StartWatchWithErrorHandler behaves like StartWatch and calls onError on config load failures.
+func (ca *Agent) StartWatchWithErrorHandler(onError func(error), prowConfig, jobConfig string, supplementalProwConfigDirs []string, supplementalProwConfigsFileNameSuffix string, additionals ...func(*Config) error) error {
```
Prucek: Is the change to `StartWatch` needed? It is not used anywhere

pkg/statusreconciler/controller.go
```diff
@@ -67,6 +68,10 @@ func NewController(continueOnError bool, addedPresubmitDenylist, addedPresubmitD
 		},
 		statusClient: sc,
 	}
+	controller.setHealthy(true)
+	controller.setActuationEnabled(false)
```
Prucek: I don't understand the purpose of this variable; mostly, it is set to the same as `healthy`. Could you explain?

cmd/admission/main.go
```diff
@@ -70,6 +70,7 @@ func main() {
 
 	pprof.Instrument(o.instrumentationOptions)
 	health := pjutil.NewHealthOnPort(o.instrumentationOptions.HealthPort)
+	health.ServeLive()
```
Prucek, line 73: If other components would like to add live health checks, they can do that on a follow-up PR. I wouldn't put it here.

pkg/flagutil/config/config.go
```diff
@@ -88,9 +93,14 @@ func (o *ConfigOptions) ConfigAgent(reuse ...*config.Agent) (*config.Agent, erro
...
 }
 
 func (o *ConfigOptions) ConfigAgentWithAdditionals(ca *config.Agent, additionals []func(*config.Config) error) (*config.Agent, error) {
-	return ca, ca.Start(o.ConfigPath, o.JobConfigPath, o.SupplementalProwConfigDirs.Strings(), o.SupplementalProwConfigsFileNameSuffix, additionals...)
+	return o.ConfigAgentWithAdditionalsAndErrorHandling(ca, additionals, nil)
+}
+
+// ConfigAgentWithAdditionalsAndErrorHandling starts the config agent with custom additionals and optional error handling.
+func (o *ConfigOptions) ConfigAgentWithAdditionalsAndErrorHandling(ca *config.Agent, additionals []func(*config.Config) error, onError func(error)) (*config.Agent, error) {
+	return ca, ca.StartWithErrorHandler(onError, o.ConfigPath, o.JobConfigPath, o.SupplementalProwConfigDirs.Strings(), o.SupplementalProwConfigsFileNameSuffix, additionals...)
```
Prucek: I think we are adding too many methods here. Also, the name gets pretty long 😄 
I'd suggest something like: 
```
type ConfigAgentOption func(*configAgentOptions)

type configAgentOptions struct {
    onError     func(error)
    additionals []func(*config.Config) error
    reuse       *config.Agent
}

func WithErrorHandler(fn func(error)) ConfigAgentOption {
    return func(o *configAgentOptions) {
        o.onError = fn
    }
}

func WithAdditionals(a ...func(*config.Config) error) ConfigAgentOption {
    return func(o *configAgentOptions) {
        o.additionals = a
    }
}

func WithReuseAgent(a *config.Agent) ConfigAgentOption {
    return func(o *configAgentOptions) {
        o.reuse = a
    }
}
```
Then we could just do:
```
agent, err := opts.ConfigAgent()

agent, err := opts.ConfigAgent(
    WithErrorHandler(handleErr),
)

agent, err := opts.ConfigAgent(
    WithReuseAgent(existing),
    WithAdditionals(extraFn),
)

File: pkg/statusreconciler/metrics.go [hunk 13]
```diff
@@ -0,0 +1,38 @@
...
+limitations under the License.
+*/
+
+package statusreconciler
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var statusReconcilerMetrics = struct {
+	controllerHealthy prometheus.Gauge
+	actuationEnabled  prometheus.Gauge
```
Prucek: I think it does not make sense to store healthyness to metrics. With the metrics, we want to measure some value over time. This could be:
1. status_reconciler_loaded_presubmit_count — Gauge
Set to len(config.PresubmitsStatic) each time a Delta is received.
2. status_reconciler_contexts_retired_total{org, repo} — Counter
Incremented in retireRemovedContexts() for each context actually retired

cmd/admission/main.go
miltalex: As discussed through slack, this is due to the changes in the `NewHealthOnPort` func

pkg/statusreconciler/metrics.go
miltalex: Thanks tbh I wasnt sure what kind of metrics we needed to expose

pkg/flagutil/config/config.go
miltalex: Thank you very much I find this indeed much cleaner

pkg/statusreconciler/controller.go
miltalex: Removed it, was trying to measure when the app was indeed not healthy. Initially when I started was trying to calculate the actuation of each config.

Prucek:
/test pull-prow-integration
/lgtm

@miltalex could you update the PR title and description? Otherwise, it looks OK.

miltalex:
@Prucek thank you very much for the review. I have updated the title and description, I kept the convo `fix(status-reconciler)` since the changes are mostly targeted to that package.

petr-muller:
I like the changes done here, but the PR grew in scope quite significantly, and I feel uncomfortable landing a change this broad as a single squashed commit that brings several loosely coupled changes across the whole Prow. Also, some `pkg/` changes are breaking downstream consumers (which do exist) more than I'd like (ConfigAgent constructor, `NewHealth/OnPort` needing `ServeLive` later). 

I'll need to figure out how to land this safely. I think the minimum would be to break this into targeted commits with individual changes, and I'd like to check if we can be friendlier to downstream consumers by providing a clear porting path (maybe provide a compatibility method they'd just rename to? not sure) and making somehow sure existing callers of `NewHealth` either dont see a behavior change or at least something really points out that they need to add subsequent `ServeLive` call.

I would not be opposed to extracting the simpler, more straightforward parts of this (actual loader fix, metrics) to a separate PRs or multiple that would be easy to merge fast.

PR #639, opened 2026-03-05
feat(decoration): add SparseCheckoutFiles to DecorationConfig

Prucek:
Add `SparseCheckoutFiles` field to `DecorationConfig` to allow jobs to set/limit the working tree to only the paths needed. The field is propagated only to the primary ref (presubmit/postsubmit) via `CompletePrimaryRefs` and is not applied to extra refs. 

Follow up of: https://github.com/kubernetes-sigs/prow/pull/626

### DIFF
--- a/pkg/pjutil/pjutil.go
+++ b/pkg/pjutil/pjutil.go
@@ -288,7 +288,11 @@ func CompletePrimaryRefs(refs prowapi.Refs, jb config.JobBase) *prowapi.Refs {
 	if jb.SkipFetchHead {
 		refs.SkipFetchHead = jb.SkipFetchHead
 	}
-	return DecorateRefs(refs, jb)
+	drefs := DecorateRefs(refs, jb)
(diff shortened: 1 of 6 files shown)


pkg/pjutil/pjutil.go
```diff
@@ -288,7 +288,11 @@ func CompletePrimaryRefs(refs prowapi.Refs, jb config.JobBase) *prowapi.Refs {
 	if jb.SkipFetchHead {
 		refs.SkipFetchHead = jb.SkipFetchHead
 	}
-	return DecorateRefs(refs, jb)
+	drefs := DecorateRefs(refs, jb)
+	if dc := jb.DecorationConfig; dc != nil && len(dc.SparseCheckoutFiles) > 0 {
+		drefs.SparseCheckoutFiles = dc.SparseCheckoutFiles
+	}
```
petr-muller, line 294: doesn't this fit in DecorateRefs better?

Prucek: Yes, it does, but I don't want to change extraRefs

Issue #617, opened 2026-02-10
Add a plugin to block merging PRs with `fixup!` commits


nojnhuh:
I usually update my PRs with `git commit --fixup` to better enable more iterative review than rewriting history with `git commit --amend`. Then I `git rebase --autosquash` when the PR is LGTM'd and approved to remove all the intermediate changes.

I usually remember to `/hold` those PRs to remind myself to squash but sometimes end up merging PRs anyway with lots of unnecessary `fixup!` noise in the git log. It would be great if there was a plugin like `mergecommitblocker` that could look for commits whose messages start with `fixup!` or `amend!` and automatically add a `do-not-merge/*` label, and remove it when no such commits exist like after a `git rebase --autosquash`.

/kind feature
/area plugins

petr-muller:
Sounds like this could go into the existing `invalidcommitmsg` plugin?

petr-muller:
There may be some configuration / logic necessary to accommodate repos that merge with squash method, and possibly also the rebase method (not sure if GH rebase method applies the `fixup!` squash.

valen-mascarenhas14:
> There may be some configuration / logic necessary to accommodate repos that merge with squash method, and possibly also the rebase method (not sure if GH rebase method applies the `fixup!` squash.

@petr-muller I feel it may be better not to rely on the repository’s merge strategy for enforcement. GitHub’s **Rebase & merge** does not apply `--autosquash`, so `fixup!` commits would be preserved in the target branch unless cleaned up beforehand. Even with **Squash and merge**, while the individual commits wouldn’t appear separately, their messages could still end up in the final commit if it's not edited.

Since merge strategy is a repo-level setting & can change over time, it might be more robust to validate the PR commits directly & require `fixup!` commits to be cleaned up before merge, regardless of how the repository merges PRs.

petr-muller:
What's acceptable to merge is mostly repository maintainers' decision, not Prow's (=ours). Blocking PRs from merge is a big hammer and Prow is used in different contexts. I think the check is useful enough for many repositories that I'd like the behavior to be on by default and ideally packed into the existing plugin - that way the value propagates to existing Prow instances without any config change needed.

[...]

PR #641, opened 2026-03-07
invalidcommitmsg: add separate handling for fixup!/amend! commits

valen-mascarenhas14:
Fix for #617 
This change extends the `invalidcommitmsg` plugin to detect temporary commits such as `fixup!` and `amend!`.

Previously, the plugin only checked for issue-closing keywords in commit messages and PR titles and applied the `do-not-merge/invalid-commit-message` label.

### Changes
- Detect `fixup!` and `amend!` commits in pull request history
- Apply a separate label: `do-not-merge/fixup-commits`
- Add a dedicated comment explaining the presence of fixup/amend commits
- Keep issue-closing keyword validation unchanged
- Allow disabling the fixup check via plugin configuration

The fixup/amend commit check is **enabled by default**.

Repositories that want to opt out can disable it in the plugin configuration:

```yaml
invalid_commit_msg:
- repos:
  - org/repo
  disable_fixup_check: true

### CODE REVIEW

File: pkg/plugins/invalidcommitmsg/invalidcommitmsg.go [hunk 1]
```diff
@@ -142,6 +180,12 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
 		})
 	}
 
+	if hasFixupCommitMsgLabel && len(fixupCommits) == 0 {
+		if err := gc.RemoveLabel(org, repo, number, fixupCommitMsgLabel); err != nil {
```
ivankatliarchuk, line 178: Most likely should prune fixup comments here too. When hasFixupCommitMsgLabel && len(fixupCommits) == 0, the label is removed but the fixup comment is never pruned. Compare with the existing invalidCommitMsgLabel path which calls cp.PruneComments on removal

Could be that, worth to review/triage
```
if hasFixupCommitMsgLabel && len(fixupCommits) == 0 {
      if err := gc.RemoveLabel(...); err != nil { ... }
      // missing: cp.PruneComments(func(...) bool { return strings.Contains(comment.Body, fixupCommitMsgCommentPruneBody) })
  }

```

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -120,10 +149,19 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
 	log.Debugf("Found %d commits in PR", len(allCommits))
 
 	var invalidCommits []github.RepositoryCommit
+	var fixupCommits []github.RepositoryCommit
+
 	for _, commit := range allCommits {
-		if CloseIssueRegex.MatchString(commit.Commit.Message) {
```
ivankatliarchuk, line 124: The PR changes the existing CloseIssueRegex check from full message to first-line only

```
// Before (matched full commit body):
  if CloseIssueRegex.MatchString(commit.Commit.Message) {

  // After (matches only subject line):
  subject := strings.Split(msg, "\n")[0]
  if CloseIssueRegex.MatchString(subject) {
```

 Potentially breaking change. Commit bodies with closing keywords (e.g. Fixes #123 in the description) would no longer trigger the label.

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -69,7 +86,7 @@ func init() {
 func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
 	// Only the Description field is specified because this plugin is not triggered with commands and is not configurable.
```
ivankatliarchuk: Is this comment still valid? The plugin is now configurable as it looks

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -58,9 +59,25 @@ When GitHub merges a Pull Request, the title is included in the merge commit. To
 </details>
 `
 	invalidTitleCommentPruneBody = "not allowed in the title of a Pull Request"
+	fixupCommitMsgCommentBody    = `Temporary commits like fixup! or amend! are not allowed in the commit history.
```
ivankatliarchuk, line 62: The template is missing any instruction about how to squash (e.g. link to git docs, global config), while the existing plugin links to GitHub help. Could improve UX.

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -107,11 +124,23 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
...
+	for _, c := range cfg.InvalidCommitMsg {
+		for _, r := range c.Repos {
+			if r == org || r == org+"/"+repo {
+				if c.DisableFixupCheck {
+					checkFixup = false
+				}
+			}
+		}
+	}
+
```
ivankatliarchuk, line 131: Last-write -> wins config loop. Most likely worth to create a separate helper function and with number of specific scenaarios and cases tested.

checkFixup can only ever go from true → false, never back. That means the effective result is: fixup is disabled if any matching entry has disable_fixup_check: true, regardless of specificity or ordering.

This creates a surprising edge case. Suppose a kubernetes team sets an org-level default to disable the check, but one repo (external-dns) wants to opt back in:

```
  invalid_commit_msg:
  - repos:
    - my-org               # org-wide: disable fixup check
    disable_fixup_check: true
  - repos:
    - my-org/special-repo  # this repo wants it enabled
    disable_fixup_check: false
```

With the current code, my-org/special-repo will still have the check disabled, because the org entry sets checkFixup = false and the repo entry's disable_fixup_check: false branch is never entered

The conventional pattern in this codebase is repo-specific match wins over org-level match, implemented as a two-pass loop(check org/repo first, fall back to org only if no repo-level entry matched). This is used consistently by ApproveFor, LgtmFor, TriggerFor, and the Welcome plugin's optionsForRepo - all of which have the same structure:

```go
  // From config.go — TriggerFor (has an explicit comment on it)
  // "Prioritize repo level triggers over org level triggers."
  func (c *Configuration) TriggerFor(org, repo string) Trigger {
      fullName := fmt.Sprintf("%s/%s", org, repo)
      for _, trigger := range c.Triggers {           // pass 1: repo-level match
          if !sets.NewString(trigger.Repos...).Has(fullName) { continue }
          return trigger
      }
      for _, trigger := range c.Triggers {           // pass 2: org-level fallback
          if !sets.NewString(trigger.Repos...).Has(org) { continue }
          return trigger
      }
      return defaultTrigger
  }

  The PR's loop does neither — it single-passes and ORs every match with no specificity:

  // PR's approach — last DisableFixupCheck: true wins, regardless of org vs repo
  for _, c := range cfg.InvalidCommitMsg {
      for _, r := range c.Repos {
          if r == org || r == org+"/"+repo {
              if c.DisableFixupCheck {
                  checkFixup = false
              }
          }
      }
  }
```

Potential solution  is to extract a `InvalidCommitMsgFor(org, repo string) *InvalidCommitMsg` helper following the two-pass pattern - exactly how ApproveFor, LgtmFor, and TriggerFor work.

pkg/plugins/config.go
```diff
@@ -2386,3 +2387,12 @@ func (c *Configuration) HasConfigFor() (global bool, orgs sets.Set[string], repo
 
 	return global, orgs, repos
 }
+
+// InvalidCommitMsg holds configuration for the invalidcommitmsg plugin.
+type InvalidCommitMsg struct {
```
ivankatliarchuk, line 2429: No config validation. Other plugins implement ValidatePluginConfig. The Repos field in InvalidCommitMsg accepts arbitrary strings — malformed repo names (e.g. missing /, typos) would silently never match. A basic validation pass would help catch config errors early.

ivankatliarchuk (changes requested):
Test cases, worth to review.

For example consider potential tests to add
 - amend! prefix (only fixup! is tested)
 -  Label removal when fixup commits are squashed
 - disable_fixup_check: true config
 - Commit with both a fixup prefix and a close-issue keyword (both labels applied?)

No config validation. Other plugins implement ValidatePluginConfig. The Repos field in InvalidCommitMsg accepts arbitrary strings — malformed repo names (e.g. missing /, typos) would silently never match. A basic validation pass would help catch config errors early.

 1. InvalidCommitMsg doesn't implement getRepos(), so it can't use validateRepoDupes.
 2. No call to validate InvalidCommitMsg entries is added to Validate().
 
The practical example: a user who writes my-org/my-repo as my-org/myrpo (typo) in their config will get no error -> the fixup check just silently stays enabled, with no
   indication the config entry was never matched. The pattern to follow is:

```go
  // What InvalidCommitMsg should add:
  func (c InvalidCommitMsg) getRepos() []string { return c.Repos }

  // And in Validate():
  if err := validateRepoDupes(c.InvalidCommitMsg); err != nil {
      return err
  }
```

valen-mascarenhas14:
Thanks @ivankatliarchuk  will address your review comments

petr-muller:
I think at this point the PR is trying to do too much, adds two things: new commit check, and it adds a new plugin configuration surface. The new commit check is actually the easier part from the two. The config surface part is tricky to get right. It also warrants some thinking ahead, because we want good configuration surface first and code to implement it second, not the other way around. Doing both at once is hard, I'd go in increments.

I'd suggest to add the commit check itself first in a separate PR, and make it disabled by default. As a temporary measure, it could be enabled by e.g. a envvar. That would allow us to deploy it and test the functionality - it would be in a "stealth mode". Like behind a feature flag.

Then a separate PR would add the new configuration surface for the plugin. The PR would be smaller and focused only on getting the config right. Some important features for the config:

- reasonable checkconfig check making sure the config is valid
- utilize the existing org/repo sharding capabilities like e.g. `lgtm` has (`mergeFrom` and `HasConfigFor` methods in the config code)
- same as APIs, booleans are antipattern - i'd prefer to not have `disable_fixup_check` (we add a new opt-outable check, we need a new boolean). I think we could have something like

```
disabled_checks:
- fixupPrefix
```

or maybe even this extensible form which allows per-check config in the future

```
checks:
- name: fixupPrefix
  disabled: true
```

The config surface PR would also enable the new check by default (and remove the opt-out condition), because there would now be a config to opt-out.

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Noted

valen-mascarenhas14: Thanks. Have added this

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Thanks, it was an oversight on my end, have fixed this

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Have removed this comment

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Have added a link on the instructions to squash

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Thanks. Have added the helper function on setting the priority & Updating config

pkg/plugins/config.go
valen-mascarenhas14: Thanks, have added it now

valen-mascarenhas14:
Thanks @ivankatliarchuk for the detailed feedback and as pointed out by @petr-muller, the current PR mixes 2 concerns: introducing the new commit check and adding a configuration surface for the plugin.

I agree that the configuration design deserves separate attention.

I'll split this into 2 parts:

• A smaller PR that introduces the commit check only (disabled by default and gated via an env var for now).
• A follow-up PR that focuses on designing the proper configuration surface for enabling/disabling checks.

This should make the review easier and allow us to iterate on the configuration design independently. I'll update the PR accordingly.

valen-mascarenhas14:
@petr-muller Have raised the PR :
https://github.com/kubernetes-sigs/prow/pull/656

PR #646, opened 2026-03-09
Allow author to test their own prs - turned off by default

prad9192:
## What
Add allow_author_test_own_pr config option to the trigger plugin that lets PR authors run /test and /retest commands on their own PRs 

## Why
External contributors currently can't test their PRs until a trusted user approves with /ok-to-test. This slows down the feedback loop. With this flag enabled, PR authors can self-serve testing


petr-muller:
I'm fairly strongly against this. I do not expect many Prow instances to enable it and I'm not keen on adding configuration options (that increase complexity and need future maintenance) for such niche cases.

pkg/plugins/config.go
```diff
@@ -512,6 +511,12 @@ type Trigger struct {
 	IgnoreOkToTest bool `json:"ignore_ok_to_test,omitempty"`
 	// TriggerGitHubWorkflows enables workflows run by github to be triggered by prow.
 	TriggerGitHubWorkflows bool `json:"trigger_github_workflows,omitempty"`
+	// AllowAuthorTestOwnPR allows PR authors to run test commands (/test, /retest) on their own PRs
+	// even if they are not trusted users. They still cannot use /ok-to-test on their own PRs.
+	// This is useful for allowing external contributors to test their changes without waiting for approval,
+	// but still requiring a trusted user to grant permanent trust via /ok-to-test.
+	// WARNING: This allows untrusted code to run in your CI. Use with caution.
```
Priyankasaggu11929, line 518: Dropping a comment from ContribEx side.

We **_don't want_** to allow any CI tests to run without having `/ok-to-test` by a trusted user (which is basically an Kubernetes org member).

If a PR author is already an org member, they will get the `ok-to-test` already applied, and therefore can use all test commands already - which is what is intended.

prad9192:
Hey folks, thanks for all the feedback.

Let me give you more context on the use case , this might change things or might not. We are using Prow in a GitHub Enterprise setup where we have internal contributors from different orgs contributing to repos outside their org. They're not external/untrusted users, they're just in a different internal org structure. Right now they can't run /test on their own PRs (they fork the repo, make changes and create the PR to review/merge forked changes), which slows down iteration significantly. The current trust model works great for public OSS projects with truly external contributors, but for enterprise setups with cross-org collaboration, it creates friction for trusted internal users who just happen to be in a different org. That said, if this is too specific to enterprise use cases and adds too much complexity for upstream, I get it. Just wanted to make sure the context was clear before closing this out.

pkg/plugins/config.go
prad9192: Ack

ivankatliarchuk:
Org-level trust in enterprise GitHub/Gitlab/etc setups is often a deliberate security boundary - just because someone is an internal employee doesn't mean their org membership grants them CI trust in another team's repo.

Another options, not fully solves but quite close.

Option 0: Outside collaborators - Add cross-org contributors as outside collaborators on the target repo. Prow's TrustedUser check accepts repo collaborators, not just org members. No code change needed, but requires repo-level access management.

Option 1: `trusted_orgs` list in the Trigger config - Prow already supports TrustedOrg (a single org). If it were extended to a list of orgs, you could declare all your internal enterprise orgs as trusted, and any member of those orgs would be fully trusted regardless of which repo they're contributing to. This is a more semantically correct model for your use case and avoids the "untrusted code running in CI" framing entirely - your contributors are trusted, they're just in a different org

Option 2: Use GitHub Enterprise teams/CODEOWNERS - If you control the GHE instance, adding cross-org contributors to a GitHub team in the target org's org, then referencing that team in trusted_teams, solves the problem without a new flag.

Option 3: Bot auto-approval -> A separate bot watches for PRs from a known-safe user list (e.g. from a corporate directory or another trusted org) and automatically comments /ok-to-test. Removes human friction but requires a separate service.

Option 4: trusted_for_testing user/team list - A list (users or teams) that gets "test-only" trust: can trigger CI on their own PRs but can't use /ok-to-test. More explicit than allow_author_test_own_pr

prad9192:
really appreciate the detailed feedback and alternative approaches! Option 1 (trusted_orgs list) is interesting - that actually sounds like it might be the cleanest solution for our use case. We have multiple internal orgs that should be mutually trusted. Would you be open to a PR that extends trusted_orgs to support a list (I see its currently a string)? That feels more semantically correct than my current approach.

BenTheElder:
> I'm fairly strongly against this. I do not expect many Prow instances to enable it and I'm not keen on adding configuration options (that increase complexity and need future maintenance) for such niche cases.

+1, thanks for contributing but I think your use case can be handled without this change, and I would really recommend against it.

Opening up self-service testing to completely arbitrary users is a critical DOS vector at minimum, but if you really want to do this you don't even need any code, you can just put `ok-to-test` in OWNERS labels at the root of the repo.

Side note: We should perhaps prohibit this in the Kubernetes org via the verify-owners plugin ...

prad9192:
Thanks everyone for the thoughtful feedback and discussion. I really appreciate the time you took to review this and suggest alternatives.

We will do a small PoC for all the options and choose the one that works best for us.

PR #648, opened 2026-03-09
linter(modernize): modernize go code

ivankatliarchuk:
- Added modernize linter https://golangci-lint.run/docs/linters/configuration/#modernize
- Added modernize.disable: [omitzero] to .golangci.yml to suppress it across the whole codebase.

All changes made by linter itself. 

```
./_bin/golangci-lint run --fix ./...
```

- Linters in general: Catch bugs, enforce consistency, and automate code review for things humans miss or disagree about. They codify "best practices" as machine-checkable rules, so reviewers can focus on logic rather than style. 
- modernize specifically: Its value is keeping code up-to-date with Go's evolving stdlib and language features. As Go adds things like slices.Contains, min/max, omitzero, etc., old hand-rolled equivalents accumulate.


alvaroaleman:
/uncc

As I am not active in the project anymore. Thanks for your contribution!

hack/make-rules/verify/golangci-lint.sh
```diff
@@ -42,4 +42,4 @@ echo "Golangci-lint version: $(./_bin/golangci-lint version)"
 export GO111MODULE=on
 export GOPROXY=https://proxy.golang.org
 export GOSUMDB=sum.golang.org
-./_bin/golangci-lint run ./...
+./_bin/golangci-lint run --fix ./...
```
petr-muller: This is a verify script so it should just lint, not fix the files I think?

ivankatliarchuk: Removed. At the moment when executing locally, there is no command - this one could just run and fix formatting and linters.

I'll open a follow-up PR then

PR #654, opened 2026-03-12
Update golanci-lint to 2.11.3

stmcginnis:
This updates the version of golangci-lint we're using to the latest release and addresses a few new issues it found. Disabled on check as it doesn't make sense for the places it triggered warnings about.


cmd/webhook-server/main.go
```diff
@@ -153,9 +153,6 @@ func main() {
 			logrus.WithError(err).Fatal("Unable to create secretmanager client", err)
 		}
 		client = newGCPClient(secretManagerClient, o.secretID)
-		if err != nil {
-			logrus.WithError(err).Fatal("Unable to create secret manager client")
-		}
```
petr-muller, line 158: I had to look three times to figure out why this is removed :D

stmcginnis: I actually had to double check myself too. :D

petr-muller:
sry I initially mis-reviewed

PR #656, opened 2026-03-15
invalidcommitmsg: check added to detect fixup! amend! commits

### LINKED ISSUE #617: Add a plugin to block merging PRs with `fixup!` commits

valen-mascarenhas14:
Fixes #617

This change extends the `invalidcommitmsg` plugin to detect temporary commits such as `fixup!` and `amend!` in pull request history.

When such commits are present, the plugin:
- Applies the `do-not-merge/fixup-commits` label
- Posts a comment advising contributors to squash or rebase these commits before merging

The existing validation for issue-closing keywords in commit messages and PR titles remains unchanged.

The fixup/amend commit check is **disabled by default** and can be enabled by setting the following environment variable:

```bash
ENABLE_FIXUP_CHECK=true

### DIFF
--- a/pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
+++ b/pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
@@ -35,7 +35,7 @@ import (
 
 const (
 	pluginName                    = "invalidcommitmsg"
-	invalidCommitMsgCommentMarker = "invalid-commit-message"
+	invalidCommitMsgCommentMarker = "<!-- [prow:invalid-commit-message] -->"
 	invalidCommitMsgLabel         = "do-not-merge/invalid-commit-message"
 )
 
@@ -138,7 +138,10 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
 
 	// unified comment handling
 	cp.PruneComments(func(comment github.IssueComment) bool {
-		return strings.Contains(comment.Body, invalidCommitMsgCommentMarker)
+		return strings.Contains(comment.Body, invalidCommitMsgCommentMarker) ||
+			// TODO: legacy markers, remove after August 2026
+			strings.Contains(comment.Body, "**The list of commits with invalid commit messages**:") ||
+			strings.Contains(comment.Body, "not allowed in the title of a Pull Request")
 	})
 
 	if hasIssues {
@@ -155,19 +158,20 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
 		if checkFixup && len(fixupCommits) != 0 {
 			sections = append(sections,
 				fmt.Sprintf(
-					"### Fixup/amend/squash commits\n\nTemporary commits like fixup!, amend!, or squash! are not allowed.\n\nUse git rebase --autosquash to fix them.\n\n%s",
+					"### Fixup/amend/squash commits\n\nTemporary commits like fixup!, amend!, or squash! are not allowed.\n\nUse [git rebase --autosquash](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash) to fix them.\n\n%s",
 					dco.MarkdownSHAList(org, repo, fixupCommits),
 				))
 		}
 
 		if invalidPRTitle {
 			sections = append(sections,
-				"### Invalid PR title\n\n[Keywords](https://help.github.com/articles/closing-issues-using-keywords) are not allowed in PR titles.\n\nUse /retitle <new-title> to fix it.",
+				"### Invalid PR title\n\n[Keywords](https://help.github.com/articles/closing-issues-using-keywords) are not allowed in PR titles.\n\nUse `/retitle <new-title>` to fix it.",
 			)
 		}
 
 		commentBody := fmt.Sprintf(
-			"[invalid-commit-message]\n\nInvalid commit message issues detected\n\n%s\n\n%s",
+			"%s\n\nInvalid commit message issues detected\n\n%s\n\n%s",
+			invalidCommitMsgCommentMarker,
 			strings.Join(sections, "\n\n"),
 			plugins.AboutThisBot,
 		)
--- a/pkg/plugins/invalidcommitmsg/invalidcommitmsg_test.go
+++ b/pkg/plugins/invalidcommitmsg/invalidcommitmsg_test.go
@@ -292,7 +292,7 @@ func TestHandlePullRequest(t *testing.T) {
 
 				comment := comments[0]
 
-				if !strings.Contains(comment, "[invalid-commit-message]") {
+				if !strings.Contains(comment, invalidCommitMsgCommentMarker) {
 					t.Errorf("Missing marker in comment")
 				}
 

### CODE REVIEW

**Comment by valen-mascarenhas14** [comment]
@petr-muller  @cjwagner @jmguzik @MadhavJivrajani @Priyankasaggu11929 Can you take a look at this PR ?

File: pkg/plugins/invalidcommitmsg/invalidcommitmsg.go [hunk 1]
```diff
@@ -107,11 +127,14 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
 		title  = pr.PullRequest.Title
 	)
 
+	checkFixup := os.Getenv("ENABLE_FIXUP_CHECK") == "true"
```
stmcginnis, line 90: Shouldn't this use a configuration setting rather than an env variable?

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -69,7 +89,7 @@ func init() {
 func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
 	// Only the Description field is specified because this plugin is not triggered with commands and is not configurable.
```
stmcginnis, line 50: I don't think this comment will be accurate anymore?

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: @stmcginnis I had set an env variable due to this
https://github.com/kubernetes-sigs/prow/pull/641#:~:text=I%27d%20suggest%20to%20add%20the%20commit%20check%20itself%20first%20in%20a%20separate%20PR%2C%20and%20make%20it%20disabled%20by%20default.%20As%20a%20temporary%20measure%2C%20it%20could%20be%20enabled%20by%20e.g.%20a%20envvar.

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: I had not modified the comment as this feature is currently behind a feature flag and as discussed here https://github.com/kubernetes-sigs/prow/pull/641#issuecomment-4061240942. I was planning on removing it once the configuration surface for the plugin  is added

valen-mascarenhas14:
@petr-muller  @cjwagner @jmguzik @MadhavJivrajani @Priyankasaggu11929  Did you get a chance to take a look at this PR ?

petr-muller:
Sorry for the delay, I was traveling last week and this week I had to deal with a backlog. I expect to get to reviews here this weekend.

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -35,6 +36,7 @@ import (
 const (
 	pluginName                  = "invalidcommitmsg"
 	invalidCommitMsgLabel       = "do-not-merge/invalid-commit-message"
+	fixupCommitMsgLabel         = "do-not-merge/fixup-commits"
```
petr-muller: I'm not sure about using multiple labels. Labels need maintenance to be present in repos and using multiple labels makes Tide configs more complicated (if we merge this, everyone needs to add a label to their repo and a clause to their Tide config). I think using just the generic previously existing `do-not-merge/invalid-commit-message` label is enough, the actual mean to communicate what is wrong is the comment. 

Ideally the plugin just runs all checks, if at least one check fails, it makes sure the `do-not-merge/invalid-commit-message"` label is present, if all checks pass, it makes sure the label is absent. If at least one check fails, there should be a single comment that says what the author needs to fix. It will lead to a simpler code, even, because we'll need to only do bookkeeping over a single label and a single comment.

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -58,9 +60,27 @@ When GitHub merges a Pull Request, the title is included in the merge commit. To
 </details>
 `
 	invalidTitleCommentPruneBody = "not allowed in the title of a Pull Request"
+	fixupCommitMsgCommentBody    = `Temporary commits like fixup! or amend! are not allowed in the commit history.
```
petr-muller: Is there a good reason to not flag `squash!` as well? https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash says it's a thing too:

> Automatically squash commits with specially formatted messages into previous commits being rebased. If a commit message starts with "squash! ", "fixup! " or "amend! "

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -58,9 +60,27 @@ When GitHub merges a Pull Request, the title is included in the merge commit. To
...
+
+%s
+</details>
+`
+
+	fixupCommitMsgCommentPruneBody = "**The list of fixup/amend commits**:"
 )
 
 var CloseIssueRegex = regexp.MustCompile(`((?i)(clos(?:e[sd]?))|(fix(?:(es|ed)?))|(resolv(?:e[sd]?)))[\s:]+(\w+/\w+)?#(\d+)`)
+var FixupCommitRegex = regexp.MustCompile(`^(fixup!|amend!)`)
```
petr-muller: I wonder if these need to be exported

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -142,6 +174,15 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
 		})
 	}
 
+	if hasFixupCommitMsgLabel && len(fixupCommits) == 0 {
```
petr-muller: nit: all these conditionals (here and also L194 and L214) should be gated by the `checkFixup` boolean. In some cases it is redundant with nonempy fixupCommits but it's better to gate external behavior (posting comments) than internal, just to be sure.

valen-mascarenhas14: Thanks, have made the changes

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Yes. I've modified it, currently it won't be exported

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Yes, have added `squash!` too. Thanks for pointing out

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Makes sense. Thanks, have made the necessary code changes

petr-muller:
@valen-mascarenhas14 I see you consolidated the code to use a single label (good!) but it still posts multiple comments; in https://github.com/kubernetes-sigs/prow/pull/656#discussion_r3035671691 I argued the plugin should manage a single comments with its results. Not sure if you overlooked that part?

valen-mascarenhas14:
@petr-muller sorry, I missed it . I've made the code changes 
Here's an example screenshot of how the comment now looks 
Screenshot From 2026-04-06 23-36-57

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -120,60 +102,80 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
...
+		}
+
+		if invalidPRTitle {
+			sections = append(sections,
+				"### Invalid PR title\n\n[Keywords](https://help.github.com/articles/closing-issues-using-keywords) are not allowed in PR titles.\n\nUse /retitle <new-title> to fix it.",
+			)
+		}
+
+		commentBody := fmt.Sprintf(
+			"[invalid-commit-message]\n\nInvalid commit message issues detected\n\n%s\n\n%s",
```
petr-muller: 1. we should inject the marker into the string with `%s` and use the named variable you added
2. also, it is awkward that it is visible and perhaps a little too generic, I'm a bit worried about accidentally removing someone's comment that mistakenly quotes it or something like:

>> [invalid-commit-message]
>> ...

> hey anyone knows how to resolve this thing

Can we hide it into a html comment maybe, like `<!-- [prow:invalid-commit-message] -->`

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -120,60 +102,80 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
...
+			sections = append(sections,
+				fmt.Sprintf(
+					"### Fixup/amend/squash commits\n\nTemporary commits like fixup!, amend!, or squash! are not allowed.\n\nUse git rebase --autosquash to fix them.\n\n%s",
+					dco.MarkdownSHAList(org, repo, fixupCommits),
+				))
+		}
+
+		if invalidPRTitle {
+			sections = append(sections,
+				"### Invalid PR title\n\n[Keywords](https://help.github.com/articles/closing-issues-using-keywords) are not allowed in PR titles.\n\nUse /retitle <new-title> to fix it.",
```
petr-muller: > `Use /retitle <new-title> to fix it.`

The `<new-title>` is a HTML tag and GH will eat it (you can see it in your screenshot). You'll need to quote it, ideally the full command: `/retitle <new-title>`

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
```diff
@@ -120,60 +102,80 @@ func handle(gc githubClient, log *logrus.Entry, pr github.PullRequestEvent, cp c
...
 
-	// if there are invalid commits, add a comment
-	if len(invalidCommits) != 0 {
-		// prune old comments before adding a new one
-		cp.PruneComments(func(comment github.IssueComment) bool {
-			return strings.Contains(comment.Body, invalidCommitMsgCommentPruneBody)
-		})
+	// unified comment handling
+	cp.PruneComments(func(comment github.IssueComment) bool {
+		return strings.Contains(comment.Body, invalidCommitMsgCommentMarker)
```
petr-muller: note that this is not backward compatible - if

1. someone has a comment on their PR
2. they update prow
3. they resolve their issue

then their old comment does not have a new marker and will be missed.

In this specific case I don't think it is the end of the world (1+2 will be pretty rare and a comment not deleted is only a nuisance), but properly handling "old" state is a common problem so I wanted to point it out.

Given there are other comments we may as well address this, I'd typically do something like this:

```
return strings.Contains(comment.Body, invalidCommitMsgCommentMarker) ||
       // TODO: legacy markers, remove after August 2026
       strings.Contains(comment.Body, "old marker) ||
       strings.Contains(comment.Body, "another old marker)
```

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
petr-muller: yeah I wanted to avoid reviewing & testing new config surface + new functionality, the idea is to merge this in a stealth mode and to remove the envvar with the config surface

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Noted. Have made the code changes

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Thanks for poiting out, have made the changes, this  is how it looks now 
image

pkg/plugins/invalidcommitmsg/invalidcommitmsg.go
valen-mascarenhas14: Yes. have made the changes

PR #657, opened 2026-03-16
jira remove QA Contact and Target Version getters and flags

deepsm007:
This PR removes the QA Contact and Target Version custom field logic from the Jira client. The diff is deletions only relative to the current branch.

## Changes

**pkg/jira/jira.go**
- Remove from `Client` interface: `GetIssueQaContact`, `GetIssueTargetVersion` and their comments.
- Remove package-level `GetIssueQaContact` and `GetIssueTargetVersion` that used hardcoded `customfield_12316243` (QA Contact) and `customfield_12319940` (Target Version).
- Remove the client methods that delegated to those package-level getters.

**pkg/jira/fakejira/fake.go**
- Remove `FakeClient.GetIssueQaContact` and `FakeClient.GetIssueTargetVersion`.

**pkg/flagutil/jira.go**
- Struct field alignment only (no behavioral change).

## Rationale

Those custom field IDs are instance-specific. Removing this API avoids tying Prow to a single Jira instance. Callers that need QA Contact or Target Version should resolve fields by name (e.g. via `GET /rest/api/2/field`) and read from the issue’s field map; downstream can provide helpers and cache IDs.

**Breaking change:** Any caller using `Client.GetIssueQaContact` or `Client.GetIssueTargetVersion` must switch to name-based resolution or a custom helper.

### DIFF
--- a/pkg/jira/jira.go
+++ b/pkg/jira/jira.go
@@ -73,10 +73,6 @@ type Client interface {
 	// is set for the issue, the returned SecurityLevel and error will both be nil and
 	// the issue will follow the default project security level.
 	GetIssueSecurityLevel(*jira.Issue) (*SecurityLevel, error)
-	// GetIssueQaContact get the user details for the QA contact. The QA contact is a custom field in Jira
-	GetIssueQaContact(*jira.Issue) (*jira.User, error)
-	// GetIssueTargetVersion get the issue Target Release. The target release is a custom field in Jira
-	GetIssueTargetVersion(issue *jira.Issue) (*[]*jira.Version, error)
 	// FindUser returns all users with a field matching the queryParam (ex: email, display name, etc.)
 	FindUser(queryParam string) ([]*jira.User, error)
 	GetRemoteLinks(id string) ([]jira.RemoteLink, error)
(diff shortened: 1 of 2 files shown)


pkg/jira/jira.go
```diff
@@ -877,30 +891,52 @@ func (jc *client) GetIssueSecurityLevel(issue *jira.Issue) (*SecurityLevel, erro
...
 	})
 	return obj, err
 }
 
+// GetIssueQaContact returns the QA contact for the issue when called without a configured client.
+// When no field ID is configured, returns (nil, nil). Prefer using Client.GetIssueQaContact with configured field ID.
+func GetIssueQaContact(issue *jira.Issue) (*jira.User, error) {
+	return getIssueQaContactWithFieldID(issue, "")
+}
+
```
bradmwilliams: Since this is a breaking change, why not just drop the functions all together?  The values used in both those functions are specific to Redhat's, now outdated, Jira instance.

deepsm007: i can remove those fields and make a generic Options map which can be used for any custom field and easy to handle for different env

PR #675, opened 2026-04-04
trigger: add hierarchical `trusted_apps` config with global/org/repo levels

petr-muller:
Replace the flat `[]string` TrustedApps field with a structured TrustedApp type with `untrusted` and `locked` fields. Trust is now resolved hierarchically: global -> org -> repo.

- `locked` may be used optionally to deny lower levels from changing trust
- `untrusted` may be used to override trust from layers above or (together  with `locked`) to deny lower levels from enabling trust to the given app

This allows use cases like the following:

- Prow admin may set trust on the global level for a service app to make sure the service works across the whole instance, and set `locked` to  make sure any lower lever does not remove trust and break the service.
- Prow admin or org owner may deny trust to an app on their level and set `locked` to prevent lower layers from trusting
- Prow admin or org owner may set trust on their level for a convenience app so that individual repositories do not need to opt-in individually, but without `locked` lower layers may still opt-out from trusting the app

For backward compatibility of the config, I added custom UnmarshalJSON implementation that accepts both the new list of structs as well as the legacy list of strings. List of strings will be eventually deprecated and removed.

Assisted by Claude Code

### DIFF
--- a/pkg/plugins/config.go
+++ b/pkg/plugins/config.go
@@ -33,6 +33,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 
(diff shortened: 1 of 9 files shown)


petr-muller:
/hold

I still need to polish it and test it but it is roughly ready for CI and review

/cc @jmguzik

pkg/plugins/config.go
```diff
@@ -1007,29 +1061,90 @@ func (c *Configuration) LgtmFor(org, repo string) *Lgtm {
...
+		state[app.Name] = &trustedAppState{
+			trusted: !app.Untrusted,
+			locked:  app.Locked,
+		}
 	}
-	// If you don't find anything, loop again looking for an org config
-	for _, trigger := range c.Triggers {
-		if !sets.NewString(trigger.Repos...).Has(org) {
+	for _, app := range repoApps {
+		if s, exists := state[app.Name]; exists && s.locked {
```
jmguzik, line 1132: `app.Locked` seems to be dropped here intentionally. Explanation or warning would be needed.

pkg/plugins/config.go
```diff
@@ -1007,29 +1061,90 @@ func (c *Configuration) LgtmFor(org, repo string) *Lgtm {
...
 	for _, trigger := range c.Triggers {
-		if !sets.NewString(trigger.Repos...).Has(fullName) {
+		repos := sets.New[string](trigger.Repos...)
+		switch {
+		case len(trigger.Repos) == 0:
+			globalApps = append(globalApps, trigger.TrustedApps...)
+		case repos.Has(fullName):
+			repoApps = append(repoApps, trigger.TrustedApps...)
+			if repoMatch == nil {
+				repoMatch = &trigger
```
jmguzik, line 1080: Is the use of a pointer here intentional (also in the bottom)? In the current design I think we only take one variable.

pkg/plugins/config.go
```diff
@@ -1007,29 +1061,90 @@ func (c *Configuration) LgtmFor(org, repo string) *Lgtm {
...
+			}
+		case repos.Has(org):
+			orgApps = append(orgApps, trigger.TrustedApps...)
+			if orgMatch == nil {
+				orgMatch = &trigger
```
jmguzik, line 1085: Same here.

jmguzik (commented):
I think the change is going in good direction, despite the fact it introduces new fields which may seem to complicate things. I have some doubts about pointer usage in `TriggerFor` (it would need change or explanation, I may not see something). Nice, both formats are supported which is definitely a plus.

PR #677, opened 2026-04-08
Add org invite configuration options

lentzi90:
Add `OrgInviteConfig` and sub-field `ProminentOrgInviteConfig` to allow
customizing the prominent organization invitation message shown to
non-members with multiple merged PRs.

Configuration for the prominent invite includes:
- `disabled`: Disable the prominent message entirely
- `merged_pr_threshold`: Customize the PR count threshold (default: 3)
- `message`: Custom message template with `{join_org_url}` placeholder

Fixes: #670

### DIFF
--- a/pkg/plugins/config.go
+++ b/pkg/plugins/config.go
@@ -479,6 +479,43 @@ type AssignOnLabel struct {
 	Label string `json:"label"`
 }
 
+// ProminentOrgInviteConfig holds configuration for the prominent org invite
+// message shown to non-org members who have contributed multiple merged PRs.
+type ProminentOrgInviteConfig struct {
+	// Disabled disables the prominent org invite functionality entirely.
+	// When true, only the regular "join the org" message is shown without
+	// querying for merged PRs.
+	Disabled bool `json:"disabled,omitempty"`
+	// MergedPRThreshold is the number of merged PRs after which the user gets
+	// a prominent message about joining the org. Default: 3.
+	// Use a pointer so that we can distinguish between "not set" (use default)
+	// and "explicitly set to 0".
+	MergedPRThreshold *int `json:"merged_pr_threshold,omitempty"`
+	// Message is a custom message template for the prominent org invite.
+	// Supports {join_org_url} as a placeholder for the org join URL.
+	// Default: ">[!TIP]\n>**We noticed you've done this a few times! Consider [joining the org]({join_org_url}) ..."
+	Message string `json:"message,omitempty"`
+}
+
+// EffectiveMergedPRThreshold returns the configured merged PR threshold,
+// or the default of 3 when not configured.
+func (c ProminentOrgInviteConfig) EffectiveMergedPRThreshold() int {
+	if c.MergedPRThreshold != nil {
+		return *c.MergedPRThreshold
+	}
+	return 3
+}
+
+// OrgInviteConfig holds configuration for the org invite functionality
+// that is shown to non-org members when they open a PR.
+// Future top-level fields (e.g. disabled, message) may be added here to
+// control the regular (non-prominent) invitation as well.
+type OrgInviteConfig struct {
+	// Prominent configures the prominent org invite message shown to
+	// non-org members who have contributed multiple merged PRs.
+	Prominent ProminentOrgInviteConfig `json:"prominent,omitempty"`
+}
+
 // Trigger specifies a configuration for a single trigger.
 //
 // The configuration for the trigger plugin is defined as a list of these structures.
@@ -507,6 +544,9 @@ type Trigger struct {
 	IgnoreOkToTest bool `json:"ignore_ok_to_test,omitempty"`
 	// TriggerGitHubWorkflows enables workflows run by github to be triggered by prow.
 	TriggerGitHubWorkflows bool `json:"trigger_github_workflows,omitempty"`
+	// OrgInvite holds configuration for the org invite message
+	// shown to non-org members when they open a PR.
+	OrgInvite OrgInviteConfig `json:"org_invite,omitempty"`
 }
 
 // Heart contains the configuration for the heart plugin.
--- a/pkg/plugins/trigger/pull-request.go
+++ b/pkg/plugins/trigger/pull-request.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


lentzi90:
First attempt ready for review. I tried to follow the suggestions in the comments on the issue. Skipped `org_join_guidelines` since there is already `org_join_url`.

> We should make sure the config can be properly layered on the org/repo level, IIRC the TriggerFor exists already so it should be possible to follow the pattern there.

This part I did not quite understand. Or, actually now that I read it again, do you mean that we should put it under `Repos` in the `Trigger` struct? So we could have different guidelines per repo even?

pkg/plugins/trigger/trigger.go
```diff
@@ -97,6 +97,10 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo)
 				JoinOrgURL:     "https://github.com/kubernetes/community/blob/master/community-membership.md",
 				OnlyOrgMembers: true,
 				IgnoreOkToTest: true,
+				OrgInvite: plugins.OrgInviteConfig{
+					MergedPRThreshold: func() *int { v := 5; return &v }(),
```
stmcginnis: We could use `ptr.To(5)` here instead of the closure.

lentzi90: Thanks! I had closures in many more places first, then switched to `ptr` but missed this one.

lentzi90:
This integration test keeps failing in different ways. Doesn't seem related to my changes from what I can tell.
/retest

pkg/plugins/trigger/trigger.go
```diff
@@ -97,6 +98,10 @@ func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo)
 				JoinOrgURL:     "https://github.com/kubernetes/community/blob/master/community-membership.md",
 				OnlyOrgMembers: true,
 				IgnoreOkToTest: true,
+				OrgInvite: plugins.OrgInviteConfig{
```
Prucek: I think this should not be here; we want to keep the defaults.

pkg/plugins/trigger/pull-request_test.go
```diff
@@ -771,10 +773,238 @@ func TestShouldHighlightJoinOrgMessageSkipsBotAuthors(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			result := effectiveMergedPRThreshold(tc.cfg)
+			if result != tc.expected {
+				t.Fatalf("effectiveMergedPRThreshold() = %d, want %d", result, tc.expected)
+			}
+		})
+	}
+}
```
Prucek: I think these tests are not necessary, as all this is tested in `TestShouldHighlightJoinOrgMessageCustomThreshold`

lentzi90: True. These were just testing the internal code which is anyway tested through `TestShouldHighlightJoinOrgMessageCustomThreshold`. Removed.

Prucek (approved):

pkg/plugins/trigger/pull-request_test.go
```diff
@@ -771,10 +773,197 @@ func TestShouldHighlightJoinOrgMessageSkipsBotAuthors(t *testing.T) {
...
+			PullRequest: &struct{}{},
+		}
+	}
+
+	testCases := []struct {
+		name             string
+		cfg              plugins.OrgInviteConfig
+		issues           map[int]*github.Issue
+		expectedContains string
+		expectNoSearch   bool
```
droslean: I don't think this is appropriate for the tests. You should be able to just compare the expected result. So if the fakeclient needs to be initialized with existing issues, those need to be encapsulated inside the specific test case.

pkg/plugins/trigger/pull-request.go
```diff
@@ -43,6 +44,15 @@ const (
 	mergedPRCountForProminentJoinOrgMessage = 3
 )
 
+// effectiveMergedPRThreshold returns the configured merged PR threshold,
+// or the default (mergedPRCountForProminentJoinOrgMessage) when not configured.
+func effectiveMergedPRThreshold(cfg plugins.OrgInviteConfig) int {
```
droslean: This could just be an OrgInviteConfig method instead.

lentzi90: True, that would be cleaner. Updated

pkg/plugins/trigger/pull-request_test.go
lentzi90: Right, I can understand this. I was thinking that there might be concern for API calls. Anyway, I updated this now. Much simpler this way at least :slightly_smiling_face:

pkg/plugins/config.go
```diff
@@ -507,6 +534,9 @@ type Trigger struct {
 	IgnoreOkToTest bool `json:"ignore_ok_to_test,omitempty"`
 	// TriggerGitHubWorkflows enables workflows run by github to be triggered by prow.
 	TriggerGitHubWorkflows bool `json:"trigger_github_workflows,omitempty"`
+	// OrgInvite holds configuration for the prominent org invite message
+	// shown to non-org members who have multiple merged PRs.
+	OrgInvite OrgInviteConfig `json:"org_invite,omitempty"`
```
petr-muller: I don't like that an `org_invite` stanza only configures a *prominent* invite but not the normal one. It is documented here as such, but the actual configs will not carry the doc and will lead to confusing configuration stanzas, like:

```
org_invite:
  disabled: true
```
^ still invites people, just less enthusiastically. I would expect to not do any invitations.

```
org_invite:
  message: "sorry we do not invite anyone"
``` 
^ still invites people with a "`Regular contributors should [join the org](%s) to skip this step`" message

I don't think we want configurability only for the prominent case (having only `prominent_org_invite` looks weird).

Could we structure the config like this:

```yaml
org_invite:
  prominent:
    disabled: true
    merged_pr_threshold: 3
    message: ...
```

That helps understanding that the fields have limited scope (may not carry entire context, but at least are not entirely misleading), and it allows someone to add a configuration layer above, like:

```yaml
org_invite:
  disabled: true # do not show any invitations
  message: "'..." # tweak the default invittation
  prominent:
    ... 
```

petr-muller (changes requested):
There's just one structural item I'd like to see addressed, I'd like a layer between `org_invite` and the config fields that only affect the prominent invitation, to avoid confusing config instances like:

```yaml
org_invite:
  disabled: true
```

that results in still inviting people by default. Happy to merge this change afterwards!

pkg/plugins/config.go
petr-muller: to be clear, I'm not asking for the full top-layer configurability, just to stuff the prominent one into a substruct, and the top-layer config code may be added by someone in the future

lentzi90: This makes sense. Thanks for explaining it so thoroughly. I'll try to adjust the PR :slightly_smiling_face:

petr-muller (approved):
Thank you!

PR #683, opened 2026-04-15
Fix expanding skipped lines when using S3

timebertt:
In contrast to the GCS range reader, the S3 range reader returns `io.EOF` at the end of the desired range. This is compliant with the `io.Reader` specification. However, the `StorageArtifact.ReadAt` function expects `io.EOF` only when reading up to the overall end of the artifact.

This leads to problems when expanding skipped lines in an S3-based spyglass build log. Expanding the skipped lines at the end of the build log works, but expanding lines at the start or in the middle results in:
```
Failed to retrieve log lines: couldn't read requested bytes: error reading from artifact: EOF
```

/kind bug


timebertt:
ping @michelle192837 @smg247 for approval

matthyx (approved):
looks good @timebertt !

PR #689, opened 2026-04-26
[WIP] fixcommitmsg: add plugin to rewrite invalid commit messages

dims:
Introduces the fixcommitmsg plugin, which exposes a /fix-commit-messages comment command. When triggered by a member of the configured maintainer team on an open pull request, the plugin:

  - Replaces close-issue keywords (closes/fixes/resolves #NNN) in commit messages with "Related: NNN", removing the "#" that GitHub resolves into a hyperlink and an automatic issue close action.
  - Strips fixup!/amend!/squash! prefixes from commit subject lines.
  - Force-pushes the rewritten history back to the PR branch.

If the push fails (e.g., the author has not enabled "Allow edits from maintainers" on a fork-based PR), the plugin posts the error as a comment and takes no further action.

Also exports FixupCommitRegex from the invalidcommitmsg package so the new plugin can reuse the same pattern without duplicating it, and adds CherryPick and Amend to the git/v2 Interactor interface so the plugin uses the standard git client abstraction rather than shelling out directly.

### DIFF
--- a/pkg/plugins/fixcommitmsg/fixcommitmsg.go
+++ b/pkg/plugins/fixcommitmsg/fixcommitmsg.go
@@ -1,0 +1,282 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package fixcommitmsg provides the /fix-commit-messages command. When
+// triggered, it rewrites commit messages on a pull request to remove
+// close-issue keywords and fixup/amend/squash prefixes that are flagged
+// as invalid by the invalidcommitmsg plugin. Only members of the configured
+// maintainer team can trigger this command.
+package fixcommitmsg
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"sigs.k8s.io/prow/pkg/config"
+	git "sigs.k8s.io/prow/pkg/git/v2"
+	"sigs.k8s.io/prow/pkg/github"
+	"sigs.k8s.io/prow/pkg/pluginhelp"
+	"sigs.k8s.io/prow/pkg/plugins"
+	"sigs.k8s.io/prow/pkg/plugins/invalidcommitmsg"
+)
+
+const (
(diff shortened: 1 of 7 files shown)


dims:
Under the hood, the plugin runs the equivalent of:

```bash
git checkout pr-branch
git rev-parse <oldest-pr-commit>^   # find the base

git checkout <base-sha>
git checkout -b fix-commit-messages-temp

# for each commit on the PR, cherry-pick and amend if needed
git cherry-pick <sha>
git commit --amend -m "fixed message"   # only for flagged commits

git checkout pr-branch
git reset --hard <new-tip>
git push --force origin pr-branch
```

dims:
For fork-based PRs, `ClientFor` is called with the PR head's org/repo (the fork), so the clone and push target the fork rather than the upstream. `PushToCentral` then pushes back to the fork's remote. This works as long as the PR author has enabled "Allow edits from maintainers" on their PR — if they haven't, the push fails and the plugin posts a comment explaining what they need to do.

pkg/plugins/fixcommitmsg/fixcommitmsg.go
```diff
@@ -0,0 +1,282 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
stmcginnis, line 2: 2026? Though with current guidance the year can be dropped. I can't remember now if the check for this was updated in this repo yet.

dims: i keep forgetting!

pkg/plugins/fixcommitmsg/fixcommitmsg.go
```diff
@@ -0,0 +1,282 @@
...
+	if body != "" {
+		fixed += "\n" + body
+	}
+	return fixCloseKeywords(fixed)
+}
+
+// fixSubject removes a leading fixup!/amend!/squash! prefix from the commit
+// subject line, leaving the remainder trimmed.
+func fixSubject(subject string) string {
+	return strings.TrimSpace(invalidcommitmsg.FixupCommitRegex.ReplaceAllString(subject, ""))
```
palnabarun, line 165: Is it possible to actually resolve the fixup! or amend! or squash! ? If we can do that programmatically, that would be best. I have had my PRs merged too early when I had fixup! commits that I wanted to autosquash before merge. This would be really helpful.

pkg/plugins/fixcommitmsg/fixcommitmsg_test.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	if len(gc.comments) != 1 {
+		t.Fatalf("expected 1 comment, got %d", len(gc.comments))
+	}
+	if !containsStr(gc.comments[0], "no commit messages need fixing") {
+		t.Errorf("unexpected comment: %s", gc.comments[0])
+	}
+}
+
+func containsStr(s, sub string) bool {
```
stmcginnis, line 233: Can we just use `strings.Contains()`?

PR #694, opened 2026-04-28
spyglass/junit: add configurable test grouping support 🤖

sosiouxme:
Adds support for grouping tests based on XPath property selectors. e.g.

```
        - lens:
            name: junit
            config:
              groups:
              - name: "Informing"
                selector: "properties/property[@name='lifecycle' and @value='informing']"
                collapsed: true
```

Groups are displayed after default test sections with the name as header.

image

🤖 Assisted by Claude Code


pkg/spyglass/lenses/junit/template.html
```diff
@@ -178,6 +178,180 @@
...
+	   {{else}}
+	   <td class="mdl-data-table__cell--non-numeric test-name">{{$firstTest.ClassName}}: {{$firstTest.Name}}</br> </br><b>Reason:</b> {{$firstTest.SkippedReason}}</td>
+	   {{end}}
+          <td class="mdl-data-table__cell--non-numeric">{{$firstTest.Duration}}</td>
+        </tr>
+      {{end}}
+    </tbody>
+  {{end}}
+  {{end}}
+  {{end}}
```
petr-muller, line 354: I would not notice myself but my LLM tells me there's a lot of duplicate content in the template and we could dedup with sub-templates - it's a nontrivial change so I'll probably submit that as a followup rather than asking you to include in this PR.

pkg/spyglass/lenses/junit/lens.go
```diff
@@ -137,9 +269,99 @@ func (lens Lens) Body(artifacts []api.Artifact, resourceDir string, data string,
...
+		}
+	}
+	newTotal := len(*b.passed) + len(*b.failed) + len(*b.flaky) + len(*b.skipped)
+	*b.numTests += (newTotal - prevTotal) - duplicates
+}
+
+func (lens Lens) getJvd(artifacts []api.Artifact, groups ...[]GroupConfig) JVD {
+	var groupConfigs []GroupConfig
+	if len(groups) > 0 {
+		groupConfigs = groups[0]
```
petr-muller: I'm not sure I get why is `groups` variadic, why don't we simply pass whatever `groups[0]` is? At worst it's gonna be an empty or nil []GroupConfig slice...

petr-muller (commented):
LGTM but I have a question about new `getJvd` param, see inline

I'll probably followup with a HTML template cleanup. 

I really appreciate updating the docs, thank you!

pkg/spyglass/lenses/junit/lens.go
sosiouxme: granted it's weird, that's two reviews that pointed it out...can address

petr-muller (approved):
Thank you!

PR #695, opened 2026-04-29
Introduce new markdown lens.

vovakrk:
Summary markdown lens is intended for rendering markdown (MD) files.


pkg/spyglass/lenses/summary/summary.go
```diff
@@ -0,0 +1,107 @@
...
+
+	var documents []document
+	for _, artifact := range artifacts {
+		content, err := artifact.ReadAll()
+		if err != nil {
+			logrus.WithError(err).WithField("artifact_url", artifact.CanonicalLink()).Warn("failed to read content")
+			continue
+		}
+		
+		htmlContent := blackfriday.Run(content)
```
michelle192837: note: I believe this is a case where we'd want to consider sanitizing the input, but `executeTemplate()` later should take care of that, so this _should_ be fine.

michelle192837 (approved):
Thanks! Looks good to me for implementation (though I'd feel better with another maintainer's approval, @petr-muller or another if you'd like to take a look).

/lgtm

/hold

(in case Petr wants to take a look)

vovakrk:
Hi @petr-muller, Could you take a look at my PR? Thank you!

pkg/spyglass/lenses/summary/summary.go
petr-muller: I think it's not fine, the output is fed to `template.HTML` below which is [documented to assume the input to be safe](https://pkg.go.dev/html/template#HTML):

> HTML encapsulates a known safe HTML document fragment. It should not be used for HTML from a third-party

I guess we could use it this way if the markdown processor is trusted to emit safe HTML from potentially unsafe markdown, but the blackfriday one [does not seem to do that at least by default](https://github.com/russross/blackfriday/tree/master#sanitize-untrusted-content):

> Blackfriday itself does nothing to protect against malicious content. If you are dealing with user-supplied markdown, we recommend running Blackfriday's output through HTML sanitizer such as [Bluemonday](https://github.com/microcosm-cc/bluemonday).

pkg/spyglass/lenses/markdown/markdown.go
```diff
@@ -0,0 +1,107 @@
...
+
+		documents = append(documents, document{
+			Title:   filepath.Base(artifact.CanonicalLink()),
+			Content: template.HTML(htmlContent),
+		})
+	}
+
+	t, err := template.ParseFiles(filepath.Join(resourceDir, "template.html"))
+	if err != nil {
+		logrus.WithError(err).Error("Error executing template.")
```
petr-muller:
```suggestion
		logrus.WithError(err).Error("Failed to load template.")
```

petr-muller (commented):
Thanks for the contribution! I thought about having a Markdown lens in the past, I believe it can be very useful.

Some high-level notes:

1. Should it just be called `markdown` lens instead of "summary"?
2. I'm no expert on Markdown Go libraries, but https://github.com/russross/blackfriday seems essentially abandoned; is there a reason why you chose it? Quick google search showed https://github.com/yuin/goldmark as an actively developed candidate but I really did not do any deeper research.
3. The logic is fairly simple but I still think we'd benefit from having basic tests. It makes it easier for folks to extend the behavior in the future.

petr-muller (changes requested):

pkg/spyglass/lenses/summary/summary.go
petr-muller: FWIW the goldmark processor [claims to be safe](https://github.com/yuin/goldmark#html-renderer-options):

> By default, goldmark does not render raw HTML or potentially dangerous links

michelle192837: Goldmark seems reasonable 👍 

I think you're right that template.HTML isn't safe, but I think [ExecuteTemplate()](https://pkg.go.dev/html/template#hdr-Introduction) is? (Though it is parsing arbitrary content, so the "template author is trusted" assumption isn't quite there).

petr-muller: :sweat_smile: I'm just roleplaying a web developer here, it _is_ possible that I'm wrong :D

pkg/spyglass/lenses/markdown/markdown.go
vovakrk: done

vovakrk:
1. Renamed the lens to markdown
2. Migrated to goldmark
3. Added unit tests

michelle192837 (approved):
Looks good to me!

/hold

Holding in case Petr or someone else wants to comment, otherwise hold-cancel at will.

petr-muller (approved):
Thanks!

PR #702, opened 2026-05-02
plugins: move transfer-issue to issue management

Amulyam24:
With the introduction of issue-management, move transfer-plugin to the folder.

Signed-off-by: Amulyam24 <amulmek1@in.ibm.com>
Assisted-by: AI

### DIFF
--- a/pkg/plugins/issue-management/issue_management.go
+++ b/pkg/plugins/issue-management/issue_management.go
@@ -39,6 +39,7 @@ var (
 	unlinkIssueRegex = regexp.MustCompile(`(?mi)^/unlink-issue\s+(.+)$`)
 	pinRegex         = regexp.MustCompile(`(?mi)^/pin-issue\s*$`)
 	unpinRegex       = regexp.MustCompile(`(?mi)^/unpin-issue\s*$`)
+	transferRegex    = regexp.MustCompile(`(?mi)^/transfer(?:-issue)?(?: +(.*))?$`)
 )
 
 type githubClient interface {
@@ -83,6 +84,13 @@ func helpProvider(_ *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.Plu
 		WhoCanUse:   "Approvers from the top-level OWNERS file",
 		Examples:    []string{"/unpin-issue"},
 	})
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/transfer[-issue] <destination repo in same org>",
+		Description: "Transfers an issue to a different repo in the same org.",
+		Featured:    true,
+		WhoCanUse:   "Org members.",
+		Examples:    []string{"/transfer-issue kubectl", "/transfer test-infra"},
+	})
 	return pluginHelp, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/plugins/issue-management/issue_management.go
```diff
@@ -100,11 +108,13 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error
 }
 
 func handleIssues(gc githubClient, oc ownersClient, log *logrus.Entry, e github.GenericCommentEvent) error {
+	// Handle link and unlink issues
 	toLink, toUnlink := parseCommentForLinkCommands(e.Body)
 	if len(toLink) > 0 || len(toUnlink) > 0 {
 		return handleLinkIssue(gc, log, e, toLink, toUnlink)
 	}
 
+	// Handle pin and unpin issues
```
Prucek: These comments are IMO unnecessary, the code is clear/clean enough.

pkg/plugins/issue-management/issue_management.go
```diff
@@ -113,6 +123,17 @@ func handleIssues(gc githubClient, oc ownersClient, log *logrus.Entry, e github.
...
+	if matches := transferRegex.FindAllStringSubmatch(e.Body, -1); len(matches) > 0 {
+		if len(matches) != 1 || len(matches[0]) != 2 || len(matches[0][1]) == 0 {
+			return gc.CreateComment(
+				e.Repo.Owner.Login, e.Repo.Name, e.Number,
+				plugins.FormatResponseRaw(e.Body, e.HTMLURL, e.User.Login, "/transfer-issue must only be used once and with a single destination repo."),
+			)
+		}
+		return handleTransfer(gc, log, e, matches[0][1])
+	}
+
```
Prucek: This should still have the cleanness of the function above. I'd put the whole logic inside the `handleTransferIssue()` function.

Amulyam24: Makes sense @Prucek , addressed it, PTAL!

Prucek (approved):

petr-muller:
@Amulyam24 this is good, but do we have an idea about how the transition will look like for instances that enable the `transfer-issue` plugin today:
```
plugins:
- transfer-issue
```

Will these blow up with errors on update, or simply lose functionality? I'd like to make sure we do not break instances even if it takes more effort.

And if we're not breaking them but they "only" lose functionality until they enable `issue-management` plugin it makes me think we should get into a habit of updating https://github.com/kubernetes-sigs/prow/blob/main/site/content/en/docs/announcements.md again. Apparently we stopped doing that which is not good. It is not the greatest comm channel of all but we dont have anything better :sweat_smile:

Amulyam24:
> @Amulyam24 this is good, but do we have an idea about how the transition will look like for instances that enable the `transfer-issue` plugin today:
> 
> ```
> plugins:
> - transfer-issue
> ```
> 
> Will these blow up with errors on update, or simply lose functionality? I'd like to make sure we do not break instances even if it takes more effort.

Hi @petr-muller, Thanks! This is a good point which I missed, I will try this out and get back.
Then we can take the next step accordingly.

Amulyam24:
Hi @petr-muller, I confirmed that with this change, using `/transfer-issue` would silently fail.
The `issue-management` plugin should be added to the `plugins.yaml`.

I can update the announcements page, should it be done post this PR merge?

Also, since the page has become kind of obsolete and this change could be breaking, we might have to make the concerned parties explicitly aware, IMO. Atleast Kubernetes and SIGs as a first step, wdyt?

Amulyam24:
Hi @petr-muller, can you please share your input on the above?
Thanks.

Amulyam24:
Hi @petr-muller, did you get a chance to look into the above? If you could please share your input, that'd be great, Thanks!

Amulyam24:
Hi @petr-muller, can you please share your thoughts on the above? We are close to merging this PR. We can take the necessary action based on your input, Thanks!

petr-muller:
@Amulyam24 

> I can update the announcements page, should it be done post this PR merge?

yeah

> Also, since the page has become kind of obsolete and this change could be breaking

I hope to get that page unobsoleted now when it's fairly easy to tell an LLM to update it ;)

>  we might have to make the concerned parties explicitly aware, IMO. Atleast Kubernetes and SIGs as a first step, wdyt?

yeah, posting a message to `#prow` will not hurt. k8s instance is open so I think for k8s its best to just open them a prow config PR, then the change is transparent

Amulyam24:
Hi @petr-muller, the relevant PRs have been created and this should be ready for final review, PTAL!

Announcement - https://github.com/kubernetes-sigs/prow/pull/825
Migration in test-infra - https://github.com/kubernetes/test-infra/pull/37650

Will announce in #prow channel as well post the merge of announcement PR.

petr-muller (approved):

petr-muller:
Thank you!

PR #706, opened 2026-05-04
Add custom KIND_CONFIG support to make dev

Daniel-Giszpenc:
Issue: Unable to develop locally with KinD cluster created by `make dev` deployed and updated via tilt remotely.
Discussed here: https://kubernetes.slack.com/archives/CDECRSC5U/p1777856158346109

Solution:
Edit `setup-kind-cluster.sh` script to accept an argument for a custom KIND_CONFIG used for `kind cluster create config=`
This argument by default is set to default which will continue the existing behavior. Edit the Prow `Makefile` at the `make dev` and `make dev-full` entrypoints will check if the KINK_CONFIG env var is not empty and if so pass it as an parameter for the `dev-env.sh` script. Edit the `dev-env.sh` script to pass the param to the setup script and update the logs.

I also made added a doc in the docs to show the process for setting up a two machine development environment how this PR enables.

### DIFF
--- a/hack/dev-env.sh
+++ b/hack/dev-env.sh
@@ -115,10 +115,15 @@ function check_prerequisites() {
   fi
 }
 
+#TODO: update for case where custom KIND_CONFIG is used
 function print_connection_info() {
   cat <<EOF
 
 ==> Local Prow environment is ready!
+
+Custom KIND_CONFIG Note:
+If a separate APISeverAddress was used then replace localhost below with that.
+The DEV_HTTP_PORT is the host_port used with the container 80 extra port mapping.
 
 Cluster: ${_KIND_CLUSTER_NAME}
 Kubeconfig context: ${_KIND_CONTEXT}
@@ -160,6 +165,7 @@ function main() {
   local rebuild_images=""
   local teardown=false
   local profile="core"
+  local kind_config=""
 
   for arg in "$@"; do
     case "${arg}" in
@@ -175,6 +181,9 @@ function main() {
           echo >&2 "invalid -profile value '${profile}': must be 'core' or 'full'"
           return 1
         fi
+        ;;
+      -kind-config=*)
+        kind_config="${arg#-kind-config=}"
         ;;
       -teardown)
         teardown=true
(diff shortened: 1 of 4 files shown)


test/integration/setup-kind-cluster.sh
```diff
@@ -63,9 +67,11 @@ function main() {
   local fakepubsub_node_port
   local http_host_port
   local https_host_port
+  local kind_config
   fakepubsub_node_port=30303
   http_host_port=80
   https_host_port=443
+  kind_config=default
```
Prucek: It should not contain the string "default" as the sentinel for "no custom config". Consider using an empty string instead

hack/dev-env.sh
```diff
@@ -115,11 +115,16 @@ function check_prerequisites() {
   fi
 }
 
+#TODO: update for case where custom KIND_CONFIG is used
```
Prucek: This is probably not supposed to be here anymore?

hack/dev-env.sh
```diff
@@ -213,10 +222,15 @@ function main() {
...
-    "-https-host-port=${DEV_HTTPS_PORT}"
+  if [[ -n "${kind_config}" ]]; then
+    log "Setting up kind cluster (Custom KIND_CONFIG: ${kind_config})"
+    "${INTEGRATION_DIR}/setup-kind-cluster.sh" "-kind-config=${kind_config}"
+  else
+    log "Setting up kind cluster (HTTP on :${DEV_HTTP_PORT}, HTTPS on :${DEV_HTTPS_PORT})"
+    "${INTEGRATION_DIR}/setup-kind-cluster.sh" \
+      "-http-host-port=${DEV_HTTP_PORT}" \
+      "-https-host-port=${DEV_HTTPS_PORT}"
+  fi
```
Prucek, line 234: We could maybe warn the user that the env vars (DEV_HTTP_PORT and DEV_HTTPS_PORT) are not taken into account when kind_config is set

Daniel-Giszpenc:
Addressed all your comments @Prucek , thanks for the review.
Also you were right, I totally forgot about that TODO (covers face).

Prucek (approved):
/ok-to-test

petr-muller (approved):
Thanks!

PR #708, opened 2026-05-05
external-plugins: add netlify-preview plugin to retry deploy previews

### LINKED ISSUE #693: Integrate with Netlify for /retest Prow command
originally raised here: https://github.com/kubernetes/test-infra/issues/35103

What would you like to be added:
If there is a Netlify preview for [k/website](https://github.com/kubernetes/website) or [k/contributor-site](https://github.com/kubernetes/contributor-site), and an org member uses the /retest command, trigger Netlify to rebuild it.

Why is this needed:
There is currently no nice way for a contributor to self-service request a new Netlify test if there has been a flake or similar failure. The author (and only the author) can push an amended commit, but this is a lot of technical work for a simple step.

Caesarsage:
Adds a new external Prow plugin, `netlify-preview`, that retries the
latest Netlify deploy preview for a pull request in response to chat
commands. Intended for repositories whose PR previews are built by
Netlify (initially `kubernetes/website` and `kubernetes/contributor-site`).

## Design

Two commands (sign-off in #693):

- `/retest` — retries the latest Netlify deploy preview **only when
  that preview is in `error` state**. If it is `ready`, the plugin
  posts a comment pointing the user to `/rebuild-preview` instead.
  Avoids wasteful rebuilds of passing previews on routine `/retest`.
- `/rebuild-preview` — forces a retry regardless of the preview's
  current state, **except** when a build is already running
  (`building` / `enqueued`). The in-flight guard prevents maintainers
  from triggering redundant builds while one is in progress.

## Per-repo site mapping

This PR adds a small configuration package that maps each repository
to a Netlify site ID (This is assuming one Netlify for all related netlify projects, hence a single PAT):

```yaml
repos:
  kubernetes/website:
    site_id: <netlify-site-id>
  kubernetes/contributor-site:
    site_id: <netlify-site-id>
```

Each repo can map to a different Netlify site independently. The
external_plugins: block in Prow configuration is only the enablement
mechanism; site mapping never goes there.

## Trust gating

Retries are gated by the same trust rules Prow's trigger plugin
applies (org membership, configured trusted apps and orgs, or
/ok-to-test on the PR). The gate is a build-minute / abuse
protection rather than a code-execution protection, Netlify has
already built the commit once, so a retry does not grant new code
execution. Wasted Netlify build minutes are the real risk.

Assisted by Claude 

Closes kubernetes-sigs/prow#693.

### DIFF
--- a/cmd/external-plugins/netlify-preview/config/config.go
+++ b/cmd/external-plugins/netlify-preview/config/config.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/cmd/external-plugins/netlify-preview/config/config.go
+++ b/cmd/external-plugins/netlify-preview/config/config.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"os"
+
+	"sigs.k8s.io/yaml"
+)
+
+type Config struct {
+	Repos map[string]Repo `json:"repos,omitempty"`
+}
+
+type Repo struct {
+	SiteID string `json:"site_id,omitempty"`
+}
+
+func Load(path string) (*Config, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	var cfg Config
+	if err := yaml.UnmarshalStrict(data, &cfg); err != nil {
+		return nil, err
+	}
+	if err := cfg.Validate(); err != nil {
+		return nil, err
+	}
+	return &cfg, nil
+}
+
+func (c *Config) Validate() error {
+	for repo, cfg := range c.Repos {
+		if cfg.SiteID == "" {
+			return fmt.Errorf("missing site_id for repo %q", repo)
+		}
+	}
+	return nil
+}
+
+func (c *Config) Repo(org, repo string) (Repo, bool) {
+	if c == nil {
+		return Repo{}, false
+	}
+	cfg, ok := c.Repos[org+"/"+repo]
+	return cfg, ok
+}
--- a/cmd/external-plugins/netlify-preview/config/config.go
+++ b/cmd/external-plugins/netlify-preview/config/config.go
@@ -24,10 +24,10 @@ import (
 )
 
 type Config struct {
-	Repos map[string]Repo `json:"repos,omitempty"`
+	Repos map[string]SiteConfig `json:"repos,omitempty"`
 }
 
-type Repo struct {
+type SiteConfig struct {
 	SiteID string `json:"site_id,omitempty"`
 }
 
@@ -55,9 +55,9 @@ func (c *Config) Validate() error {
 	return nil
 }
 
-func (c *Config) Repo(org, repo string) (Repo, bool) {
+func (c *Config) Repo(org, repo string) (SiteConfig, bool) {
 	if c == nil {
-		return Repo{}, false
+		return SiteConfig{}, false
 	}
 	cfg, ok := c.Repos[org+"/"+repo]
 	return cfg, ok
--- a/cmd/external-plugins/netlify-preview/main.go
+++ b/cmd/external-plugins/netlify-preview/main.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/cmd/external-plugins/netlify-preview/main.go
+++ b/cmd/external-plugins/netlify-preview/main.go
@@ -1,0 +1,147 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// netlify-preview retries Netlify deploy previews for pull requests.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	previewconfig "sigs.k8s.io/prow/cmd/external-plugins/netlify-preview/config"
+	"sigs.k8s.io/prow/cmd/external-plugins/netlify-preview/netlify"
+	netlifypreview "sigs.k8s.io/prow/cmd/external-plugins/netlify-preview/plugin"
+	"sigs.k8s.io/prow/pkg/config/secret"
+	"sigs.k8s.io/prow/pkg/flagutil"
+	prowflagutil "sigs.k8s.io/prow/pkg/flagutil"
+	pluginsflagutil "sigs.k8s.io/prow/pkg/flagutil/plugins"
+	"sigs.k8s.io/prow/pkg/interrupts"
+	"sigs.k8s.io/prow/pkg/logrusutil"
+	"sigs.k8s.io/prow/pkg/pjutil"
+	"sigs.k8s.io/prow/pkg/pluginhelp/externalplugins"
+)
+
+type options struct {
+	port int
+
+	pluginsConfig          pluginsflagutil.PluginOptions
+	dryRun                 bool
+	github                 prowflagutil.GitHubOptions
+	instrumentationOptions prowflagutil.InstrumentationOptions
+	logLevel               string
+
+	webhookSecretFile string
+	netlifyTokenFile  string
+	configPath        string
+	netlifyAPIURL     string
+}
+
+func (o *options) Validate() error {
+	for idx, group := range []flagutil.OptionGroup{&o.github} {
+		if err := group.Validate(o.dryRun); err != nil {
+			return fmt.Errorf("%d: %w", idx, err)
+		}
+	}
+	if o.netlifyTokenFile == "" {
+		return fmt.Errorf("--netlify-token-file is required")
+	}
+	if o.configPath == "" {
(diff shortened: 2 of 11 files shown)


Caesarsage:
I also tested locally against the real Netlify API and the real GitHub API,
with `phony` delivering `issue_comment` webhook events:
https://github.com/Caesarsage/website/pull/5

Decision branches verified end-to-end:

- `/retest` on a `ready` preview: refused, points to `/rebuild-preview`
- `/rebuild-preview` on a `ready` preview: retried, comment posted
- `/rebuild-preview` while a build is in progress: refused (no retry call made)
- `/rebuild-preview` for a repo with no site mapping: refused with config-error message
- Comment from the bot user: event skipped silently

Caesarsage:
cc @petr-muller @smg247

cmd/external-plugins/netlify-preview/config/config.go
```diff
@@ -0,0 +1,64 @@
+/*
+Copyright The Kubernetes Authors.
```
singh1203, line 2:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

cmd/external-plugins/netlify-preview/config/config.go
```diff
@@ -0,0 +1,64 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+type Config struct {
+	Repos map[string]Repo `json:"repos,omitempty"`
+}
+
+type Repo struct {
+	SiteID string `json:"site_id,omitempty"`
+}
```
singh1203, line 32: Nit: type name `Config` & `Repo` is a bit misleading wrapper.
maybe:

```suggestion
type SiteConfig struct {
    SiteID string `json:"site_id,omitempty"`
}
type Config struct {
    Repos map[string]SiteConfig `json:"repos,omitempty"`
}
```

lmktfy:
I agree with the intent of this PR.

cmd/external-plugins/netlify-preview/config/config.go
Caesarsage: Config is the configuration wrapper to the repository and each repository mapped to their netlify site id. 

such as:

```
repos:
  kubernetes/website:
    site_id: <netlify-site-id>
  kubernetes/contributor-site:
    site_id: <netlify-site-id>
```

cmd/external-plugins/netlify-preview/config/config.go
Caesarsage: reverting this, adding Years is no longer supported

Caesarsage:
cc @stmcginnis 

Could you help take a look

cmd/external-plugins/netlify-preview/main.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	pluginAgent, err := o.pluginsConfig.PluginAgent()
+	if err != nil {
+		log.WithError(err).Fatal("Error loading plugin config.")
+	}
+	previewConfig, err := previewconfig.Load(o.configPath)
+	if err != nil {
+		log.WithError(err).Fatal("Error loading netlify-preview config.")
+	}
```
petr-muller, line 122: Not necessary to address right away, but deployments will eventually run into state where someone changes the config but the change will only became effective after the plugin restarts. 

We'll eventually want to add something like the `PluginAgent` above that's able to reload the config continuously.

cmd/external-plugins/netlify-preview/server.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if botUserChecker(commentAuthor) {
+		l.Debug("Comment is made by the bot, skipping.")
+		return nil
+	}
+
+	if trusted, err := s.trustedForCommand(org, repo, number, ic.Issue.User.Login, commentAuthor); err != nil {
+		return err
+	} else if !trusted {
+		return s.comment(ic, "Cannot retry the Netlify deploy preview until a trusted user reviews the PR and leaves an `/ok-to-test` message.")
```
petr-muller: Deny comment may be in order for the explicit comment (`/rebuild-preview`) but not when we react to `/retest`, in that case the comment just increases noise.

cmd/external-plugins/netlify-preview/server.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	repoConfig, ok := s.previewConfig.Repo(org, repo)
+	if !ok {
+		l.WithField("action", "config_error").Info("Repository has no Netlify preview site mapping.")
+		return s.comment(ic, "This repository does not have a Netlify preview site mapping configured, so I can't retry a deploy preview for it.")
```
petr-muller: same here, comment only appropriate when we handle `/rebuild-preview` but not `/retest`

cmd/external-plugins/netlify-preview/server.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	deploys, err := s.netlifyClient.ListDeploys(ctx, repoConfig.SiteID)
```
petr-muller: is the underlying API paginated? I wonder what happens in repos with many deploys, do we always get all of them?

cmd/external-plugins/netlify-preview/server.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	if decision.ShouldRetry {
+		if !s.dryRun {
+			if err := s.netlifyClient.RetryDeploy(ctx, preview.ID); err != nil {
+				return err
+			}
+		}
+		l.Info("Requested Netlify deploy preview retry.")
+		return s.comment(ic, fmt.Sprintf("Retrying the latest Netlify deploy preview for this PR: %s", preview.DeploySSLURL))
```
petr-muller: I'm not sure if this needs a comment, maybe it is useful because it at least give the URL? Not sure. But personally I think the pending GH status context is probably enough (just like `/test` does not post a "jobs were triggered" comment but we rely on GH UI).

cmd/external-plugins/netlify-preview/server.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	l.Info("Did not request Netlify deploy preview retry.")
```
petr-muller: this is borderline, could be noisy for `/retest` events where we do not expect to do anything most of the time so `Debug` would be more appropriate

cmd/external-plugins/netlify-preview/plugin/plugin.go
```diff
@@ -0,0 +1,117 @@
...
+	"sigs.k8s.io/prow/pkg/pluginhelp"
+)
+
+const PluginName = "netlify-preview"
+
+type Command string
+
+const (
+	RetestCommand         Command = "retest"
+	RebuildPreviewCommand Command = "rebuild-preview"
```
petr-muller: I think the targeted command should mention netlify, like `/netlify-rebuild` or `/netlify-preview`.

cmd/external-plugins/netlify-preview/server.go
```diff
@@ -0,0 +1,207 @@
...
+	return s.comment(ic, responseForDecision(command, decision, previewPtr))
```
petr-muller: That's definitely too noisy, especially for `/retest` commands. Most users will not do `/retest` and expect anything netlify-relevant , so responding to them is IMO just noise and confusing. When we react to the targeted command then comment response is probably in order.

apullo777:
Hi @Caesarsage, following up from kubernetes-sigs/kueue#13474. Kueue has the same operational need: Netlify deploy previews can occasionally fail for reasons outside the repository. Two failures I could verify occurred during `preparing repo` with a host-key error and during `Install dependencies` because of upstream API rate limiting, leaving required checks blocked without a self-service way to retrigger them.

@mimowo asked me to help explore supporting Kueue, so I’d be happy to represent the Kueue use case and help with the integration and testing from that side.

I read through the implementation and the existing review. Kueue looks like a useful second consumer, and it reinforces a few of the open points:

- `/retest` behavior: as @petr-muller noted, responding to ordinary `/retest` commands could be noisy. Kueue uses `/retest` frequently for unrelated Prow flakes while the Netlify preview is already green. Around 20+ PRs were opened here on a single day this week, so silently ignoring that case and keeping an explicitly Netlify-specific rebuild command would work better for us.
- Pagination: if the deploys endpoint is paginated, this could become relevant for Kueue’s PR volume, since the preview for an older PR may no longer appear in the first page of results.
- Additional preview states: Kueue uses a Netlify `build.ignore` rule, which may result in canceled or skipped previews. Those currently appear to fall into the unsupported-state path in `Evaluate()`.
- Credential configuration: the plugin currently assumes a single Netlify token for all configured repositories. Kueue uses a separate Netlify project (kubernetes-sigs-kueue), and I’m checking whether the same credential can access it. If not, supporting Kueue may require per-repository credentials rather than just another `site_id`.

Would you be open to adding `kubernetes-sigs/kueue` as another supported repository? I’m happy to help with the required code, configuration, or testing once we agree on the direction. Please let me know which parts would be most useful for me to pick up.

mimowo:
Thank you @apullo777 for the excellent summary. Indeed we are affected by the "stuck" netifly issue, so having this solved at the prow  level would be excellent. 

Let me just add some bits of my perspective, but none of that is blocking whichever way the preference lands.

> /retest behavior: as @petr-muller noted, responding to ordinary /retest commands could be noisy. Kueue uses /retest frequently for unrelated Prow flakes while the Netlify preview is already green. Around 20+ PRs were opened here on a single day this week, so silently ignoring that case and keeping an explicitly Netlify-specific rebuild command would work better for us.

Yes, but I think "/retest" only reacts to failed builds. If we could also short-circuit for Netifly and skip succeeded  ones, then using "/retest" will not generate much traffic.

> Pagination: if the deploys endpoint is paginated, this could become relevant for Kueue’s PR volume, since the preview for an older PR may no longer appear in the first page of results.

Yes, we have a lot  of PRs. but not so many documentation PRs which require Netifly - I consider if we restrict Netifly on most PRs, then we should be below 5 docs PR a day. So the need for pagination would probably depend - how many PRs would display without pagination / how many projects included etc. Still, probably worth doing at some point.

> Credential configuration: the plugin currently assumes a single Netlify token for all configured repositories. Kueue uses a separate Netlify project (kubernetes-sigs-kueue), and I’m checking whether the same credential can access it. If not, supporting Kueue may require per-repository credentials rather than just another site_id.

Excellent point, thanks for raising. The solution should allow to accommodate more projects as we go.

Caesarsage:
Thanks everyone, this is helpful. I’m going to make /retest a quiet best-effort path and keep user-facing responses only for the explicit Netlify command. I’ll also rename that command to /netlify-rebuild, since that makes the ownership and action clearer.

Caesarsage:
I’m also going to review and handle deploy pagination. Netlify documents page/per_page pagination and Link headers for multi-page API results.

For Kueue, I’m very open to supporting it. I’d like to keep this PR focused on the generic plugin behavior first, especially the quiet /retest behavior and pagination. Once we know whether the existing Netlify token can access kubernetes-sigs/kueue or whether we need per-repository credentials, we can add Kueue’s site mapping/config cleanly.

cmd/external-plugins/netlify-preview/main.go
Caesarsage: Thanks this makes sense, but since it was called out as not required immediately, I’ll leave this for follow-up rather than expanding this PR further.

cmd/external-plugins/netlify-preview/server.go
Caesarsage: It is paginated, via the Link header, 100 items per page max. There's no review_id filter server-side, so some walking is unavoidable, but since we only need the newest preview for one PR and results come newest-first, the client now walks pages with early exit on the first match and a 10-page cap instead of fetching everything (deploy-previews=true also narrows the search server-side).

PR #709, opened 2026-05-06
Bump golang version to 1.25.9

Amulyam24:
- Bump golang to 1.25.9


stmcginnis:
Just to note - a lot of community projects are actively moving to the use of a `.go-version` file. The main benefit (which doesn't apply here) is a lot of them have GitHub Action workflows that need to be updated to use the desired Go version. Since the prow repo does not have any of those, there's less of an argument to using it.

The one thing that _could_ be in favor of keeping this separation is to have a separation between what we build and test against, versus what we are declaring is the minimum compatible Go version. For instance, in `kind` these are deliberately kept separate. We want to build and test using the current/latest version of Go, but we want to keep the `go.mod` file back a few versions so anyone using the project as a Go library can do so without being forced to the same upgrade policy.

Not sure if any of that is applicable with prow, but wanted to point it out as something to consider.

petr-muller:
> We want to build and test using the current/latest version of Go, but we want to keep the go.mod file back a few versions so anyone using the project as a Go library can do so without being forced to the same upgrade policy.

Yes. I'd like to avoid bumping `go.mod` too aggressively, it causes headaches downstream. Being a library is not a primary Prow concern but we do have downstream consumers. I'm in favor of using the latest and greatest to build & test the project ourselves, but we do not need to bump `go.mod` for that.

Amulyam24:
Thank you @stmcginnis and @petr-muller, that makes sense. 
Modified to only change the `.go-version` file, PTAL!

stmcginnis:
Thanks! I think we could even bump up to the latest Go 1.26 release, but this is good.

Note: the PR description is no longer accurate. 🤷 

/lgtm

Amulyam24:
> Note: the PR description is no longer accurate. 🤷

Missed removing it, all good now :)

petr-muller (approved):
🚀

PR #710, opened 2026-05-07
fix(peribolos): skip enterprise teams during reconciliation

hoxhaeris:
## Summary

GitHub Enterprise Teams (`type: "enterprise"`) are managed at the enterprise level and cannot be modified via the org-level API. When an enterprise team is assigned to an org, peribolos fails in two ways:

1. **Team reconciliation (422):** Peribolos sees the enterprise team in `GET /orgs/{org}/teams`, doesn't find it in the config, and tries to delete it. The API rejects the operation with `422 Unprocessable Entity`.
2. **Member reconciliation (403):** Enterprise team members are injected into the org membership. Peribolos tries to remove them (not in config) and gets `403 Forbidden` because enterprise-granted membership can't be revoked at the org level.

This PR adds an `--ignore-enterprise-teams` flag that filters enterprise teams from all reconciliation paths:

- **`configureTeams`**: Enterprise teams are excluded from the teams map, so they are never deleted or updated
- **`dumpOrgConfig`**: Enterprise teams are excluded from the dumped config
- **`configureOrgMembers`**: Enterprise team members are collected and removed from the `have` set before diffing against `want`, so they never enter the removal set

Uses the `type` field from `GET /orgs/{org}/teams` to identify enterprise teams. The filtering is gated behind the flag to avoid changing default behavior.

## Changes

- `pkg/github/types.go` – Add `TeamTypeEnterprise` constant and `Type` field to `Team` struct
- `cmd/peribolos/main.go` – Add `--ignore-enterprise-teams` flag; filter enterprise teams in `configureTeams`, `dumpOrgConfig`, and `configureOrgMembers`; expand `orgClient` interface with `ListTeams` and `ListTeamMembersBySlug`
- `cmd/peribolos/main_test.go` – Add 7 test cases covering flag-on filtering, flag-off pass-through, role reconciliation for dual members, error propagation, and dump filtering

>*Tests were written with AI assistance (Claude Code), as well as the description summary.*

### DIFF
--- a/cmd/peribolos/main.go
+++ b/cmd/peribolos/main.go
@@ -59,8 +59,9 @@ type options struct {
 	fixTeamRepos      bool
 	fixRepos          bool
 	fixCollaborators  bool
-	ignoreInvitees    bool
-	ignoreSecretTeams bool
+	ignoreInvitees        bool
+	ignoreSecretTeams     bool
+	ignoreEnterpriseTeams bool
 	allowRepoArchival bool
 	allowRepoPublish  bool
 	github            flagutil.GitHubOptions
@@ -88,6 +89,7 @@ func (o *options) parseArgs(flags *flag.FlagSet, args []string) error {
 	flags.BoolVar(&o.dumpFull, "dump-full", false, "Output current config of the org as a valid input config file instead of a snippet")
 	flags.BoolVar(&o.ignoreInvitees, "ignore-invitees", false, "Do not compare missing members with active invitations (compatibility for GitHub Enterprise)")
 	flags.BoolVar(&o.ignoreSecretTeams, "ignore-secret-teams", false, "Do not dump or update secret teams if set")
+	flags.BoolVar(&o.ignoreEnterpriseTeams, "ignore-enterprise-teams", false, "Skip enterprise teams and their members during reconciliation")
 	flags.BoolVar(&o.fixOrg, "fix-org", false, "Change org metadata if set")
 	flags.BoolVar(&o.fixOrgMembers, "fix-org-members", false, "Add/remove org members if set")
 	flags.BoolVar(&o.fixTeams, "fix-teams", false, "Create/delete/update teams if set")
@@ -162,7 +164,7 @@ func main() {
 	}
 
 	if o.dump != "" {
-		ret, err := dumpOrgConfig(githubClient, o.dump, o.ignoreSecretTeams, o.github.AppID)
+		ret, err := dumpOrgConfig(githubClient, o.dump, o.ignoreSecretTeams, o.ignoreEnterpriseTeams, o.github.AppID)
 		if err != nil {
 			logrus.WithError(err).Fatalf("Dump %s failed to collect current data.", o.dump)
 		}
@@ -213,7 +215,7 @@ type dumpClient interface {
 	BotUser() (*github.UserData, error)
 }
 
-func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, appID string) (*org.Config, error) {
+func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ignoreEnterpriseTeams bool, appID string) (*org.Config, error) {
 	out := org.Config{}
 	meta, err := client.GetOrg(orgName)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


hoxhaeris:
/hold
Hold while I finish testing. While I've tested that this works fine to exclude teams, it might still fail when trying to remove an enterprise user from the Org membership.

hoxhaeris:
/unhold
Fixed a small bug (we were ignoring members, but not maintainers). In my tests, the org membership management also works fine (we ignore members who gained membership through an Enterprise Team).

hoxhaeris:
@petr-muller, making `ignoreEnterpriseTeams` default to `true` would probably make sense here (it will always fail when set to `false`, with an `admin:org` scope). However, that would change the existing default behavior, which is why I opted to keep the default as `false`.

Even with enterprise-level credentials (admin:enterprise scope), enterprise teams are managed through a separate API ([/enterprises/{enterprise}/teams](https://docs.github.com/en/rest/enterprise-teams/enterprise-teams?apiVersion=2026-03-10)), which Peribolos does not use, it only operates on `/orgs/{org}/teams`. So enterprise teams are fundamentally not manageable by Peribolos regardless of credentials, and we could just as well remove the flag altogether.

hoxhaeris:
The difference compared to `ignoreSecretTeams` is that secret teams are a policy choice, so one might legitimately want to manage them.

cmd/peribolos/main.go
```diff
@@ -439,6 +447,33 @@ func configureOrgMembers(opt options, client orgClient, orgName string, orgConfi
...
+			return fmt.Errorf("failed to list %s teams: %w", orgName, err)
+		}
+		enterpriseMembers := sets.Set[string]{}
+		for _, t := range allTeams {
+			if t.Type != github.TeamTypeEnterprise {
+				continue
+			}
+			members, err := client.ListTeamMembersBySlug(orgName, t.Slug, github.RoleAll)
+			if err != nil {
+				logrus.WithError(err).Warnf("Failed to list enterprise team %s members, skipping", t.Slug)
```
petr-muller, line 467: nit: would be better to log team name as a structured field and not in a message but no big deal

cmd/peribolos/main.go
```diff
@@ -439,6 +447,33 @@ func configureOrgMembers(opt options, client orgClient, orgName string, orgConfi
...
+			}
+			for _, m := range members {
+				enterpriseMembers.Insert(github.NormLogin(m.Login))
+			}
+		}
+		if len(enterpriseMembers) > 0 {
+			logrus.Infof("Excluding %d enterprise team members from org member reconciliation: %s",
+				len(enterpriseMembers), strings.Join(sets.List(enterpriseMembers), ", "))
+			haveAdmins = haveAdmins.Difference(enterpriseMembers)
+			haveMembers = haveMembers.Difference(enterpriseMembers)
```
petr-muller: just to make sure, are `haveAdmins` / `haveMembers` normalized at this point?  I see we normalize `have/want` several lines below, and we normalize `enterpriseMembers` above. if haveAdmins/haveMembers are not normalized we'd get case mismatches and the set difference would not be correct

petr-muller (approved):
/hold

LGTM but please check out the the concern about normalization before unholding

I wonder if this even needs to be flag gated, just as you point out. Anyone who has enterprise teams will have to enable this or run into the same errors you describe. I guess the only concern if this is not gated is one additional `ListTeams` API call - I wonder if there's a way to determine "can this org have enterprise teams" through some simpler API call and gate by that.

cmd/peribolos/main.go
hoxhaeris: Good point, we do not normalize currently, wrong ordering, fixing it.

hoxhaeris: Fixed the order.

petr-muller (approved):
Neat, thanks!

hoxhaeris:
Thanks @petr-muller!
>I guess the only concern if this is not gated is one additional ListTeams API call - I wonder if there's a way to determine "can this org have enterprise teams" through some simpler API call and gate by that.

I have unhold the PR to get our automation unstuck, and I’ll check this as well and follow up with another PR (if that's possible).

petr-muller:
> I have unhold the PR to get our automation unstuck, and I’ll check this as well and follow up with another PR (if that's possible).

Yep, no problem with increment :+1:

PR #715, opened 2026-05-11
buildlog: fail gracefully on large build logs

### LINKED ISSUE #176: Deck's build log lens should fail gracefully on large build-log.txt files.
(re-filed from https://github.com/kubernetes/test-infra/issues/27497)

> **What happened**:
> A large (e.g. >26MB) `build-log.txt` cannot be rendered by Spyglass' build log lens due to the file size. The lens shows the spinning loading symbol forever.
> 
> **What you expected to happen**:
> I don't expect the lens to be able to highlight and render such a large file, but it should fail more gracefully. We should display an error if the file is large enough to have no chance of rendering in time and if a timeout is reached that should be displayed on the front end rather than spinning the loading symbol forever.
> 
> **How to reproduce it (as minimally and precisely as possible)**:
> Create a job with lots of standard output.
> 
> /help-wanted
> 

/kind bug

carterpewpew:
Large build logs cause the buildlog lens to hang. ReadAll() succeeds but highlightLines() regex matching is too expensive. The frontend shows a spinner forever. Additionally, when ReadAll() fails, the lens silently skips the artifact with no user-facing output.

Changes:
- Skip regex highlighting for artifacts larger than 10 MiB (lines still render)
- Show errors in the UI when ReadAll() fails instead of silent continue
- Add nil-regex guard in highlightLines()
- Log and disable highlighting when Size() returns an error, as the artifact may be too large to highlight safely
- Show a warning banner in the UI when highlighting is disabled

Fixes #176

### DIFF
--- a/pkg/spyglass/lenses/buildlog/lens.go
+++ b/pkg/spyglass/lenses/buildlog/lens.go
@@ -45,6 +45,8 @@ const (
 	priority        = 10
 	neighborLines   = 5 // number of "important" lines to be displayed in either direction
 	minLinesSkipped = 5
+
+	highlightSizeThreshold = 10 * 1024 * 1024 // 10 MiB
 )
 
 var defaultHighlightLineLengthMax = 10000 // Default maximum length of a line worth highlighting
@@ -161,6 +163,8 @@ type LogArtifactView struct {
 	ShowRawLog   bool
 	CanSave      bool
 	CanAnalyze   bool
+	Error        string
+	Warning      string
 }
 
 // buildLogsView holds each log file view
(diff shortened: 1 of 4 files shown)


pkg/spyglass/lenses/buildlog/lens.go
```diff
@@ -224,8 +227,19 @@ func (lens Lens) Body(artifacts []api.Artifact, resourceDir string, data string,
...
 			continue
 		}
+
+		// For large logs, skip regex highlighting to avoid excessive CPU usage.
+		// Lines are still rendered and grouped, just without colored highlights.
+		highlightRegex := conf.highlightRegex
+		sz, err := a.Size()
+		if err == nil && sz > highlightSizeThreshold {
+			highlightRegex = nil
+		}
```
elmiko, line 247: should we log if err is not nil for some reason?

also, if err is not nil should `highlightRegex` be disabled?

elmiko (commented):
i'm not overly familiar with the prow code, but i have a question about the error state on `Size`.

also, i think if the highlighting is disabled due to file size it would be nice to have a warning note on the generated page that says something like "highlighting disabled due to file size".

pkg/spyglass/lenses/buildlog/lens.go
carterpewpew: Updated to log a warning when Size() errors and disable highlighting in that case too, since we can't be sure the artifact isn't large enough to cause a hang. Thanks for pointing this out.

carterpewpew:
> i'm not overly familiar with the prow code, but i have a question about the error state on `Size`.
> 
> also, i think if the highlighting is disabled due to file size it would be nice to have a warning note on the generated page that says something like "highlighting disabled due to file size".

@elmiko Thanks for the review, I have added a warning banner that shows up in the UI when highlighting is skipped. It says "Syntax highlighting disabled (log exceeds 10 MiB)" for large logs, or "Syntax highlighting disabled (unable to determine log size)" if Size() errors. PTAL.

elmiko (commented):
thanks for the updates!

/lgtm

petr-muller (approved):
Thank you!

PR #716, opened 2026-05-11
generic-autobumper: add GitHub App authentication support

jmguzik:
Add a new --pr-source-mode=branch option that uses GitHub App credentials (via standard flagutil.GitHubOptions CLI flags) to push branches directly to the upstream repository and create same-repo pull requests. This avoids the need for a personal access token and fork-based workflow.

The legacy PAT/fork flow remains the default and is completely untouched. When --github-app-id and --github-app-private-key-path are provided (or auto-detected), the tool uses GitClientFactory + PushToCentral for git operations and OrgAwareClient for PR creation -- the same pattern used by other Prow components.

Co-authored-by Cursor

### DIFF
--- a/cmd/generic-autobumper/bumper/bumper.go
+++ b/cmd/generic-autobumper/bumper/bumper.go
@@ -32,6 +32,7 @@ import (
 
 	"sigs.k8s.io/prow/cmd/generic-autobumper/updater"
 	"sigs.k8s.io/prow/pkg/config/secret"
+	"sigs.k8s.io/prow/pkg/flagutil"
 	"sigs.k8s.io/prow/pkg/github"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -276,6 +312,102 @@ func processGitHub(ctx context.Context, o *Options, prh PRHandler) error {
...
+			logrus.WithError(cleanErr).Warn("Failed to clean git client factory")
+		}
+	}()
+
+	repoClient, err := gitClientFactory.ClientFromDir(o.GitHubOrg, o.GitHubRepo, ".")
+	if err != nil {
+		return fmt.Errorf("create repo client for %s/%s: %w", o.GitHubOrg, o.GitHubRepo, err)
+	}
+
+	logrus.WithField("branch", o.HeadBranchName).Info("Pushing branch directly to upstream repo")
```
stmcginnis, line 390: This block should be after the `if o.SkipPullRequest` below so a branch isn't pushed when no PR is going to be created.

cmd/generic-autobumper/bumper/bumper.go
```diff
@@ -276,6 +312,102 @@ func processGitHub(ctx context.Context, o *Options, prh PRHandler) error {
...
+	if err := repoClient.PushToCentral(o.HeadBranchName, true); err != nil {
```
stmcginnis, line 391: I wonder if we should try to use [`MinimalGitPush`](https://github.com/kubernetes-sigs/prow/blob/5c4ab968b6d50fd6568f99f30d5fb30c47ac04ad/cmd/generic-autobumper/bumper/bumper.go#L527) here instead.

jmguzik: can't be used here because it's tied to the PAT/fork auth model

cmd/generic-autobumper/bumper/bumper.go
jmguzik: done

droslean (approved):

PR #719, opened 2026-05-13
 mergecommitblocker: allow use of Git subtrees with `excluded_paths` config option

RadaBDimitrova:
Up until now the `mergecommitblocker` plugin has had no configuration options. Hence, when the it was enabled, it would prevent workflows that rely on Git subtrees from functioning as intended. Git subtrees inherently depend on merge commits (e.g., those created by git subtree pull). These merge commits are required for the subtree mechanism to track history correctly. 

This PR allows configuring path patterns (e.g., `^tools/`) so that merge commits changing only those paths are permitted. 
Fixes https://github.com/kubernetes-sigs/prow/issues/666

I've chosen to do both the config introduction and the "business" logic in one PR, since the introduction of the config is a relatively minimal change and it may be easier to follow alongside the introduction of the `excluded_paths` field's logic. 

Example usage:
```yaml
merge_commit_blocker:
  - repos:
      - kubernetes/test-infra
    excluded_paths:
      - ^tools/
      - ^vendor/
```

### DIFF
--- a/pkg/git/v2/interactor.go
+++ b/pkg/git/v2/interactor.go
@@ -74,6 +74,10 @@ type Interactor interface {
 	Diff(head, sha string) (changes []string, err error)
 	// MergeCommitsExistBetween determines if merge commits exist between target and HEAD
 	MergeCommitsExistBetween(target, head string) (bool, error)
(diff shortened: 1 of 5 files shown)


RadaBDimitrova:
Is there anything additional I should do for the feature and PR?

PR #723, opened 2026-05-18
fix(clonerefs): fall back to full clone when sparse checkout fails

Prucek:
## Problem

When sparse checkout is enabled, the clone can fail for various reasons — for example, `git fetch --filter=blob:none` fails with `Could not access submodule` if the fetched commit modifies a submodule gitlink entry and the submodule path is excluded from the sparse working tree.

**Example failure:** [pull-ci-openshift-kueue-operator-release-1.3-images](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_kueue-operator/1846/pull-ci-openshift-kueue-operator-release-1.3-images/2055293913366794240)

## Fix

Instead of patching individual failure modes, add a generic fallback: if `runClone()` fails and sparse checkout was enabled, wipe the clone directory and retry with a regular full clone (no sparse checkout, no blobless filter).

The fallback record preserves commands from both the failed sparse attempt and the successful full clone for debugging visibility.

### DIFF
--- a/pkg/pod-utils/clone/clone.go
+++ b/pkg/pod-utils/clone/clone.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"net/url"
+	"os"
 	"os/exec"
 	"path"
 	"strconv"
@@ -70,10 +71,30 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
 			return sets.New[string](token)
 		}))
 	}
+
+	record = runClone(refs, dir, gitUserName, gitUserEmail, cookiePath, env, user, token)
+
+	if record.Failed && isSparseCheckoutSet(refs) {
+		logrus.WithFields(logrus.Fields{"refs": refs}).Warn("Sparse checkout clone failed, falling back to full clone")
+		failedCommands := record.Commands
+		cloneDir := PathForRefs(dir, refs)
+		if err := os.RemoveAll(cloneDir); err != nil {
+			logrus.WithError(err).Warn("Failed to clean up clone directory for fallback")
+		}
+		fullRefs := refs
+		fullRefs.SparseCheckoutFiles = nil
+		record = runClone(fullRefs, dir, gitUserName, gitUserEmail, cookiePath, env, user, token)
+		record.Commands = append(failedCommands, record.Commands...)
+	}
+
+	record.Duration = time.Since(startTime)
+	return record
+}
+
+func runClone(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, env []string, user, token string) Record {
+	record := Record{Refs: refs}
 	logrus.WithFields(logrus.Fields{"refs": refs}).Info("Cloning refs")
 
-	// This function runs the provided commands in order, logging them as they run,
-	// aborting early and returning if any command fails.
 	runCommands := func(commands []runnable) error {
 		for _, command := range commands {
 			startTime := time.Now()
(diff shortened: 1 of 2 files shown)


petr-muller:
I'd prefer some kind of fallback to non-sparse checkout when there is trouble, seems to be a safer option than silently modifying the behavior is a somewhat surprising ways.

petr-muller:
/hold

we're discussing this

pkg/pod-utils/clone/clone_test.go
```diff
@@ -945,6 +947,123 @@ func TestGitHeadTimestamp(t *testing.T) {
 	}
 }
 
+func fakeGitRepoSHA(t *testing.T, dir string) string {
```
droslean: This doesn't look fake at all.

pkg/pod-utils/clone/clone_test.go
```diff
@@ -945,6 +947,123 @@ func TestGitHeadTimestamp(t *testing.T) {
...
+	c.Dir = dir
+	out, err := c.CombinedOutput()
+	if err != nil {
+		t.Fatalf("git rev-parse HEAD in %s: %v\n%s", dir, err, out)
+	}
+	return strings.TrimSpace(string(out))
+}
+
+func TestRunSparseCheckout(t *testing.T) {
+	t.Run("fallback on failure", func(t *testing.T) {
```
droslean: Let's put the cases under a list and run the same behavior for each one of them.

pkg/pod-utils/clone/clone.go
```diff
@@ -70,13 +71,33 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
...
+func runClone(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, env []string, user, token string) Record {
+	record := Record{Refs: refs}
 	logrus.WithFields(logrus.Fields{"refs": refs}).Info("Cloning refs")
 
-	// This function runs the provided commands in order, logging them as they run,
-	// aborting early and returning if any command fails.
 	runCommands := func(commands []runnable) error {
 		for _, command := range commands {
-			startTime := time.Now()
+			cmdStart := time.Now()
```
droslean: not related to your changes.

pkg/pod-utils/clone/clone.go
```diff
@@ -92,7 +113,7 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
 				Command:  censorToken(string(secret.Censor([]byte(formattedCommand))), token),
 				Output:   censorToken(string(secret.Censor([]byte(output))), token),
 				Error:    censorToken(string(secret.Censor([]byte(message))), token),
-				Duration: time.Since(startTime),
+				Duration: time.Since(cmdStart),
```
droslean: same here.

pkg/pod-utils/clone/clone.go
```diff
@@ -70,13 +71,33 @@ func Run(refs prowapi.Refs, dir, gitUserName, gitUserEmail, cookiePath string, e
 			return sets.New[string](token)
 		}))
 	}
+
+	record = runClone(refs, dir, gitUserName, gitUserEmail, cookiePath, env, user, token)
+
+	if record.Failed && isSparseCheckoutSet(refs) {
+		logrus.WithFields(logrus.Fields{"refs": refs}).Info("Sparse checkout clone failed, falling back to full clone")
```
droslean: How will the user know about any sparse checkout permanent failures? In the end, all of the clones will be full, but we will always try to sparse-clone first, which will lead to extra work, making the whole feature pointless.

Prucek: That's why I did not want to do this fallback. But it is working almost everywhere. There are cases, like the submodules and others we don't know yet, where it does not work. The logs are stored in every PJ artifacts in `clone-log.txt`.

droslean:
/close

Closing this. I would expect another PR that fixes the module's behavior.

PR #730, opened 2026-05-23
Add gemini-agent Prow plugin

ameukam:
Implements a Prow plugin triggered by /gemini-agent <task> comments on GitHub issues and PRs. The plugin collects GitHub context (issue metadata, PR diffs, recent comments) and sends it to Gemini on Vertex AI for analysis.

Assisted by Opus 4.6

### DIFF
--- a/cmd/external-plugins/geminiagent/main.go
+++ b/cmd/external-plugins/geminiagent/main.go
@@ -1,0 +1,127 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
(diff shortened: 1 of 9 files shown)


petr-muller:
TBH I'm not a fan of these integrations being internal plugins (how many `/X-agent` implementations would we maintain inside hook?), I think they have a potential to grow in functionality and in how they are coupled with the given AI service to be standalone external plugins.

ameukam:
> TBH I'm not a fan of these integrations being internal plugins (how many `/X-agent` implementations would we maintain inside hook?), I think they have a potential to grow in functionality and in how they are coupled with the given AI service to be standalone external plugins.

I don't think we have clear policy on what's eligible for internal plugins. it's not only the AI service but everything that's interact with a third-party service.
I'll defer do your decision but IMHO we should provide more clarity on `internal`.

midnightconman:
Internal / external plugins should have better docs, but the gist of it is that external plugins (like cherrypicker) aren't compiled into hook and as such are easier to manage. 

@petr-muller if this was converted to an external plugin, could it still be staged here in the prow repo as an example?

- https://docs.prow.k8s.io/docs/components/external-plugins/cherrypicker/
- https://github.com/kubernetes-sigs/prow/tree/main/cmd/external-plugins

ameukam:
@petr-muller @midnightconman rewritten as an external plugin. PTAL!. 🙏🏾

kannon92:
I really like this idea. Who is paying for the tokens for this though?

ameukam:
> I really like this idea. Who is paying for the tokens for this though?

@kannon92 we can leverage the Community Infrastructure for this.

PR #733, opened 2026-05-27
auxiliary extra_refs

pohly:
Some periodic jobs have to check out repositories via extra_refs although they test something else:
- kops jobs must currently check out the kops repo to build the kops testgrid2 tester.
- Kubernetes E2E node jobs must check out test-infra to get the jobs/e2e_node files.

The kops tester could be built into the test image, but checking out test-infra is likely to be the permanent solution. Marking the kops resp. test-infra repo as "auxiliary" tells Prow that it must ignore them when determining the version for the started.json file. A kubetest2 tester then can set the actual version of what it is testing (typically a pre-built Kubernetes release package) in the metadata.json file.

This approach works with TestGrid's current logic of preferring a version from started.json over a metadata
version (https://github.com/GoogleCloudPlatform/testgrid/blob/d8498c5e9f91d4f8d76f1d62b9f4f7bc32862514/metadata/job.go#L185-L188) because it'll use the metadata version from finished.json when started.json doesn't have one. The metadata.json information gets copied into finished.json by https://github.com/kubernetes-sigs/prow/blob/71428b9c282ee8c9e7e9512068fccce86e7915da/pkg/sidecar/run.go#L127.

Discussion in Slack: https://kubernetes.slack.com/archives/C09QZ4DQB/p1779863293900129

### DIFF
--- a/pkg/apis/prowjobs/v1/types.go
+++ b/pkg/apis/prowjobs/v1/types.go
@@ -1222,6 +1222,19 @@ type Refs struct {
 	// directory.
 	WorkDir bool `json:"workdir,omitempty"`
 
+	// Auxiliary indicates that the repository really only provides
+	// auxiliary files for the job and is not the main repository that is
+	// being tested.
+	//
+	// This is relevant when determining which version to record in a
+	// periodic job's started.json file: the first repository where
+	// Auxiliary is false or unset is considered the main repository
+	// and determines the version.
+	//
+	// In presubmit jobs the version always comes from the repository
+	// for which the job is defined.
+	Auxiliary bool `json:"auxiliary,omitempty"`
+
 	// CloneURI is the URI that is used to clone the
 	// repository. If unset, will default to
 	// `https://github.com/org/repo.git`.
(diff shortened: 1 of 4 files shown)


pkg/apis/prowjobs/v1/types.go
```diff
@@ -1222,6 +1222,19 @@ type Refs struct {
...
+	// being tested.
+	//
+	// This is relevant when determining which version to record in a
+	// periodic job's started.json file: the first repository where
+	// Auxiliary is false or unset is considered the main repository
+	// and determines the version.
+	//
+	// In presubmit jobs the version always comes from the repository
+	// for which the job is defined.
+	Auxiliary bool `json:"auxiliary,omitempty"`
```
pohly, line 1236: Maybe this should also affect how the pod is labeled. I haven't looked at how that code works and as we have other open issues around that (like how to handle pods without any refs) I'd prefer to defer this.

petr-muller (approved):
LGTM 

Thanks!

PR #735, opened 2026-05-27
trigger: add /test-manual-required command

### LINKED ISSUE #729: [Feature] Trigger required tests with /test-required
It is already possible to trigger all tests with `/test all`, and to rerun any failed tests with `/retest`. I would want to expand on this by making it possible to trigger all *required* jobs that may not have run yet.

The background is that we commonly have a couple of required, but NOT automatically triggered tests in the metal3-io project. This helps us save resources. Contributors and/or reviewers trigger these tests manually when ready. We don't want to `/test all` though, that would trigger a bunch of tests that are rarely necessary. We just want those *required* tests to run.

What we currently do is simply to copy the test name from the pending checks and trigger them with `/test <name-of-test-1> <name-of-test-2>`, as seen below. This works but it gets annoying after a while, especially when you have to do it across multiple dependabot PRs throughout all the repositories. So to simplify, I would like to be able to do `/test-required` and be done with it.

Image

Amulyam24:
This PR adds `/test-manual-required command` to trigger required presubmits that need an explicit trigger and have no file-based conditions.

Fixes: #729

### DIFF
--- a/pkg/pjutil/filter.go
+++ b/pkg/pjutil/filter.go
@@ -35,6 +35,9 @@ var RetestRe = regexp.MustCompile(`(?m)^/retest\s*$`)
 // RetestRe provides the regex for `/retest-required`
 var RetestRequiredRe = regexp.MustCompile(`(?m)^/retest-required\s*$`)
 
+// TestRequiredRe provides the regex for `/test-required`
+var TestRequiredRe = regexp.MustCompile(`(?m)^/test-required\s*$`)
+
 var OkToTestRe = regexp.MustCompile(`(?m)^/ok-to-test\s*$`)
 
 // OkToTestCancelRe checks for cancellation of the `/ok-to-test` approval
@@ -263,6 +266,27 @@ func (rrf *RetestRequiredFilter) Name() string {
 	return "retest-required-filter"
 }
 
+type TestRequiredFilter struct {
+	allContexts sets.Set[string]
+}
+
+func NewTestRequiredFilter(allContexts sets.Set[string]) *TestRequiredFilter {
+	return &TestRequiredFilter{
+		allContexts: allContexts,
+	}
+}
+
+func (trf *TestRequiredFilter) ShouldRun(ps config.Presubmit) (bool, bool, bool) {
+	if ps.Optional {
+		return false, false, false
+	}
+	return !trf.allContexts.Has(ps.Context), ps.NeedsExplicitTrigger(), false
+}
+
+func (trf *TestRequiredFilter) Name() string {
+	return "test-required-filter"
+}
+
 type contextGetter func() (sets.Set[string], sets.Set[string], error)
 
 // PresubmitFilter creates a filter for presubmits
@@ -290,6 +314,14 @@ func PresubmitFilter(honorOkToTest bool, contextGetter contextGetter, body strin
 			return nil, err
 		}
 		filters = append(filters, NewRetestRequiredFilter(failedContexts, allContexts))
+	}
+	if TestRequiredRe.MatchString(body) {
+		logger.Info("Using test-required filter")
+		_, allContexts, err := contextGetter()
+		if err != nil {
+			return nil, err
+		}
+		filters = append(filters, NewTestRequiredFilter(allContexts))
 	}
 	if (honorOkToTest && OkToTestRe.MatchString(body)) || TestAllRe.MatchString(body) {
 		logger.Debug("Using test-all filter.")
--- a/pkg/pjutil/filter.go
+++ b/pkg/pjutil/filter.go
@@ -35,6 +35,9 @@ var RetestRe = regexp.MustCompile(`(?m)^/retest\s*$`)
 // RetestRe provides the regex for `/retest-required`
 var RetestRequiredRe = regexp.MustCompile(`(?m)^/retest-required\s*$`)
 
+// TestManualRequiredRe provides the regex for `/test-manual-required`
+var TestManualRequiredRe = regexp.MustCompile(`(?m)^/test-manual-required\s*$`)
+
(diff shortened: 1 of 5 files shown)


pkg/pjutil/filter.go
```diff
@@ -263,6 +266,27 @@ func (rrf *RetestRequiredFilter) Name() string {
...
+	return &TestRequiredFilter{
+		allContexts: allContexts,
+	}
+}
+
+func (trf *TestRequiredFilter) ShouldRun(ps config.Presubmit) (bool, bool, bool) {
+	if ps.Optional {
+		return false, false, false
+	}
+	return !ps.NeedsExplicitTrigger() && !trf.allContexts.Has(ps.Context), false, false
```
Prucek: I think `!ps.NeedsExplicitTrigger()` should not be here. It does need an explicit trigger.

Prucek:
I think this command is missing a plugin help and a description in the help page.

pkg/pjutil/filter.go
Amulyam24: Right, added it. PTAL!

Amulyam24:
Hi @Prucek @petr-muller , can you PTAL?
Thanks.

petr-muller:
Sorry for not responding sooner, I had some trouble grokking the intent and some of the decisions about the selection logic of the command (what jobs get triggered), because some of them feel arbitrary and some are inconsistent with other `/test` commands behavior.

1. Existing `/test` command unconditionally triggers the target jobs (`all` or named) no matter what their current result is. The new command adds a logic that takes existing results into account, which is new for a `/test` command and inconsistent with existing ones (except `/retest` commands). It is not really a "run required jobs" command, it is more like a "run missing jobs" and I'm not sure I like the inconsistency.
2. The command also assumes certain branch protection configuration, and this decision feels arbitrary. The situation described in #729 is not entirely standard; it is not even default. By default, jobs with `always_run: false`, `optional: false` but no conditional trigger are *not* required by Tide to merge a PR, unless someone triggers them manually (they have a "required when present" semantics). It is possible to configure branch protection / tide in a way that will enable treating these jobs as required (`require_manually_triggered_jobs`), but that's an opt-in config.

I think something like "trigger the missing required jobs" is useful, but I'd like to think if we can remove the behavior inconsistency (`/test-missing-required` is precise but awfully long...), and also check if we can make the comment respect the `require_manually_triggered_jobs` settings, just like Tide does.

Amulyam24:
Thank you @petr-muller for the detailed inputs!
Apologies for the delay in getting back on this.

As per the above, I implemented the changes to trigger all required jobs on `/test-required` irrespective of the contexts status and to respect the `require_manually_triggered_jobs` flag when set to true and trigger the jobs accordingly.

Seeing a couple of failures in the UT, will push the changes shortly.

petr-muller:
I have revisited this PR together with #729 and I have second thoughts about the command interpreting the `require_manually_triggered_jobs` config (= trying to guess what exactly beyond the automatically triggered jobs the PR needs to merge). It is what #729 directly asks but ultimately I dislike trying to be too smart and overoptimize for a certain case. Sorry for being flappy about this.

I wonder if the best way is to have a `/test-manual-required` command that only triggers (unconditionally, for consistency with other `/test` commands) the manually triggered (`always_run: false` **without** conditional -- `run_if`/`skip_if` triggers) required (`optional: false`) jobs. Ignore the `require_manually_triggered_jobs` setting - it seems that the #729 case does not even use it and instead make the jobs required *only* in branch protection (which can be seen as misconfiguration?).

That way:
- Without `require_manually_triggered_jobs`, Tide will merge PRs without these jobs if nobody triggers them (unless the separate BP policy -- the `Required` bit in GH UI -- forbids that, looks like #729 is this case), and will require them to pass if someone does
- With `require_manually_triggered_jobs`, Tide will not merge PRs without these jobs until someone triggers them manually and they pass (it seems that Tide will also trigger these jobs itself if the become mergeable per label etc criteria, TIL)

So the new `/test-manual-required` has a meaningful use in both cases, which sounds useful? I could see it being even called `/test-required` but because that name is misleading (hints that even autmatially triggered required jobs would be included) it would need to be super prominently documented.

Amulyam24:
Thank you @petr-muller! This approach defines the usage in a clear way without overlapping with the existing commands and is in line with the `/test` commands.

I have added the changes, PTAL!

One question regarding usage of the command on PRs created by non-members -- in that case, as no jobs would run, a maintainer would have to trigger` /ok-to-test` and for the explicit jobs, follow it up` /test-manual-required` , if needed, right?
Would this be the expected behaviour?

petr-muller:
> One question regarding usage of the command on PRs created by non-members -- in that case, as no jobs would run, a maintainer would have to trigger /ok-to-test and for the explicit jobs, follow it up /test-manual-required , if needed, right?

I'd expect the same like for `/test`. If the PR does not have `ok-to-test`:
- if the non-member tries to do `/test-manual-required`, it is disallowed, preferably with a  message.
- if a member does `/test-manual-required`, it should trigger the jobs even without the label as a one off

Amulyam24:
> > One question regarding usage of the command on PRs created by non-members -- in that case, as no jobs would run, a maintainer would have to trigger /ok-to-test and for the explicit jobs, follow it up /test-manual-required , if needed, right?
> 
> I'd expect the same like for `/test`. If the PR does not have `ok-to-test`:
> 
> * if the non-member tries to do `/test-manual-required`, it is disallowed, preferably with a  message.
> * if a member does `/test-manual-required`, it should trigger the jobs even without the label as a one off

Makes sense, the expectation satisfies the current behaviour, PTAL!

Amulyam24:
Hi @petr-muller, can you PTAL? Thanks!

petr-muller (approved):
Thanks!

PR #743, opened 2026-06-03
fix(autobumper): use HEAD refspec when pushing in App auth mode

Prucek:
## Summary
- `processGitHubAppAuth` commits on the checked-out branch (usually `main`) but `PushToCentral` pushes by local branch name (defaulting to `autobump`)
- Since no local branch `autobump` is created, git fails with `error: src refspec autobump does not match any`
- Fix: push `HEAD:autobump` explicitly so git pushes the current revision to the remote branch without requiring a local branch to exist

Bug introduced in #716.


petr-muller (commented):
If possible I'd prefer to put explicit control on what do we push and where, so the resulting command becomes `git push <remote> <revision>:autobump`,  like `git push <remote> HEAD:autobump`. Relying on being on a correct working copy branch feels brittle.

petr-muller (approved):

PR #744, opened 2026-06-08
tide: optionally abort superseded batch jobs before retriggering

### LINKED ISSUE #650: `tide`: obsolete batch ProwJobs not aborted when a new batch supersedes them
When the base branch SHA advances (either from a tide merge or a manual merge), tide starts a new batch without aborting the ProwJobs from the previous batch. The old ProwJobs are still running, but become invisible to the next sync cycle because dividePool() filters ProwJobs by baseSHA. Tide sees batchPending == 0 and triggers a fresh batch, leaving the old one to run to completion.                                                                                                                                                                

In the Azure/ARO-HCP case (https://prow.ci.openshift.org/tide-history?repo=Azure%2FARO-HCP), a manual merge of PR #4269, while two batches were already running, caused a third parallel batch to start. 

<img width="4402" height="3722" alt="Image" src="https://github.com/user-attachments/assets/77c9bbf7-5f8a-44e3-9f23-69aa20b323de" />
                                                                                                                                                                                                        
This is straightforward to reproduce: start a batch job, manually merge any PR in the same pool, and tide will trigger another batch on the next sync without cancelling the running one.                                                                                                                                                                                                          

Expected: when tide decides to trigger a new batch, any ProwJobs from a superseded batch (same org/repo/branch, different baseSHA) should be aborted.

/label area/tide
/kind bug

carterpewpew:
## Summary

When the base branch SHA advances (from a tide merge or manual merge), dividePool() filters ProwJobs by the new baseSHA, making old batch jobs invisible. Tide sees batchPending == 0 and triggers a fresh batch while the stale one runs to completion, wasting compute.

This adds abort of superseded batch ProwJobs just before triggering a new batch. Enabled by default, with opt-out per org/repo.

### Configuration

```yaml
tide:
  abort_superseded_batch_jobs:
    myorg/myrepo: false  # opt out for a specific repo
```

Defaults to true (enabled). Uses the standard per-org/repo map pattern (consistent with `prioritize_existing_batches`, `batch_size_limit`). Set to false to disable.

### How it works

- New method `abortSupersededBatchJobs(sp subpool)` lists all Tide-created batch ProwJobs for the same org/repo/branch regardless of baseSHA, then patches any with a stale baseSHA to AbortedState
- Called in `takeAction` right before `pickBatch`, only when the config is enabled
- Abort failures are non-fatal (logged, then new batch proceeds normally)
- Completed jobs and jobs matching the current baseSHA are skipped

Fixes #650

### DIFF
--- a/pkg/config/prow-config-documented.yaml
+++ b/pkg/config/prow-config-documented.yaml
@@ -1471,6 +1471,12 @@ slack_reporter_configs:
 # is: https://github.com/kubernetes/test-infra/issues.
 status_error_link: ' '
 tide:
+    # AbortSupersededBatchJobsMap configures on org or org/repo level whether Tide
+    # should abort older in-flight batch ProwJobs for the same branch when a new
+    # base SHA requires triggering a new batch.
+    # Use '*' as key to set this globally. Defaults to true.
+    abort_superseded_batch_jobs:
+        "": true
     # BatchSizeLimitMap is a key/value pair of an org or org/repo as the key and
     # integer batch size limit as the value. Use "*" as key to set a global default.
     # Special values:
--- a/pkg/config/tide.go
+++ b/pkg/config/tide.go
@@ -247,6 +247,11 @@ type Tide struct {
 	// starting a new one requires to start new instances of all tests.
 	// Use '*' as key to set this globally. Defaults to true.
 	PrioritizeExistingBatchesMap map[string]bool `json:"prioritize_existing_batches,omitempty"`
+	// AbortSupersededBatchJobsMap configures on org or org/repo level whether Tide
+	// should abort older in-flight batch ProwJobs for the same branch when a new
+	// base SHA requires triggering a new batch.
+	// Use '*' as key to set this globally. Defaults to true.
+	AbortSupersededBatchJobsMap map[string]bool `json:"abort_superseded_batch_jobs,omitempty"`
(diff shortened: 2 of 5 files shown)


stmcginnis:
Seems good to have, looks good to me.

/lgtm

petr-muller:
I like the idea. Could we be bold and make this behavior default, possibly with opt-out? Are there situations / deployments where the aborts would be actively harmful? Or maybe we'd make this opt-in first, test the behavior and flip the default later?

carterpewpew:
> I like the idea. Could we be bold and make this behavior default, possibly with opt-out? Are there situations / deployments where the aborts would be actively harmful? Or maybe we'd make this opt-in first, test the behavior and flip the default later?

Thank you @petr-muller, I went with opt-in for safety, but happy to flip the default to true in this PR if you'd prefer that.

stmcginnis:
I'd say usually when adding a new behavior we would want the default to be the same as it was before the change so the user has the ability to opt-in to it. But I think in this case it makes sense to make this new behavior enabled by default.

carterpewpew:
> I'd say usually when adding a new behavior we would want the default to be the same as it was before the change so the user has the ability to opt-in to it. But I think in this case it makes sense to make this new behavior enabled by default.

Thanks for taking a look, Flipped the default to true as suggested.

carterpewpew:
Hi @petr-muller, @stmcginnis , do we require any more changes for this PR

PR #746, opened 2026-06-09
tide: fix status reporting for missing newly required contexts

### LINKED ISSUE #736: Tide status controller falsely reports "In merge pool." for PRs blocked by a never-triggered required context
## Problem

When a presubmit is changed from `optional`/`always_run: false` to `always_run: true`, open PRs with all required labels get stuck: the sync loop correctly blocks them, but the status controller continues to show **"In merge pool."** The PR silently never merges.

The status reconciler is supposed to trigger the newly required job, which would eventually surface the blockage. But even when the reconciler fails (it has its own bugs), correct Tide status reporting should provide a visible signal to PR authors. This is that missing second line of defence.

## Root Cause

Two gaps in the status controller combine to suppress the correct status.

**Gap 1 — `requirementDiff` only sees contexts present on the commit.**
It calls `unsuccessfulContexts` on GitHub statuses already posted. A never-triggered required context simply isn't there, so `diff == 0`, `hasFulfilledQuery = true`, and the code falls through.

**Gap 2 — `contextCheckerGetterFactory` overwrites `RequiredContexts` with an empty slice.**
`contextCheckerGetterFactory` (`status.go:762`) does:
```go
contextPolicy.RequiredContexts = requiredContexts
```
`requiredContexts` comes from `requiredContextsMap(filteredPools)` (`tide.go:584`), which only iterates PRs that *survived* `filterSubpools`. A PR filtered out for a missing required context has no entry in the map, so this line discards everything `GetTideContextPolicy` correctly computed and replaces it with `nil`. `MissingRequiredContexts` returns nothing and the status controller emits **"In merge pool."**

Both gaps must fire together: the context must be absent from the commit (Gap 1 bypass) *and* the PR must be filtered out of the pool (Gap 2 bypass). Adding a new required presubmit to a repo with open PRs hits both simultaneously.

## Observed

https://github.com/openshift/operator-framework-operator-controller/pull/737 after https://github.com/openshift/release/pull/79764 made `openshift-e2e-aws-techpreview` required. Sync loop logged:
```
"msg":"filtering out PR as unsuccessful context is not pending",
"context":"ci/prow/openshift-e2e-aws-techpreview","pr":737
```
Status controller simultaneously reported `tide: success — "In merge pool."`. The status only corrected itself after the missing job was manually triggered, posting a `pending` GitHub status that Gap 1 could finally detect.

## Fix

**Gap 2 (higher priority):** Don't overwrite `RequiredContexts` with an empty slice in `contextCheckerGetterFactory`. Either skip the assignment when `requiredContexts` is nil/empty, or merge instead of replace:
```go
contextPolicy.RequiredContexts = append(contextPolicy.RequiredContexts, requiredContexts...)
```

**Gap 1:** `requirementDiff` should also check for required contexts missing from the commit entirely, not only contexts that are present but unsuccessful.

## Code locations

- `pkg/tide/status.go:762` — Gap 2 (overwrite)
- `pkg/tide/status.go:228–246` — Gap 1 (`requirementDiff` context check)
- `pkg/tide/status.go:349` — `MissingRequiredContexts` silently returns nothing due to Gap 2
- `pkg/tide/tide.go:814–826` — `requiredContextsMap` skips filtered-out PRs

Report assisted by Claude Code driven by a human 😅

valen-mascarenhas14:
Fixes #736

# tide: fix status reporting for missing newly required contexts

## Summary

Fixes Tide status controller incorrectly showing "In merge pool" for PRs that are blocked due to missing newly required contexts. PRs now correctly show "Not mergeable. Job <context> has not succeeded."

### 1. Detect missing required contexts (Gap 1)
Added logic to check for required contexts that are completely absent from the PR, not just contexts that are present but unsuccessful.

```go
// Check for required contexts that are missing entirely
presentContexts := sets.New[string]()
for _, ctx := range headContexts {
    presentContexts.Insert(string(ctx.Context))
}
for _, requiredContext := range cc.MissingRequiredContexts(presentContexts.UnsortedList()) {
    contexts = append(contexts, requiredContext)
}
```

### 2. Preserve base required contexts (Gap 2)
Changed `contextCheckerGetterFactory` to append additional required contexts instead of replacing them.

```go
// Before: contextPolicy.RequiredContexts = requiredContexts
if len(requiredContexts) > 0 {
    contextPolicy.RequiredContexts = append(contextPolicy.RequiredContexts, requiredContexts...)
}
```

### 3. Deduplicate context names
Added `slices.Compact()` to remove duplicate context names in error messages.

```go
sort.Strings(contexts)
contexts = slices.Compact(contexts)
```

### DIFF
--- a/pkg/tide/status.go
+++ b/pkg/tide/status.go
@@ -239,9 +239,7 @@ func requirementDiff(pr *PullRequest, q *config.TideQuery, cc contextChecker, me
 			for _, ctx := range headContexts {
 				presentContexts.Insert(string(ctx.Context))
 			}
-			for _, requiredContext := range cc.MissingRequiredContexts(presentContexts.UnsortedList()) {
-				contexts = append(contexts, requiredContext)
-			}
+			contexts = append(contexts, cc.MissingRequiredContexts(presentContexts.UnsortedList())...)
 		}
 	}
 	diff += len(contexts)
--- a/pkg/tide/status.go
+++ b/pkg/tide/status.go
@@ -46,9 +46,9 @@ import (
 )
 
 const (
-	statusContext               = "tide"
-	statusInPool                = "In merge pool."
-	statusInPoolDespiteBlocked  = "In merge pool (despite BLOCKED)."
+	statusContext              = "tide"
+	statusInPool               = "In merge pool."
+	statusInPoolDespiteBlocked = "In merge pool (despite BLOCKED)."
 	// statusNotInPool is a format string used when a PR is not in a tide pool.
 	// The '%s' field is populated with the reason why the PR is not in a
 	// tide pool or the empty string if the reason is unknown. See requirementDiff.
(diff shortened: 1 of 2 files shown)


valen-mascarenhas14:
@matthyx  @cjwagner Did you get a chance to review this PR ?

matthyx (approved):
looks good, thanks @valen-mascarenhas14

PR #748, opened 2026-06-10
Fix malformed pagination URL when resp.Request.URL differs from paged Path

jmguzik:
When the GitHub API redirects a request (e.g. repo rename/transfer, org case mismatch), Go's http.Client follows the redirect and sets resp.Request.URL to the final location. The pagination prefix logic used TrimSuffix on the full RequestURI (path+query) to detect a GHE base path like /api/v3. When the response URL differs from what we sent, the suffix doesn't match, leaving prefix set to the entire response URI. The subsequent TrimPrefix then strips too much from the next-page Link URL, producing a bare "&page=2" that gets concatenated with the base host — e.g. "https://api.github.com&page=2" — causing a DNS failure.

Fix:
- Compare only URL.Path (not the full URI with query string) when computing the GHE prefix, making it immune to query param differences.
- Add a safety check: if pagedPath doesn't start with '/' after prefix stripping, fall back to the full Link header RequestURI.

Note: the fallback can re-introduce path duplication on GHE instances that also involve a redirect. This is a known limitation — a 404 from a doubled /api/v3 prefix is strictly better than a fatal DNS crash, and the redirect+GHE combination is extremely rare. A follow-up can address it with host-aware prefix detection.

Made with Cursor

### DIFF
--- a/pkg/github/client.go
+++ b/pkg/github/client.go
@@ -1989,24 +1989,28 @@ func (c *client) readPaginatedResultsWithValuesWithContext(ctx context.Context, 
 			break
 		}
 
-		// Example for github.com:
-		// * c.bases[0]: api.github.com
-		// * initial call: api.github.com/repos/kubernetes/kubernetes/pulls?per_page=100
-		// * next: api.github.com/repositories/22/pulls?per_page=100&page=2
-		// * in this case prefix will be empty and we're just calling the path returned by next
-		// Example for github enterprise:
-		// * c.bases[0]: <ghe-url>/api/v3
-		// * initial call: <ghe-url>/api/v3/repos/kubernetes/kubernetes/pulls?per_page=100
-		// * next: <ghe-url>/api/v3/repositories/22/pulls?per_page=100&page=2
-		// * in this case prefix will be "/api/v3" and we will strip the prefix. If we don't do that,
-		//   the next call will go to <ghe-url>/api/v3/api/v3/repositories/22/pulls?per_page=100&page=2
-		prefix := strings.TrimSuffix(resp.Request.URL.RequestURI(), pagedPath)
+		// For GitHub Enterprise, c.bases[0] may include a path prefix like /api/v3.
+		// We need to strip that prefix from the Link header URL so that when we
+		// prepend c.bases[hostIndex] we don't duplicate it.
+		// For standard GitHub (with or without ghproxy), there is no path prefix.
+		//
+		// We compare only the Path components (ignoring query strings) to compute
+		// the prefix. The previous approach used the full RequestURI (path+query)
+		// for TrimSuffix, which breaks when resp.Request.URL differs from what we
+		// sent (e.g. due to a redirect): the suffix doesn't match, leaving prefix
+		// set to the entire response URI. A subsequent TrimPrefix then strips too
+		// much from the next-page URL, producing a malformed path like "&page=2".
+		pathOnly := strings.SplitN(pagedPath, "?", 2)[0]
+		prefix := strings.TrimSuffix(resp.Request.URL.Path, pathOnly)
 
 		u, err := url.Parse(link)
 		if err != nil {
 			return fmt.Errorf("failed to parse 'next' link: %w", err)
 		}
 		pagedPath = strings.TrimPrefix(u.RequestURI(), prefix)
+		if len(pagedPath) == 0 || pagedPath[0] != '/' {
+			pagedPath = u.RequestURI()
+		}
 	}
 	return nil
 }
--- a/pkg/github/client_test.go
+++ b/pkg/github/client_test.go
@@ -1373,6 +1373,120 @@ func TestReadPaginatedResults(t *testing.T) {
 	}
 }
 
+func TestReadPaginatedResultsWithValuesSamePathPagination(t *testing.T) {
+	requestCount := 0
+	ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		requestCount++
+		var labels []Label
+		switch {
+		case r.URL.Query().Get("page") == "":
+			labels = []Label{{Name: "page1"}}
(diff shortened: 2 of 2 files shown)


pkg/github/client.go
```diff
@@ -1989,24 +1989,28 @@ func (c *client) readPaginatedResultsWithValuesWithContext(ctx context.Context,
...
+		// We need to strip that prefix from the Link header URL so that when we
+		// prepend c.bases[hostIndex] we don't duplicate it.
+		// For standard GitHub (with or without ghproxy), there is no path prefix.
+		//
+		// We compare only the Path components (ignoring query strings) to compute
+		// the prefix. The previous approach used the full RequestURI (path+query)
+		// for TrimSuffix, which breaks when resp.Request.URL differs from what we
+		// sent (e.g. due to a redirect): the suffix doesn't match, leaving prefix
+		// set to the entire response URI. A subsequent TrimPrefix then strips too
+		// much from the next-page URL, producing a malformed path like "&page=2".
```
petr-muller, line 2002: Can we keep the examples in the comment, and maybe just extend the example for the "redirect" case?

pkg/github/client.go
```diff
@@ -1989,24 +1989,28 @@ func (c *client) readPaginatedResultsWithValuesWithContext(ctx context.Context,
...
-		prefix := strings.TrimSuffix(resp.Request.URL.RequestURI(), pagedPath)
+		// For GitHub Enterprise, c.bases[0] may include a path prefix like /api/v3.
+		// We need to strip that prefix from the Link header URL so that when we
+		// prepend c.bases[hostIndex] we don't duplicate it.
+		// For standard GitHub (with or without ghproxy), there is no path prefix.
+		//
+		// We compare only the Path components (ignoring query strings) to compute
+		// the prefix. The previous approach used the full RequestURI (path+query)
+		// for TrimSuffix, which breaks when resp.Request.URL differs from what we
+		// sent (e.g. due to a redirect): the suffix doesn't match, leaving prefix
```
petr-muller, line 2000: I don't think we need  to mention "the previous approach ... that breaks...". Just say thaat we must not use full URI because redirect may cause the two to differ.

pkg/github/client_test.go
```diff
@@ -1373,6 +1373,120 @@ func TestReadPaginatedResults(t *testing.T) {
...
+				newURL += "?" + r.URL.RawQuery
+			}
+			http.Redirect(w, r, newURL, http.StatusMovedPermanently)
+		case "/repos/new-org/new-repo/branches":
+			var labels []Label
+			if r.URL.Query().Get("page") == "" {
+				labels = []Label{{Name: "page1"}}
+				w.Header().Set("Link", fmt.Sprintf(
+					`<https://%s/repos/new-org/new-repo/branches?per_page=100&page=2>; rel="next"`,
+					r.Host))
```
petr-muller, line 1448: The Link header here drops the original query params, and as a result this test does not actually reproduce the bug: I checked out the pre-fix `client.go` (`HEAD~1`) and both new tests pass against it.

The old code only breaks when the `next` Link is a literal extension of the redirected response's RequestURI (which is what GitHub really sends: current URL + `&page=N`). Here the page-1 response URI ends in `?per_page=100&protected=false` but the Link says `?per_page=100&page=2`, so the old `TrimPrefix` is a no-op and the old code accidentally produces a valid URL.

Echoing the actual query fixes that — with this change I verified the test fails on the pre-fix code with exactly the error from the PR description (`parse "https://127.0.0.1:<port>&page=2": invalid port`) and passes with the fix:

```suggestion
				w.Header().Set("Link", fmt.Sprintf(
					`<https://%s/repos/new-org/new-repo/branches?%s&page=2>; rel="next"`,
					r.Host, r.URL.RawQuery))
```

pkg/github/client_test.go
```diff
@@ -1373,6 +1373,120 @@ func TestReadPaginatedResults(t *testing.T) {
...
+		switch {
+		case r.URL.Query().Get("page") == "":
+			labels = []Label{{Name: "page1"}}
+			w.Header().Set("Link", fmt.Sprintf(
+				`<https://%s%s?per_page=100&page=2>; rel="next"`,
+				r.Host, r.URL.Path))
+		case r.URL.Query().Get("page") == "2":
+			labels = []Label{{Name: "page2"}}
+		default:
+			t.Fatalf("Unexpected request: %s", r.URL.String())
```
petr-muller, line 1390: nit: `t.Fatalf` here runs on the HTTP handler goroutine, not the test goroutine — `FailNow` is documented to only work from the goroutine running the test, so this can hang or fail to stop the test. `t.Errorf` is safe from other goroutines (the redirect test's `default:` branch below does it right):

```suggestion
			t.Errorf("Unexpected request: %s", r.URL.String())
```

Same applies to the `t.Fatalf` on the marshal errors in both new tests. (Pre-existing tests in this file share the flaw, so feel free to ignore.)

pkg/github/client_test.go
```diff
@@ -1373,6 +1373,120 @@ func TestReadPaginatedResults(t *testing.T) {
...
+			return &[]Label{}
+		},
+		func(obj interface{}) {
+			labels = append(labels, *(obj.(*[]Label))...)
+		},
+	)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	expected := []Label{{Name: "page1"}, {Name: "page2"}}
```
petr-muller, line 1484: nit: consider counting requests and asserting 3 (redirect + page1 + page2) like the test above does — cheap extra signal that pagination neither looped nor short-circuited.

petr-muller:
https://github.com/kubernetes-sigs/prow/pull/748#discussion_r3390042397 is the only thing that is somewhat important to me, plus I kinda liked the the old comment with the examples. The others are nits and ignore them if you want

petr-muller (approved):
We agreed with Jakub to merge the PR and I'll followup with a small PR to address the review

PR #749, opened 2026-06-11
fix(pipeline): reject duplicate cluster endpoints at startup

### LINKED ISSUE #609: Pipeline controller deletes PipelineRuns when two cluster contexts share the same cluster
By default, Prow uses the default cluster 

We wanted to use multiple build clusters (for various workflows/testing upgrades) so we created cluster-a/cluster-b/cluster-c as well as leaving the 'default'. As we wanted to change the default cluster we use post upgrade, we used specific named clusters as the cluster: variable within our prow pre/post/periodic jobs instead of leaving unnamed (default) but now we have an issue where prow is getting confused between 2 clusters (default and another) with the same kubeconfig.

This is the code that is triggering:
https://github.com/kubernetes-sigs/prow/blob/main/cmd/pipeline/controller.go#L463

Essentially the same cluster context is duplicated with different names and the prowjobs are being deleted as the default cluster and the other cluster are technically the same build cluster, just named differently.

Is there a way to deal with this easily or is code needed to validate that the contexts are for the same cluster but named differently?

carterpewpew:
What this PR does : 

When an operator configures multiple kubeconfig context names pointing at the same physical cluster (e.g. "default" and "cluster-a" both targeting the same API server), the pipeline controller creates a separate informer per context. Both informers watch the same PipelineRun resources. When a PipelineRun event arrives through the "other" context's informer, the reconcile logic at controller.go:462 sees a context mismatch (ClusterToCtx(pj.Spec.Cluster) != ctx), concludes the PipelineRun doesn't belong, and deletes it. This is a data-loss bug: valid, running PipelineRuns get deleted silently. 

This PR adds a startup validation function (validateUniqueClusters) that detects when multiple contexts share the same API server endpoint and exits with a clear error telling the operator which contexts collide and what to fix. Only affects the pipeline (Tekton) controller running with --all-contexts. Pod-backed ProwJobs and single-context mode are not affected.

Fixes: #609

### DIFF
--- a/cmd/pipeline/main.go
+++ b/cmd/pipeline/main.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sort"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


petr-muller (approved):
Thanks!

PR #751, opened 2026-06-11
Add retry logic to handle successive failure attempts

bradmwilliams:
The Jira API can surface different problematic fields across successive attempts (the server only reports some conflicts at a time).  This PR adds a simple loop that repeatedly calls `unsetProblematicFields`, and retries, capped at 5 retries.

rh-pre-commit.version: 2.4.0
rh-pre-commit.check-secrets: ENABLED


AlexNPavel (approved):
/lgtm

PR #752, opened 2026-06-11
.*: add `delete-spam-issue` prow plugin

MadhavJivrajani:
_Disclaimer: this commit is authored with the help of Claude Code._

This PR enables https://github.com/kubernetes/org/pull/4994 

There has been a plethora of spam issues created by bots across different repos today and cleaning them up is bottlenecked on the availability of GH admins. We've had amazing folks volunteer and report such spam in a timely manner (thank you @dims!) and they are also highly trusted members of the community.

Deleting issues requires admin privileges, and rather than giving volunteers these elevated privileges, this commit introduces a prow plugin that a subset of *allowed* contributors can use to delete an issue and not rely on GH admins since we've had an increasing number of "spam storms" wherein an account creates nonsense issues in a loop resulting in a lot of them just lurking around.

The current flow that I see taking place with this is as follows:
1. Contributor drops a message on the #github-management slack noting the user ID of the account that created the spam issue (in case admins need to block/report).
2. Contributor comments `/delete-spam-issue`

TODO: the above contributor workflow also needs to be documented in k/community.

Step (1) can be further automated since we already have alerts based on certain GH events to slack setup if needed.

Currently, the allowed contributors are as follows:
1. GitHub admins
2. @dims (https://github.com/kubernetes/org/pull/4994)

We can also use this to create a named contributor role for anyone that helps out in reporting spam. 

/assign @kubernetes-sigs/owners 
for initial review

### DIFF
--- a/pkg/plugins/delete-spam-issue/delete-spam-issue.go
+++ b/pkg/plugins/delete-spam-issue/delete-spam-issue.go
@@ -1,0 +1,160 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package deletespamissue implements the delete-spam-issue plugin.
+package deletespamissue
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+
+	githubql "github.com/shurcooL/githubv4"
+	"github.com/sirupsen/logrus"
+
+	"sigs.k8s.io/prow/pkg/config"
+	"sigs.k8s.io/prow/pkg/github"
+	"sigs.k8s.io/prow/pkg/pluginhelp"
+	"sigs.k8s.io/prow/pkg/plugins"
+)
+
+const pluginName = "delete-spam-issue"
+
+var deleteSpamIssueRe = regexp.MustCompile(`(?mi)^/delete-spam-issue\s*$`)
+
+type githubClient interface {
+	CreateComment(org, repo string, number int, comment string) error
+	TeamBySlugHasMember(org string, teamSlug string, memberLogin string) (bool, error)
+	MutateWithGitHubAppsSupport(context.Context, any, githubql.Input, map[string]any, string) error
+}
+
+func init() {
+	plugins.RegisterGenericCommentHandler(pluginName, handleGenericCommentEvent, helpProvider)
+}
+
+func helpProvider(_ *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	yamlSnippet, err := plugins.CommentMap.GenYaml(&plugins.Configuration{
+		DeleteSpamIssue: plugins.DeleteSpamIssue{
+			AllowedTeams: []string{"spam-fighters"},
+			AllowedUsers: []string{"alice", "bob"},
+		},
+	})
+	if err != nil {
+		logrus.WithError(err).Warnf("cannot generate comments for %s plugin", pluginName)
+	}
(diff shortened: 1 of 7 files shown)


pkg/plugins/delete-spam-issue/delete-spam-issue.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if !canDelete {
+		return gc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, canNotDeleteReason))
+	}
+
+	issue, err := gc.GetIssue(org, repo, number)
+	if err != nil {
+		return fmt.Errorf("failed to get issue: %w", err)
+	}
+
```
cblecker: You shouldn't need this `GetIssue` call — `GenericCommentEvent` already carries `NodeID` (it's populated from `t.Issue.NodeID` in `helpers.go`). You can use `e.NodeID` directly here, which is what the `transfer-issue` plugin does for its GraphQL mutation.

Dropping this removes a redundant API call, eliminates a failure mode, and closes the TOCTOU window between fetching the issue and deleting it. You'd also be able to remove `GetIssue` from the `githubClient` interface.

cblecker (commented):
Nice work — this is a well-motivated plugin and the implementation is clean. A few thoughts:

**Consider allowing root approvers from OWNERS to run this too.** Right now the only way to authorize someone is to add them to `allowed_users` or `allowed_teams` in the central Prow config. You could add something like `allow_top_level_approvers: true` that lets top-level OWNERS file approvers for the repo also run the command. The `pin-issue` plugin already does this via `owners.TopLevelApprovers()` (see `pkg/plugins/issue-management/pin-issue.go`). That would make this more self-service for repos without needing a config change every time a new trusted person volunteers.

**Consider optional safeguards to limit blast radius.** Since issue deletion is permanent and irreversible, it might be worth adding optional config like `max_issue_age_days` and `max_issue_comments` — if an issue is older than N days or has more than M comments, the command would refuse to delete it. Spam issues are typically brand new with no legitimate interaction, so reasonable thresholds would catch accidental `/delete-spam-issue` on real issues without getting in the way of actual spam cleanup. These could default to zero (no limit) so the behavior is opt-in.

**Minor test coverage note:** The `canUserDeleteIssue` error path from `TeamBySlugHasMember` (when the API call itself fails) doesn't have test coverage. Given this is a destructive operation, it'd be good to have a test confirming the plugin returns an error and doesn't proceed with deletion when the permission check hits an API failure.

pkg/plugins/delete-spam-issue/delete-spam-issue.go
MadhavJivrajani: Good point! Done.

MadhavJivrajani:
@cblecker 

> Consider allowing root approvers from OWNERS to run this too. Right now the only way to authorize someone is to add them to allowed_users or allowed_teams in the central Prow config. You could add something like allow_top_level_approvers: true that lets top-level OWNERS file approvers for the repo also run the command. The pin-issue plugin already does this via owners.TopLevelApprovers() (see pkg/plugins/issue-management/pin-issue.go). That would make this more self-service for repos without needing a config change every time a new trusted person volunteers.

I like this idea, I've pushed a commit that does this. 

> Consider optional safeguards to limit blast radius. Since issue deletion is permanent and irreversible, it might be worth adding optional config like max_issue_age_days and max_issue_comments — if an issue is older than N days or has more than M comments, the command would refuse to delete it. Spam issues are typically brand new with no legitimate interaction, so reasonable thresholds would catch accidental /delete-spam-issue on real issues without getting in the way of actual spam cleanup. These could default to zero (no limit) so the behavior is opt-in.

Hmm, that makes sense to me. However, spam issues can lurk around in repos that aren't super actively monitored and top level OWNERS can cleanup now as needed. Maybe we can do this as a followup if folks run into scenarios of accidental deletion (outcome is same if we don't have it right now vs have it and it be opt-in for a later point). 

> Minor test coverage note: The canUserDeleteIssue error path from TeamBySlugHasMember (when the API call itself fails) doesn't have test coverage. Given this is a destructive operation, it'd be good to have a test confirming the plugin returns an error and doesn't proceed with deletion when the permission check hits an API failure.

Done.

petr-muller:
I wonder if this needs a separate new plugin - @Amulyam24's recent `issue-management` plugin seems like it could be a good home for the command. It currently does not have a config surface afaik but it should not be a problem to add it.

PR #755, opened 2026-06-14
assign: add only_org_members config with warn/block action

### LINKED ISSUE #391: assign: add configuration per repo so that only org users can assign issues
Every GSoC we see participants doing drive-by assigns for good-first-issues in our org. Example: https://github.com/kubevirt/kubevirt/issues/14085#issuecomment-2694071352

It would be helpful if we could restrict only org members to be able to assign issues to outside org contributors during that period.

Task would be to add a configuration for the assign plugin per repo, currently only one bool `onlyOrgMembers` I can imagine. This config would be injected into the handler and evaluated right before the add [1] .

See also:
* [config](https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/config.go#L52)

[1]: https://github.com/kubernetes-sigs/prow/blob/36c0045040067bb399e59b49ea2e78dded701d04/pkg/plugins/assign/assign.go#L148

### LINKED ISSUE #762: Restrict `/assign` to be run only by org members
Assign plugin commands can be used by anyone, which generates some noise from bots and drive-by contributors.
I would like to see commands like `/assign` only be trigger-able by org members.
https://github.com/kubernetes-sigs/prow/blob/37d12bec71174fdb925833014b32133404abd775/pkg/plugins/assign/assign.go#L53

The new phrasing would change from:
> Anyone can use the command, but the target user(s) must be an org member, a repo collaborator, or should have previously commented on the issue or PR.

to:
> Only an or member can use the command and the target user(s) must be an org member, a repo collaborator, or should have previously commented on the issue or PR.

carterpewpew:
Adds per-repo config for the assign plugin to restrict /assign to org members, with label-aware exemptions.

Configuration uses `map[string]*Assign` keyed by org or org/repo (like Dco):

```yaml
assign:
  org/repo:
    only_org_members: true
    exempt_labels: ["good first issue"]
    action: block
```

Options:
1. `only_org_members`: enable org membership checks on /assign
2. `action`: "warn" (default) allows the assign but posts a nudge comment; "block" rejects it
3. `exempt_labels`: labels that bypass the restriction (e.g. "good first issue")

When no config is set, behavior is unchanged.

Fixes: #391 
Fixes: #762

### DIFF
--- a/pkg/plugins/assign/assign.go
+++ b/pkg/plugins/assign/assign.go
@@ -41,16 +41,31 @@ func init() {
 	plugins.RegisterGenericCommentHandler(pluginName, handleGenericComment, helpProvider)
 }
 
-func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
-	// The Config field is omitted because this plugin is not configurable.
+func helpProvider(config *plugins.Configuration, enabledRepos []config.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	configInfo := map[string]string{}
+	for _, repo := range enabledRepos {
+		cfg := config.AssignFor(repo.Org, repo.Repo)
+		if cfg.OnlyOrgMembers {
+			action := string(plugins.AssignActionWarn)
+			if cfg.Action == plugins.AssignActionBlock {
+				action = string(plugins.AssignActionBlock)
+			}
+			msg := fmt.Sprintf("Non-org-member assigns are handled with action=%q.", action)
+			if len(cfg.ExemptLabels) > 0 {
+				msg += fmt.Sprintf(" Issues with any of the following labels are exempt: %s.", strings.Join(cfg.ExemptLabels, ", "))
+			}
+			configInfo[repo.String()] = msg
+		}
+	}
 	pluginHelp := &pluginhelp.PluginHelp{
+		Config: configInfo,
 		Description: "The assign plugin assigns or requests reviews from users/teams. Specific users can be assigned with the command '/assign @user1' or have reviews requested of them with the command '/cc @user1'. If no users are specified, the commands default to targeting the user who created the command. Assignments and requested reviews can be removed in the same way that they are added by prefixing the commands with 'un'.",
 	}
 	pluginHelp.AddCommand(pluginhelp.Command{
 		Usage:       "/[un]assign [[@]<username>...|[@]<orgname/teamname>...]",
 		Description: "Assigns assignee(s) or team(s) to the PR",
 		Featured:    true,
-		WhoCanUse:   "Anyone can use the command, but the target user(s) must be an org member, a repo collaborator, or should have previously commented on the issue or PR.",
+		WhoCanUse:   "Anyone can use the command, but the target user(s) must be an org member, a repo collaborator, or should have previously commented on the issue or PR. When only_org_members is configured, non-org-member assigns are either warned (default) or blocked depending on the action setting. Issues with exempt labels bypass this restriction.",
 		Examples:    []string{"/assign", "/unassign", "/assign @spongebob", "/assign spongebob patrick", "/assign @kubernetes/sig-foo-bar"},
 	})
 	pluginHelp.AddCommand(pluginhelp.Command{
@@ -71,13 +86,17 @@ type githubClient interface {
 	UnrequestReview(org, repo string, number int, logins []string) error
 
 	CreateComment(owner, repo string, number int, comment string) error
+
+	GetIssueLabels(org, repo string, number int) ([]github.Label, error)
+	IsMember(org, user string) (bool, error)
 }
 
 func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error {
 	if e.Action != github.GenericCommentActionCreated {
 		return nil
 	}
-	err := handle(newAssignHandler(e, pc.GitHubClient, pc.Logger))
+	cfg := pc.PluginConfig.AssignFor(e.Repo.Owner.Login, e.Repo.Name)
+	err := handle(newAssignHandler(e, pc.GitHubClient, pc.Logger, cfg))
 	if e.IsPR {
 		err = combineErrors(err, handle(newReviewHandler(e, pc.GitHubClient, pc.Logger)))
 	}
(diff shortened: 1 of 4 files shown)


pkg/plugins/config.go
```diff
@@ -100,6 +101,25 @@ type Configuration struct {
...
+	// assign themselves to issues labeled "good first issue" while still
+	// blocking drive-by assigns on other issues.
+	ExemptLabels []string `json:"exempt_labels,omitempty"`
+	// Action controls what happens when a non-org-member tries to use
+	// /assign on an issue without an exempt label. Valid values:
+	//   "warn" (default): allow the assign but post a nudge comment.
+	//   "block": reject the assign and post an error comment.
+	Action string `json:"action,omitempty"`
+}
+
```
Prucek, line 129: I think we want the config to look like this:
```
assign:
    org/repo:
      only_org_members: true
      exempt_labels: ["good first issue"]
      action: block
```
rather than:
```
assign:
    - repos:
        - org/repo
      only_org_members: true
      exempt_labels: ["good first issue"]
      action: block
```
Switching to `map[string]*AssignConfig` (like Dco) would make it cleaner and more idiomatic for this codebase

carterpewpew: Done. Switched to map[string]*Assign keyed by org or org/repo, matching the Dco pattern. AssignFor() does the same repo > org > * fallback lookup. Thanks for the suggestion, much cleaner. PTAL

pkg/plugins/assign/assign.go
```diff
@@ -187,16 +206,88 @@ type handler struct {
 	userType string
 }
 
-func newAssignHandler(e github.GenericCommentEvent, gc githubClient, log *logrus.Entry) *handler {
+func newAssignHandler(e github.GenericCommentEvent, gc githubClient, log *logrus.Entry, cfg *plugins.Assign) *handler {
+	if cfg == nil {
+		cfg = &plugins.Assign{}
+	}
```
Prucek: `AssignFor()` never returns nil, this check is not necessary

pkg/plugins/config.go
```diff
@@ -100,6 +101,23 @@ type Configuration struct {
...
+	// OnlyOrgMembers restriction, allowing anyone to self-assign.
+	// For example, setting this to ["good first issue"] lets non-members
+	// assign themselves to issues labeled "good first issue" while still
+	// blocking drive-by assigns on other issues.
+	ExemptLabels []string `json:"exempt_labels,omitempty"`
+	// Action controls what happens when a non-org-member tries to use
+	// /assign on an issue without an exempt label. Valid values:
+	//   "warn" (default): allow the assign but post a nudge comment.
+	//   "block": reject the assign and post an error comment.
+	Action string `json:"action,omitempty"`
```
Prucek: I would add some validation of this field, maybe even a type

pkg/plugins/assign/assign.go
```diff
@@ -187,16 +206,88 @@ type handler struct {
...
+			return assignErr
+		}
+
+		if block {
+			addFailureResponse = func(mu github.MissingUsers) string {
+				return fmt.Sprintf("Cannot assign the following users: %s.\n\nOnly [%s members](https://%s/orgs/%s/people) can be assigned when `only_org_members` is configured.", strings.Join(mu.Users, ", "), org, github.DefaultHost, org)
+			}
+		}
+	}
+
```
Prucek, line 224: I would like to have this as a separate function or two

pkg/plugins/assign/assign.go
carterpewpew: Good catch, removed. AssignFor() always returns &Assign{} as the fallback so this was indeed dead code. Thanks

pkg/plugins/config.go
carterpewpew: Done. Introduced an AssignAction type with "warn" and "block" constants, and added a validateAssign() function that runs at config load to reject unknown values.

pkg/plugins/assign/assign.go
carterpewpew: Refactored into three named functions: orgMemberAssignFunc (returns the closure), splitByMembership (separates logins into members/non-members), and blockNonMembers (handles the block action logic). PTAL.

pkg/plugins/assign/assign.go
```diff
@@ -205,6 +234,87 @@ func newAssignHandler(e github.GenericCommentEvent, gc githubClient, log *logrus
...
+		members, nonMembers, err := splitByMembership(gc, org, logins)
+		if err != nil {
+			return err
+		}
+
+		if cfg.Action == plugins.AssignActionBlock {
+			return blockNonMembers(gc, owner, repo, number, members, nonMembers)
+		}
+
```
Prucek, line 257: I'd add a comment here that this is the path when `AssignActionWarn` is used, it is not obvious right away.

Prucek (commented):
/ok-to-test

Prucek:
I think we can also add- fixes: https://github.com/kubernetes-sigs/prow/issues/762

pkg/plugins/assign/assign.go
carterpewpew: Added. The warn path now has a comment making the intent explicit. PTAL, thanks for the review again.

Prucek (approved):

matthyx:
@carterpewpew can you update the docs with `hack/make-rules/verify/codegen.sh` ?

matthyx:
cc @petr-muller looks like `verify/codegen.sh` isn't a blocking presubmit...

petr-muller (changes requested):
configuration surface is an interface that's hard to change once established, and configuration structs like to grow. the three new fields are coupled (=two only have meaning of the third is `true`), and if the config struct gains more knobs unrelated to the restriction, the config struct becomes messy. So I'd do something like the following:

```
assign:
  org/repo:
     restrict: # vague, but allows us to change semantics if we want
       action: # mandatory, block or warn
       exempt_labels: [] # optional list, omitzero
```

This has following advantages:
- implicit enablement when present without a problematic boolean field
- clear coupling of the two knobs to restriction feature
- `restrict` name does not lock us into the "org membership" semantics - if we need more methods  for how to restrict who can do what, we can extend the `restrict` struct any way we need

Several more notes:
- `checkconfig`'s `HasConfigFor` needs to be made aware uof the new struct for org/repo config sharding to work correctly in instances that use it
- is it supposed to work for org level as well? (= instead of `org/repo` the key woudl be simply an `org`). Most finegrained configs allow that.
- we need to make sure this only affects issues, not PRs. a non-member should still  be able to `/assign <approver>` on a PR just as the  approval instrcutions say
- I'd like the docs to be clearer about how the restriction work. The command has two modes, self (`/assign`) and someone (`/assign someone`). I'd expect non-member to be restricted from self-assignment, that feels clear. Can a member assign a non-member? Can a non-member assign a member? It is unclear from the description.

petr-muller:
The above came up a bit harsh, especially with the `lgtm` removal. That was not the intent  :sweat_smile: I appreciate the effort and I think the feature is valuable. Even my notes about the behavior are kinda secondary, the logic is not hard to change, especially for an opt-in feature like this. But the config struct layer is hard to change once it lands once, so I tend to apply more scrutiny on changes that bring new interfaces like this.

PR #770, opened 2026-06-24
trigger: expose triggering comment ID and URL in JOB_SPEC

tiborv:
Parsing arguments out of a command (e.g. `/test <args>`) is currently unreliable for jobs: a job only receives the PR refs, not the comment that triggered it. Reconstructing the comment from the PR number alone is racy — concurrent or edited `/test` comments resolve to the wrong body, and UI reruns have no comment at all.

This captures the triggering comment in `ProwJobSpec` via a new optional `trigger_comment` object and surfaces it to comment-triggered presubmits as two env vars:

- `TRIGGER_COMMENT_ID` — numeric GitHub comment ID
- `TRIGGER_COMMENT_URL` — full HTML URL to the comment

Both are also available in the JSON-encoded `JOB_SPEC` under `trigger_comment.id` / `trigger_comment.html_url`. The comment ID is a stable, unique reference to the exact comment that triggered the run, so a job can fetch its body via the GitHub API directly. The URL also enables a clickable link in reports or distinguishing a manual rerun from an automatic one.


petr-muller:
> Parsing arguments out of a command (e.g. /test <args>) is currently unreliable for jobs

It's not unreliable, it's a hack at best. We actively did not want presubmits with manually supplied arguments to exist in Prow, so I'm not too keen about enabling it through coupling job workloads with trigger comments.

tiborv:
> > Parsing arguments out of a command (e.g. /test ) is currently unreliable for jobs
> 
> It's not unreliable, it's a hack at best. We actively did not want presubmits with manually supplied arguments to exist in Prow, so I'm not too keen about enabling it through coupling job workloads with trigger comments.

If the concern is primarily about the indeterminism of job triggers if args are parsed dynamically, can I instead just try to surface more metadata about the trigger event in general? E.g. TRIGGER_EVENT_TYPE=IssueCommentEvent, TRIGGER_EVENT_ID, TRIGGER_EVENT_URL etc.

Or is the goal to actively discourage a job from 'knowing' its specific triggering event entirely?

PR #772, opened 2026-06-24
git: add SSH commit signing support to the git client factory

Prucek:
## Summary
- Add `--git-signing-key-path` flag to `GitHubOptions` for SSH-based commit signing
- When configured, the git client factory automatically sets `gpg.format=ssh`, `user.signingkey`, and `commit.gpgsign=true` on every cloned repo
- Any component using `GitHubOptions` (cherrypicker, tide, etc.) gets signing support automatically

Generated with the help of AI (Claude Code)

### DIFF
--- a/pkg/flagutil/github.go
+++ b/pkg/flagutil/github.go
@@ -48,6 +48,7 @@ type GitHubOptions struct {
 	AllowDirectAccess bool
 	AppID             string
 	AppPrivateKeyPath string
+	SigningKeyPath    string
 
 	ThrottleHourlyTokens int
 	ThrottleAllowBurst   int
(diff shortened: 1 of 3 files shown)


droslean:
/hold

I don't think this is the right direction. This feature needs to be embedded into the git client, since the component also uses and initializes. The key path could leave in the [github options](https://github.com/kubernetes-sigs/prow/blob/main/pkg/flagutil/github.go#L42-L68) too.

jmguzik:
After changes and discussion 
/lgtm
/unhold
/cc @droslean

pkg/flagutil/github.go
```diff
@@ -48,6 +48,7 @@ type GitHubOptions struct {
 	AllowDirectAccess bool
 	AppID             string
 	AppPrivateKeyPath string
+	SigningKeyPath    string
```
petr-muller, line 51: It is surprising to see this in GH flags - does not seem relevant to GH interaction, just git one?

petr-muller: Ah, I guess this is because we build git clients from this struct via`GitClientFactory`

petr-muller (approved):
lgtm and approve as far as I am concerned. I'll hold to provide @droslean an opportunity to express whether his concerns were addressed

/hold

petr-muller:
/hold cancel

I consider the opportunity as given :D

PR #774, opened 2026-06-26
peribolos: add internal repository visibility support

cblecker:
## Summary

- Replace `Private *bool` with `Visibility *RepoVisibility` on `org.Repo`, supporting `public`, `private`, and `internal` values for GitHub Enterprise organizations
- Add backward-compatible `UnmarshalJSON` that translates old `private: true/false` configs to the equivalent visibility value with a deprecation warning, and rejects configs specifying both fields
- Update `sanitizeRepoDelta` so that only transitions to `public` visibility require `--allow-repo-publish`; `private`↔`internal` transitions are allowed freely
- Add `Visibility` string field to `github.Repo` and replace `Private *bool` with `Visibility *string` on `RepoRequest`
- Validate visibility values at parse time to catch typos before any GitHub mutations
- Handle `private: null` as an omitted field for backward compatibility

## Test plan

- [x] Existing peribolos tests pass (251 tests across 2 packages)
- [x] `private: true` unmarshals to `visibility: "private"`
- [x] `private: false` unmarshals to `visibility: "public"`
- [x] `visibility: internal` unmarshals directly
- [x] Both `private` and `visibility` set produces an error
- [x] Neither set leaves `Visibility` nil
- [x] Unknown visibility values (e.g. typo `"interal"`) are rejected at parse time
- [x] `private: null` is treated as absent (leaves visibility nil)
- [x] `private→public` without `--allow-repo-publish` is blocked
- [x] `internal→public` without `--allow-repo-publish` is blocked
- [x] `private↔internal` transitions are allowed without flag
- [x] Dump/export reads `Visibility` from GitHub response and prunes `"public"` as default
- [x] `go vet` passes on all modified packages

## Attribution note

This PR was created with the assistance of AI tooling.

### DIFF
--- a/cmd/peribolos/main.go
+++ b/cmd/peribolos/main.go
@@ -368,10 +368,17 @@ func dumpOrgConfig(client dumpClient, orgName string, ignoreSecretTeams bool, ig
 		}
 		logrus.WithField("repo", full.FullName).Debug("Recording repo.")
 
+		var visibility org.RepoVisibility
+		var visibilityPtr *org.RepoVisibility
+		if err := visibility.UnmarshalText([]byte(full.Visibility)); err != nil {
+			logrus.WithField("repo", full.FullName).WithField("visibility", full.Visibility).Warn("Unknown repo visibility from GitHub API, omitting field")
+		} else {
+			visibilityPtr = &visibility
+		}
 		repoConfig := org.PruneRepoDefaults(org.Repo{
 			Description:      &full.Description,
 			HomePage:         &full.Homepage,
-			Private:          &full.Private,
+			Visibility:       visibilityPtr,
 			HasIssues:        &full.HasIssues,
 			HasProjects:      &full.HasProjects,
 			HasWiki:          &full.HasWiki,
@@ -1023,12 +1030,17 @@ type repoClient interface {
 }
 
 func newRepoCreateRequest(name string, definition org.Repo) github.RepoCreateRequest {
+	var visibility *string
+	if definition.Visibility != nil {
+		s := string(*definition.Visibility)
+		visibility = &s
+	}
 	repoCreate := github.RepoCreateRequest{
 		RepoRequest: github.RepoRequest{
 			Name:                     &name,
 			Description:              definition.Description,
 			Homepage:                 definition.HomePage,
-			Private:                  definition.Private,
+			Visibility:               visibility,
 			HasIssues:                definition.HasIssues,
 			HasProjects:              definition.HasProjects,
 			HasWiki:                  definition.HasWiki,
@@ -1090,12 +1102,19 @@ func newRepoUpdateRequest(current github.FullRepo, name string, repo org.Repo) g
 		}
 		return nil
 	}
+	var visibilityPtr *string
+	if repo.Visibility != nil {
+		want := string(*repo.Visibility)
+		if want != current.Visibility {
+			visibilityPtr = &want
+		}
+	}
 	repoUpdate := github.RepoUpdateRequest{
 		RepoRequest: github.RepoRequest{
 			Name:                     setString(current.Name, &name),
 			Description:              setString(current.Description, repo.Description),
 			Homepage:                 setString(current.Homepage, repo.HomePage),
-			Private:                  setBool(current.Private, repo.Private),
+			Visibility:               visibilityPtr,
 			HasIssues:                setBool(current.HasIssues, repo.HasIssues),
 			HasProjects:              setBool(current.HasProjects, repo.HasProjects),
 			HasWiki:                  setBool(current.HasWiki, repo.HasWiki),
@@ -1123,9 +1142,9 @@ func sanitizeRepoDelta(opt options, delta *github.RepoUpdateRequest) []error {
 		delta.Archived = nil
 		errs = append(errs, fmt.Errorf("asked to archive a repo but this is not allowed by default (see --allow-repo-archival)"))
 	}
-	if delta.Private != nil && !(*delta.Private || opt.allowRepoPublish) {
-		delta.Private = nil
-		errs = append(errs, fmt.Errorf("asked to publish a private repo but this is not allowed by default (see --allow-repo-publish)"))
+	if delta.Visibility != nil && *delta.Visibility == string(org.RepoVisibilityPublic) && !opt.allowRepoPublish {
+		delta.Visibility = nil
+		errs = append(errs, fmt.Errorf("asked to change repo visibility to public but this is not allowed by default (see --allow-repo-publish)"))
 	}
 
 	return errs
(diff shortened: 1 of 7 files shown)


cblecker:
this touches some sensitive stuff, so I want to get a through review
/hold

hoxhaeris:
Tested this against a non-enterprise test org. Results:

- `private: true` (backward compat): works, deprecation warning fires correctly
- `visibility: private`: works, idempotent on second run
- `visibility: internal` on a non-enterprise org: clear 422 from GitHub as expected
- Setting both `private` and `visibility`: rejected at parse time with a clear error

On the first run after switching from `private: true` to `visibility: private`, there's a one-time harmless delta (sends a no-op `visibility: private` update to GitHub). Second run is clean. This is expected when switching field types.

One thing to note: `ToRepo()` in `types.go` no longer sets `repo.Private` (only sets `repo.Visibility`). This only affects dry-run mode since that's where `ToRepo()` is used. No peribolos code path reads `.Private` from the dry-run result, and branchprotector reads it from real API responses, so it's fine in practice. Worth keeping in mind if other consumers of the `github` package rely on it though.

Prucek (approved):

cmd/peribolos/main.go
```diff
@@ -1123,9 +1142,9 @@ func sanitizeRepoDelta(opt options, delta *github.RepoUpdateRequest) []error {
 		delta.Archived = nil
 		errs = append(errs, fmt.Errorf("asked to archive a repo but this is not allowed by default (see --allow-repo-archival)"))
 	}
-	if delta.Private != nil && !(*delta.Private || opt.allowRepoPublish) {
-		delta.Private = nil
-		errs = append(errs, fmt.Errorf("asked to publish a private repo but this is not allowed by default (see --allow-repo-publish)"))
+	if delta.Visibility != nil && *delta.Visibility == string(org.RepoVisibilityPublic) && !opt.allowRepoPublish {
+		delta.Visibility = nil
+		errs = append(errs, fmt.Errorf("asked to change repo visibility to public but this is not allowed by default (see --allow-repo-publish)"))
```
hoxhaeris, line 1128: nit: The `--allow-repo-publish` flag help text (line 101) still says `"If set, making private repos public is allowed while updating repos"`, but this check now also blocks `internal -> public`. Consider updating the flag description to something like `"If set, changing repo visibility to public is allowed while updating repos"` to match the new three-state model.

Also, minor edge case: on older GitHub Enterprise Server instances where the `visibility` field may not be populated (returns empty string), a repo that's effectively public would produce a spurious delta (`"public" != ""`), and this check would block it. The operator would need `--allow-repo-publish` for what is effectively a no-op. Not sure how common that is in practice, but worth noting.

pkg/github/types.go
```diff
@@ -423,7 +424,7 @@ func (r RepoRequest) ToRepo() *FullRepo {
 	setString(&repo.Name, r.Name)
 	setString(&repo.Description, r.Description)
 	setString(&repo.Homepage, r.Homepage)
-	setBool(&repo.Private, r.Private)
+	setString(&repo.Visibility, r.Visibility)
```
hoxhaeris: nit: `ToRepo()` no longer sets `repo.Private`, which stays at zero value (`false`) regardless of visibility. The `Private` bool field still exists on `github.Repo` (line 334).

Currently no production code reads `.Private` from a `ToRepo()` result (peribolos discards `UpdateRepo` returns, and branchprotector reads from real API responses). But `ToRepo()` is used in dry-run mode by `CreateRepo` and `UpdateRepo` in `client.go`, so any future consumer that reads `.Private` from a dry-run `FullRepo` would see `false` even for private/internal repos.

Not blocking, but worth considering whether `ToRepo()` should also set `repo.Private = (visibility == "private" || visibility == "internal")` for consistency.

cmd/peribolos/main_test.go
```diff
@@ -34,6 +34,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
```
hoxhaeris, line 36: nit: This helper (and `strPtr` in `org_test.go`) can be replaced with `ptr.To` from `k8s.io/utils/ptr`, which is already used elsewhere in the codebase. For example: `ptr.To(org.RepoVisibilityPublic)`.

cmd/peribolos/main.go
```diff
@@ -1023,12 +1030,17 @@ type repoClient interface {
 }
 
 func newRepoCreateRequest(name string, definition org.Repo) github.RepoCreateRequest {
+	var visibility *string
+	if definition.Visibility != nil {
+		s := string(*definition.Visibility)
+		visibility = &s
```
hoxhaeris: nit: This `*RepoVisibility -> *string` conversion (nil-check, cast, address-of) appears identically in `newRepoUpdateRequest` as well (lines 1105-1111). Could be a small helper to avoid the duplication.

pkg/config/org/org_test.go
```diff
@@ -131,3 +133,87 @@ func TestPruneRepoDefaults(t *testing.T) {
...
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if !reflect.DeepEqual(tc.expected, actual) {
+				t.Errorf("result differs from expected:\n%s", diff.ObjectReflectDiff(tc.expected, actual))
+			}
+		})
+	}
```
hoxhaeris, line 203: suggestion: All backward-compat tests use `json.Unmarshal` directly, but the production config-loading path goes through `yaml.Unmarshal` (which converts YAML to JSON internally, then calls `json.Unmarshal`). This works today because `sigs.k8s.io/yaml` preserves the `private` key during YAML-to-JSON conversion and the custom `UnmarshalJSON` picks it up.

Consider adding at least one YAML-level test, e.g.:

```go
func TestRepoUnmarshalYAML(t *testing.T) {
    var repo Repo
    err := yaml.Unmarshal([]byte("private: true"), &repo)
    if err != nil {
        t.Fatalf("unexpected error: %v", err)
    }
    if repo.Visibility == nil || *repo.Visibility != RepoVisibilityPrivate {
        t.Errorf("expected visibility private, got %v", repo.Visibility)
    }
}
```

This would catch a regression if `sigs.k8s.io/yaml` ever changes its internal deserialization strategy.

hoxhaeris (commented):
Tested end-to-end against a GitHub Enterprise org to validate `internal` visibility. Covered: repo creation with all three visibility levels, all transition directions (private/internal/public, both allowed and blocked), backward compat with `private: true/false`, parse-time validation (typos, both fields set, `private: null`), idempotency, and dump round-trip. All 15 scenarios passed.

Inline comments are all nits, nothing blocking.

pkg/config/org/org_test.go
cblecker: Addressed!

cmd/peribolos/main.go
cblecker: Addressed!

cmd/peribolos/main_test.go
cblecker: Addressed!

pkg/github/types.go
cblecker: Addressed!

cmd/peribolos/main.go
cblecker: Addressed the help text. Not going to worry about the GHES stuff as it's quite an old version that it would break on. But will accept a bug report if someone actually runs into this.

cblecker:
@hoxhaeris I pushed some [changes](https://github.com/kubernetes-sigs/prow/compare/bb546f3bb24c91075730fae3706a61d11b341e8e..674a52169165592fc021626d57c951cc4654df3b) in order to simplify/DRY out the code a bit and add some more testing. I think this should be good now.

PR #775, opened 2026-06-26
fix(peribolos): include repo name in configureRepos error messages

hoxhaeris:
## Summary

- When `GetRepo`, `CreateRepo`, or `UpdateRepo` fails in `configureRepos`, the error now includes the repo name
- Previously the fatal error only showed the org name, making it hard to identify which repo caused the failure


hoxhaeris:
Closing this. The repo name is already present in the error-level log line above the fatal via the structured `"repo"` field. While having it in the fatal message too would be a nice to have, it is not really worth the review cycle.

PR #778, opened 2026-06-29
Add /override-sticky command for persistent overrides

### LINKED ISSUE #768: Support sticky overrides via /override-sticky command
## Summary

When a job is overridden via `/override`, Tide will re-trigger it when the target branch HEAD moves, or another PR is merged. This makes sense for overrides where a human reviewed the results and deemed them acceptable ("they'll be fixed later"), but not for cases where the job is simply broken or irrelevant to the PR. In the latter case, the override should "stick" and survive retests on the same PR HEAD SHA.

This was previously discussed in #238 but closed without resolution.

## Proposal

Add a `sticky_for_head` config option to the override plugin that makes `/override` persist across retests on the same PR HEAD SHA. When enabled, overrides include a `[prow:sticky]` sentinel in the GitHub status description. Tide checks for this sentinel and treats those contexts as passing, preventing unnecessary job re-triggering.

### Configuration

```yaml
override:
  sticky_for_head: true
```

### Commands

- `/override <job-name>` — when `sticky_for_head` is enabled, applies an override that persists across retests on the same HEAD SHA. When disabled, behaves as before.
- `/override-cancel [<job-name>]` — removes sticky overrides for a specific job (or all jobs if no name given), allowing normal retest semantics to resume. Only available when `sticky_for_head` is enabled.

### How it works

- The sticky override is encoded in the GitHub status description itself (`"Overridden by @user [prow:sticky]"` instead of `"Overridden by @user"`).
- Tide's `prowJobsFromContexts()` checks for the `[prow:sticky]` sentinel and synthesizes passing ProwJobs regardless of baseSHA, preventing re-triggering.
- Trigger's `/retest` naturally skips sticky overrides because they have `state=success` and are not in the failed contexts set. No trigger changes needed.
- `/test all` and `/test <job>` honor explicit user requests and will re-run the job regardless of sticky override state.

### Scope

- **Sticky across retests on the same SHA:** When the base branch moves or Tide retests, the override persists and the job is not re-triggered.
- **Not sticky across new commits:** Pushing a new commit to the PR produces a new SHA with no statuses — jobs run normally. This covers 90%+ of use cases while keeping the implementation simple.

## Use case

Repos with known-broken required jobs that can't be immediately disabled or fixed. Today, maintainers have to keep re-overriding the same job every time Tide retests, which is tedious and slows down merges.

## Related

- #238 — original discussion of Tide re-running overridden jobs

smg247:
Adds `/override-sticky` — a variant of `/override` that persists across Tide retests on the same PR HEAD SHA. Regular `/override` now also embeds the baseSHA in the status description, fixing a bug where overrides expired after ~48h when Sinker reaped the synthetic ProwJob.

### Commands

- `/override <context>` — forces a context to green. Expires when the base branch moves (same behavior as any normal CI result).
- `/override-sticky <context>` — forces a context to green and persists across base branch movement. Pushing a new commit clears it (new SHA = clean slate).
- `/override-cancel [context]` — removes sticky overrides. If no context given, cancels all sticky overrides on the PR.

### How it works

- Sticky overrides embed a `[prow:skip-retest]` sentinel in the GitHub status description. This is a generic contract owned by `config` (not override-specific) — Tide treats any successful status with this sentinel as permanently passing for the current HEAD SHA.
- Regular overrides now embed the baseSHA via `config.ContextDescriptionWithBaseSha()`, so Tide can synthesize passing ProwJobs after Sinker reaps the originals.
- Tide no longer imports the override plugin. The dependency direction is: override consumes Tide/config contracts, not the other way around.
- Trigger doesn't need changes — `/test` commands re-trigger regardless of override status, and `/retest` naturally skips successful contexts.

Fixes kubernetes-sigs/prow#768

### DIFF
--- a/pkg/plugins/config.go
+++ b/pkg/plugins/config.go
@@ -2251,10 +2251,6 @@ type Override struct {
 	// AllowedGitHubTeams is a map of orgs and/or repositories (eg "org" or "org/repo") to list of GitHub team slugs,
 	// members of which are allowed to override contexts
 	AllowedGitHubTeams map[string][]string `json:"allowed_github_teams,omitempty"`
-	// StickyForHEAD makes /override persist across retests on the same PR HEAD SHA.
-	// When enabled, Tide will not re-trigger overridden jobs. Pushing a new commit clears overrides.
-	// Also enables the /override-cancel command to remove sticky overrides.
-	StickyForHEAD bool `json:"sticky_for_head,omitempty"`
 }
 
 func (c *Configuration) mergeFrom(other *Configuration) error {
--- a/pkg/plugins/config.go
+++ b/pkg/plugins/config.go
@@ -2251,6 +2251,10 @@ type Override struct {
 	// AllowedGitHubTeams is a map of orgs and/or repositories (eg "org" or "org/repo") to list of GitHub team slugs,
 	// members of which are allowed to override contexts
 	AllowedGitHubTeams map[string][]string `json:"allowed_github_teams,omitempty"`
+	// StickyForHEAD makes /override persist across retests on the same PR HEAD SHA.
+	// When enabled, Tide will not re-trigger overridden jobs. Pushing a new commit clears overrides.
+	// Also enables the /override-cancel command to remove sticky overrides.
+	StickyForHEAD bool `json:"sticky_for_head,omitempty"`
 }
 
 func (c *Configuration) mergeFrom(other *Configuration) error {
--- a/pkg/plugins/override/override.go
+++ b/pkg/plugins/override/override.go
@@ -323,12 +323,14 @@ func authorizedGitHubTeamMember(gc githubClient, log *logrus.Entry, teamSlugs ma
 	return false
 }
 
+const overrideDescriptionPrefix = "Overridden by"
+
 func description(user string) string {
-	return fmt.Sprintf("Overridden by %s", user)
+	return fmt.Sprintf("%s %s", overrideDescriptionPrefix, user)
 }
 
 func stickyDescription(user string) string {
-	return fmt.Sprintf("Overridden by %s %s", user, config.SkipRetestSentinel)
+	return fmt.Sprintf("%s %s %s", overrideDescriptionPrefix, user, config.SkipRetestSentinel)
 }
 
 func formatList(list []string) string {
@@ -649,7 +651,7 @@ func handleOverrideCancel(oc overrideClient, log *logrus.Entry, e *github.Generi
 	cancelDesc := fmt.Sprintf("Override cancelled by %s", user)
 
 	for _, status := range statuses {
-		if !strings.Contains(status.Description, "Overridden by") {
+		if !strings.Contains(status.Description, overrideDescriptionPrefix) {
 			continue
 		}
 		if !cancelAll && !cancelContexts.Has(status.Context) {
--- a/pkg/plugins/override/override.go
+++ b/pkg/plugins/override/override.go
@@ -199,7 +199,7 @@ func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhel
 	})
 	pluginHelp.AddCommand(pluginhelp.Command{
 		Usage:       "/override-cancel [context]",
-		Description: "Cancels sticky overrides. If a context is given, only that override is cancelled. If no context is given, all sticky overrides on the PR are cancelled.",
+		Description: "Removes overrides by setting the status back to failure. Works on both regular and sticky overrides. If a context is given, only that override is removed. If no context is given, all overrides on the PR are removed.",
 		Featured:    false,
 		WhoCanUse:   whoCanUse(overrideConfig, "", ""),
 		Examples:    []string{"/override-cancel pull-repo-whatever", "/override-cancel"},
@@ -242,13 +242,24 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error 
 
 	options := pc.PluginConfig.Override
 
-	if overrideCancelRe.MatchString(e.Body) {
-		return handleOverrideCancel(c, pc.Logger, &e, options)
+	// process all three handlers, we can have cancellations, overrides, and sticky overrides in the same comment
+	if err := handleOverrideCancel(c, pc.Logger, &e, options); err != nil {
+		return err
 	}
-	if overrideStickyRe.MatchString(e.Body) {
-		return handle(c, pc.Logger, &e, options, true)
+	if err := handle(c, pc.Logger, &e, options, true); err != nil {
+		return err
 	}
 	return handle(c, pc.Logger, &e, options, false)
+}
+
+func isAuthorized(oc overrideClient, log *logrus.Entry, org, repo, user string, options plugins.Override, pr *github.PullRequest) bool {
+	if authorizedUser(oc, log, org, repo, user) {
+		return true
+	}
+	if len(options.AllowedGitHubTeams) > 0 && authorizedGitHubTeamMember(oc, log, options.AllowedGitHubTeams, org, repo, user) {
+		return true
+	}
+	return authorizedTopLevelOwner(oc, options.AllowTopLevelOwners, log, org, repo, user, pr)
 }
 
 func authorizedUser(gc githubClient, log *logrus.Entry, org, repo, user string) bool {
@@ -356,10 +367,12 @@ func handle(oc overrideClient, log *logrus.Entry, e *github.GenericCommentEvent,
 		return nil
 	}
 
+	cmdName := "/override"
 	descFn := description
 	successMsg := "Overrode contexts on behalf of %s: %s"
 	re := overrideRe
 	if sticky {
+		cmdName = "/override-sticky"
 		descFn = stickyDescription
 		successMsg = "Overrode contexts on behalf of %s: %s\n\nThese overrides will persist across retests on the current HEAD SHA. Pushing a new commit will clear them. Use `/override-cancel` to remove them."
 		re = overrideStickyRe
@@ -378,21 +391,11 @@ func handle(oc overrideClient, log *logrus.Entry, e *github.GenericCommentEvent,
 	overrides := sets.New[string]()
 	for _, m := range mat {
 		if m[1] == "" {
-			resp := "/override requires failed status contexts to operate on, but none was given"
+			resp := fmt.Sprintf("%s requires failed status contexts to operate on, but none was given", cmdName)
 			log.Debug(resp)
 			return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
 		}
 		overrides.Insert(parseOverrideInput(m[2])...)
-	}
-
-	authorized := authorizedUser(oc, log, org, repo, user)
-	if !authorized && len(options.AllowedGitHubTeams) > 0 {
-		authorized = authorizedGitHubTeamMember(oc, log, options.AllowedGitHubTeams, org, repo, user)
-	}
-	if !authorized && !options.AllowTopLevelOwners {
-		resp := fmt.Sprintf("%s unauthorized: /override is restricted to %s", user, whoCanUse(options, org, repo))
-		log.Debug(resp)
-		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
 	}
 
 	pr, err := oc.GetPullRequest(org, repo, number)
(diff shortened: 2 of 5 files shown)


pkg/tide/tide.go
```diff
@@ -46,6 +46,7 @@ import (
 	"sigs.k8s.io/prow/pkg/io"
 	"sigs.k8s.io/prow/pkg/kube"
 	"sigs.k8s.io/prow/pkg/pjutil"
+	"sigs.k8s.io/prow/pkg/plugins/override"
```
petr-muller: I think the dependency direction should be the opposite: instead of making Tide aware of override-speficic concepts, let Tide declare some (not override specific) contract for what status descriptions it would not retest, and override will just respect that contract.

I imagine some use cases for not retesting that are not related to overrides, and I don't think Tide needs to respect other code's principles - it's the other way around.

pkg/plugins/config.go
```diff
@@ -2251,6 +2251,10 @@ type Override struct {
 	// AllowedGitHubTeams is a map of orgs and/or repositories (eg "org" or "org/repo") to list of GitHub team slugs,
 	// members of which are allowed to override contexts
 	AllowedGitHubTeams map[string][]string `json:"allowed_github_teams,omitempty"`
+	// StickyForHEAD makes /override persist across retests on the same PR HEAD SHA.
+	// When enabled, Tide will not re-trigger overridden jobs. Pushing a new commit clears overrides.
+	// Also enables the /override-cancel command to remove sticky overrides.
+	StickyForHEAD bool `json:"sticky_for_head,omitempty"`
```
petr-muller: I'm not a fan of global toggle to change a command behavior, and I don't think even eventual per-repo / per-org configs would make it that better - differences in behavior would be confusing, I think. I'd generally like to avoid changing existing `/override` command semantics, it has a long established behavior.

I'd prefer a new command that's explicitly sticky (and maybe an inverse command to unstick that override).

petr-muller (commented):
Not a detailed code review yet but some early high-level feedback for now.

petr-muller:
btw, it seems that override never learned the `BaseSHA: xxx`  convention and instead relies only on the synthetic passing Prowjob creation (that would be reaped by Sinker and so the override would be retested after a Sinker cycle even if the base sha never moved...)? sounds like something we'd want to fix as well.

pkg/plugins/config.go
smg247: That is the direction I went with at first, but it seemed to me that it would be confusing for users and that most would want the _sticky_ behavior. The eventual goal would be to remove the standard behavior (or at least have sticky be the default). Since this implementation is only sticky for the same HEAD SHA, I believe that this would be the generally desirable behavior.

There is a command to unstick: `/override cancel [job-name]`.

I can go back to the separate `/override-sticky` command group though if this is an issue.

petr-muller:
> That is the direction I went with at first, but it seemed to me that it would be confusing for users and that most would want the sticky behavior. The eventual goal would be to remove the standard behavior (or at least have sticky be the default). Since this implementation is only sticky for the same HEAD SHA, I believe that this would be the generally desirable behavior.

I think the current `/override` semantics now exists long enough so that we need to be really careful to change its behavior, especially if the change goes into the direction that make things convenient while increasing risk of a merged regression (the overriden failure may not be the same failure that would be observed on a retest). I think it makes sense for more risky behavior to be explicit and not default.

I also think the many unwanted override retests were actually caused by the BaseSHA mark gap - if I understand things correctly, even if the base SHA did not move, any override was still expired by sinker. That fix alone will IMO make standard `/override` much more reasonable.

pkg/plugins/override/override.go
```diff
@@ -183,11 +185,25 @@ func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhel
...
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/override-sticky [context1] [context2]",
+		Description: "Like /override, but the override persists across retests on the current HEAD SHA even when the base branch moves. Pushing a new commit clears them. Use /override-cancel to remove.",
+		Featured:    false,
+		WhoCanUse:   whoCanUse(overrideConfig, "", ""),
+		Examples:    []string{"/override-sticky pull-repo-whatever", "/override-sticky \"test / Unit Tests\""},
+	})
+	pluginHelp.AddCommand(pluginhelp.Command{
+		Usage:       "/override-cancel [context]",
+		Description: "Cancels sticky overrides. If a context is given, only that override is cancelled. If no context is given, all sticky overrides on the PR are cancelled.",
```
petr-muller: it's not clear to me what "Cancel" means here. will it just unstick it (as if we'd use plain `/override` or will it somehow (how?) un-override it?

pkg/plugins/override/override.go
```diff
@@ -223,7 +239,16 @@ func handleGenericComment(pc plugins.Agent, e github.GenericCommentEvent) error
...
+
+	options := pc.PluginConfig.Override
+
+	if overrideCancelRe.MatchString(e.Body) {
+		return handleOverrideCancel(c, pc.Logger, &e, options)
+	}
+	if overrideStickyRe.MatchString(e.Body) {
+		return handle(c, pc.Logger, &e, options, true)
+	}
+	return handle(c, pc.Logger, &e, options, false)
```
petr-muller: It's probably a corner case but I think this means we would not handle multiple different commands in a single comment. Could we split the comment by lines and match/handle them separately / sequentially?

pkg/plugins/override/override.go
```diff
@@ -518,7 +556,7 @@ If you are trying to override a checkrun that has a space in it, you must put a
 			contextsWithCreatedJobs.Insert(status.Context)
 		}
 		status.State = github.StatusSuccess
-		status.Description = description(user)
+		status.Description = config.ContextDescriptionWithBaseSha(descFn(user), baseSHA)
```
petr-muller, line 564: Not sure if this the right place but can we add a check _somewhere_ to guard the description limit? I *think* we're fine today but it'd be great to be robust against future code changes.

descriptions are 140 chars
basesha is `BaseSHA:` + 40 chars = 48 chars
[prow:skip-retest] is 18 chars
"Overridden by " is 14
GH usernames are 39 chars max
plus two separators between segments

48+18+14+39=122

pkg/plugins/override/override.go
```diff
@@ -556,6 +594,95 @@ If you are trying to override a checkrun that has a space in it, you must put a
...
+		resp := fmt.Sprintf("Cannot get PR #%d in %s/%s", number, org, repo)
+		log.WithError(err).Warn(resp)
+		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
+	}
+
+	if !authorized && !authorizedTopLevelOwner(oc, options.AllowTopLevelOwners, log, org, repo, user, pr) {
+		resp := fmt.Sprintf("%s unauthorized: override cancel is restricted to %s", user, whoCanUse(options, org, repo))
+		log.Debug(resp)
+		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
+	}
```
petr-muller: This is the same on both standard and cancel paths, right? Can we extract it into a helper?

pkg/plugins/override/override.go
```diff
@@ -556,6 +594,95 @@ If you are trying to override a checkrun that has a space in it, you must put a
...
+
+	sha := pr.Head.SHA
+
+	// Determine which contexts to cancel
+	cancelAll := false
+	cancelContexts := sets.New[string]()
```
petr-muller, line 634: probably no need to address, but my brain thinks this code deals with `context.Context` and cancellations 😅

pkg/plugins/override/override.go
```diff
@@ -556,6 +594,95 @@ If you are trying to override a checkrun that has a space in it, you must put a
...
+		resp := fmt.Sprintf("Cannot get commit statuses for PR #%d in %s/%s", number, org, repo)
+		log.WithError(err).Warn(resp)
+		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
+	}
+
+	cancelled := sets.New[string]()
+	cancelDesc := fmt.Sprintf("Override cancelled by %s", user)
+
+	for _, status := range statuses {
+		if !config.IsSkipRetest(status.Description) {
```
petr-muller: Let's also check for `Overriden by` presence, I have other plans for `[prow:skip-retest]` and I dont want override canncellations to interfere :D

pkg/plugins/override/override.go
```diff
@@ -556,6 +594,95 @@ If you are trying to override a checkrun that has a space in it, you must put a
...
+			continue
+		}
+		if !cancelAll && !cancelContexts.Has(status.Context) {
+			continue
+		}
+		status.State = github.StatusFailure
```
petr-muller, line 660: Hmmm we inject an actual failure here. Is there a good reason why not make this work also on non-sticky overrides? Would kinda match the `/override-cancel` command anyway.

pkg/plugins/override/override.go
petr-muller: nit: we'll respond with `/override requires` (also in other error responses later) even on `/override-sticky` command, could be a bit confusing

pkg/plugins/override/override.go
```diff
@@ -492,20 +537,13 @@ If you are trying to override a checkrun that has a space in it, you must put a
...
 		// Create the overridden prow result if necessary
 		if pre != nil {
-			baseSHA, err := baseSHAGetter()
-			if err != nil {
-				resp := "Cannot get base ref of PR"
-				log.WithError(err).Warn(resp)
-				return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
-			}
-
 			pj := pjutil.NewPresubmit(*pr, baseSHA, *pre, e.GUID, nil)
```
petr-muller, line 545: If we are able to rely on statuses now, do we actually need to create the synthetic presubmits? I'm not sure. Maybe we can consider removing this as a followup...

pkg/plugins/override/override.go
```diff
@@ -472,14 +510,21 @@ If you are trying to override a checkrun that has a space in it, you must put a
 		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
 	}
 
+	baseSHA, err := baseSHAGetter()
```
petr-muller, line 518: This gets the base sha at the time of running the command, which does not seem right? But apparently that's what we do when creating the synthetic passing Prowjobs so 🤷 

We either override a Prowjob which embedded the BaseSHA into its description and then I think it more correct to reuse that. Or we override something that dos not have a BaseSHA embeddeded and I then I guess we can either get the current base sha as a fallback.

petr-muller (commented):
Thanks, bunch of feedback inline but nothing major. I think that with increased complexity it may be in order to tell an LLM to generate a new page for https://docs.prow.k8s.io/docs/components/plugins/ to document the behavior (and quirks) of the individual command behavior and interaction with Tide (standard override behaves like any other job = will be retested if the overriden job was run agastina different base,  sticky override disables that protection by Tide).

pkg/plugins/override/override.go
smg247: It un-overrides it by setting a failing status which will make it need retesting prior to merge. I will make that more clear.

pkg/plugins/override/override.go
smg247: that is a good idea. the standard override variety could also benefit from this

petr-muller:
:wave: I was not able to get here after it was updated and now I'm leaving for two weeks; no need to block on me if someone else is able to review and lgtm while I'm away

pkg/tide/tide.go
```diff
@@ -1050,8 +1051,10 @@ func (c *syncController) prowJobsFromContexts(pr *CodeReviewCommon, baseSHA stri
 		if headContext.State != githubql.StatusStateSuccess {
 			continue
 		}
-		if baseSHAForContext := config.BaseSHAFromContextDescription(string(headContext.Description)); baseSHAForContext != "" && baseSHAForContext == baseSHA {
-			passingCurrentContexts = append(passingCurrentContexts, string((headContext.Context)))
+		desc := string(headContext.Description)
+		baseSHAForContext := config.BaseSHAFromContextDescription(desc)
+		if config.IsSkipRetest(desc) || baseSHAForContext != "" && baseSHAForContext == baseSHA {
```
Prucek: I'd prefer to add explicit parentheses, which would make the intent obvious and prevent misreading

pkg/plugins/override/override.go
```diff
@@ -556,6 +597,85 @@ If you are trying to override a checkrun that has a space in it, you must put a
...
+		resp := fmt.Sprintf("Cannot get commit statuses for PR #%d in %s/%s", number, org, repo)
+		log.WithError(err).Warn(resp)
+		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
+	}
+
+	cancelled := sets.New[string]()
+	cancelDesc := fmt.Sprintf("Override cancelled by %s", user)
+
+	for _, status := range statuses {
+		if !strings.Contains(status.Description, "Overridden by") {
```
Prucek: Nit: "Overridden by" could be extracted as a constant. If the description format in `description()` and `stickyDescription()` ever changes, this stops working.

Prucek (approved):

PR #779, opened 2026-06-29
git: Disable gc.autoDetach to prevent zombie processes

petr-muller:
Prow components that use git (e.g. tide, hook) accumulate zombie `[git]` processes when running as PID 1 in a container. Git operations internally trigger `git gc --auto`, which by default forks a detached background process (`gc.autoDetach` defaults to `true` since git 2.0). Go's runtime at PID 1 only reaps children it created via `os/exec`, not orphaned processes reparented from exited git commands.

Passes `-c gc.autoDetach=false` to every git invocation in the censoring executor, making gc run synchronously so it gets properly reaped.


petr-muller:
/hold

fiddling with some alternatives

PR #781, opened 2026-06-29
deck: filter tide history per-record instead of per-pool

petr-muller:
`filterHistory` previously aggregated tenant IDs from all records in a pool into a single set, then used `HasAll` to check if Deck's tenant IDs were a superset. A single record with a non-matching tenant ID caused the entire pool's history to be hidden.

Changed to evaluate each record individually: records with matching tenant IDs are kept, others are dropped. Pools with at least one matching record remain visible.

This fixes tide history being completely invisible for repositories where some records (e.g. from batch merges) carry tenant IDs that the serving Deck instance is not configured for.

### DIFF
--- a/cmd/deck/tide.go
+++ b/cmd/deck/tide.go
@@ -187,23 +187,21 @@ func noTenantIDOrDefaultTenantID(ids []string) bool {
 	return true
 }
 
-func recordIDs(records []history.Record) sets.Set[string] {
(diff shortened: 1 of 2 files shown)


Prucek (approved):
Nice! 🚀

petr-muller:
/hold

Need to check something

PR #782, opened 2026-06-30
owners-label: add ignore_merge_commits config option

### LINKED ISSUE #610: `owners-label` should ignore merge commits
Often, new users will mistakenly push a merge commit to a PR, causing prow to then apply a large number of `sig/` and `area/` labels based on what files were changed in the merge, and then `mergecommitblocker` tags it with its label and adds a comment explaining what the user did wrong. But since `owners-label` only ever _adds_ labels, and never removes them, the erroneously-added labels stay there even after the user force-pushes over the bad merge PR, screwing up everyone's searches/filters/etc.

It would be cool if `owners-label` could be smart and not add labels based on commits that will need to be reverted.

It looks like `owners-label` always just adds all labels based on every file currently in the PR, rather than diffing against the last push. So it can just ignore any push containing a merge commit, and then it will re-sync correctly after the user force-pushes over the merge commit.

carterpewpew:
## Description

Add an `ignore_merge_commits` config option to the `owners-label` plugin. When enabled, `owners-label` skips adding labels on any pull request push that contains a merge commit. The labels will be correctly applied after the user force-pushes to remove the merge commit, since the plugin re-syncs on every `synchronize` event. This prevents spurious `sig/*` and `area/*` labels from being added when users accidentally push merge commits to their PRs.

## Motivation

When a user accidentally pushes a merge commit to a PR:

1. `owners-label` adds labels based on all files touched by the merge (potentially hundreds of unrelated `sig/` and `area/` labels).
2. `mergecommitblocker` flags the PR and the user force-pushes to fix it.
3. `mergecommitblocker` removes its label, but `owners-label` never removes labels, so the incorrect labels persist and pollute searches/filters.

With `ignore_merge_commits: true`, step 1 is skipped entirely. After the user force-pushes, `owners-label` fires again on the clean commits and applies the correct labels.

This is opt-in as suggested in the issue discussion, since not all prow deployments use `mergecommitblocker`.

Fixes: #610

### DIFF
--- a/pkg/plugins/config.go
+++ b/pkg/plugins/config.go
@@ -252,6 +252,11 @@ type Owners struct {
 	// Filenames allows configuring repos to use a separate set of filenames for
 	// any plugin that interacts with these files. Keys are in "org" or "org/repo" format.
 	Filenames map[string]ownersconfig.Filenames `json:"filenames,omitempty"`
+
+	// IgnoreMergeCommits is a list of org and org/repo strings specifying
+	// repos where the owners-label plugin should skip labeling on any
+	// pull request push that contains merge commits.
+	IgnoreMergeCommits []string `json:"ignore_merge_commits,omitempty"`
 }
 
 // OwnersFilenames determines which filenames to use for OWNERS and OWNERS_ALIASES for a repo.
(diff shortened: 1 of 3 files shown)


pkg/plugins/config.go
```diff
@@ -252,6 +252,10 @@ type Owners struct {
 	// Filenames allows configuring repos to use a separate set of filenames for
 	// any plugin that interacts with these files. Keys are in "org" or "org/repo" format.
 	Filenames map[string]ownersconfig.Filenames `json:"filenames,omitempty"`
+
+	// IgnoreMergeCommits causes the owners-label plugin to skip labeling
+	// on any pull request push that contains merge commits.
+	IgnoreMergeCommits bool `json:"ignore_merge_commits,omitempty"`
```
Prucek: Every other field on the Owners struct supports per-org or per-org/repo scoping. The new `IgnoreMergeCommits` bool is a flat global toggle.

Prucek:
@carterpewpew I think you need to run `make verify-codegen` again here. The config field is not reflected in the generated plugin-config-documented.yaml.

pkg/plugins/config.go
```diff
@@ -297,6 +302,18 @@ func (c *Configuration) SkipCollaborators(org, repo string) bool {
...
+func (c *Configuration) IgnoreMergeCommitsFor(org, repo string) bool {
+	full := fmt.Sprintf("%s/%s", org, repo)
+	for _, elem := range c.Owners.IgnoreMergeCommits {
+		if elem == org || elem == full {
+			return true
+		}
+	}
+	return false
+}
+
```
Prucek, line 316: This functionality is repeated 3 times already, it would be nice if we could factor it out

carterpewpew:
Apologies for the delay, @petr-muller, PTAL

PR #783, opened 2026-06-30
repoowners: add advisory_approvers OWNERS field

smg247:
Advisory approvers retain /approve power but are excluded from auto-assignment by LeafApprovers. This allows teams to list senior maintainers or leads who should be able to approve but shouldn't be automatically assigned as reviewers.

Implements: https://github.com/kubernetes-sigs/prow/issues/784

### DIFF
--- a/pkg/repoowners/repoowners.go
+++ b/pkg/repoowners/repoowners.go
@@ -56,6 +56,7 @@ type Config struct {
 	Approvers         []string `json:"approvers,omitempty"`
 	Reviewers         []string `json:"reviewers,omitempty"`
 	RequiredReviewers []string `json:"required_reviewers,omitempty"`
+	AdvisoryApprovers []string `json:"advisory_approvers,omitempty"`
 	Labels            []string `json:"labels,omitempty"`
 }
 
@@ -67,7 +68,7 @@ type SimpleConfig struct {
 
 // Empty checks if a SimpleConfig could be considered empty
 func (s *SimpleConfig) Empty() bool {
-	return len(s.Approvers) == 0 && len(s.Reviewers) == 0 && len(s.RequiredReviewers) == 0 && len(s.Labels) == 0
+	return len(s.Approvers) == 0 && len(s.Reviewers) == 0 && len(s.RequiredReviewers) == 0 && len(s.AdvisoryApprovers) == 0 && len(s.Labels) == 0
 }
 
 // FullConfig contains Filters which apply specific Config to files matching its regexp
@@ -247,6 +248,7 @@ type RepoOwner interface {
 	IsAutoApproveUnownedSubfolders(directory string) bool
 	LeafApprovers(path string) sets.Set[string]
 	Approvers(path string) layeredsets.String
+	AdvisoryApprovers(path string) sets.Set[string]
 	LeafReviewers(path string) sets.Set[string]
 	Reviewers(path string) layeredsets.String
 	RequiredReviewers(path string) sets.Set[string]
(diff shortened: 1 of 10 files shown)


pkg/repoowners/repoowners_test.go
```diff
@@ -1067,6 +1067,98 @@ func TestGetApprovers(t *testing.T) {
...
+func TestAdvisoryApprovers(t *testing.T) {
+	ro := &RepoOwners{
+		approvers: map[string]map[*regexp.Regexp]sets.Set[string]{
+			baseDir: regexpAll("alice", "bob", "advisory-anna"),
+			leafDir: regexpAll("carl", "dave", "advisory-dan"),
+		},
+		advisoryApprovers: map[string]map[*regexp.Regexp]sets.Set[string]{
+			baseDir: regexpAll("advisory-anna"),
+			leafDir: regexpAll("advisory-dan"),
+		},
```
JoelSpeed, line 1079: Does this imply that if I were updating the OWNERS file, I'd set it up something like below where the advisory approver is listed both in approvers and separately in advisory_approvers? Or should the lists be completely disjoint?
```
approvers:
- @Bob 
- @Bill
advisory_approvers:
- @Bob
```

smg247: Doing that would have no effect. you would want `@Bob` only in `advisory_approvers`.

smg247: Updated the test cases so it is less confusing

Prucek:
Shouldn't we update this: https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/verify-owners/verify-owners.go#L400 as well?

pkg/repoowners/repoowners.go
```diff
@@ -877,10 +897,21 @@ func (o *RepoOwners) entriesForFile(path string, people map[string]map[*regexp.R
...
+	// Advisory approvers are included in o.approvers so Approvers() recognizes
+	// their /approve authority, but must be excluded here to prevent auto-assignment.
+	all := o.entriesForFile(path, o.approvers, true).Set()
+	advisory := o.entriesForFile(path, o.advisoryApprovers, true).Set()
+	return all.Difference(advisory)
+}
+
+// AdvisoryApprovers returns the set of advisory approvers for the requested
+// file. These users retain approval power but are excluded from auto-assignment.
+func (o *RepoOwners) AdvisoryApprovers(path string) sets.Set[string] {
```
Prucek, line 913: This method is never called anywhere

pkg/repoowners/repoowners.go
```diff
@@ -749,6 +753,21 @@ func (o *RepoOwners) applyConfigToPath(path string, re *regexp.Regexp, config *C
...
+			o.advisoryApprovers[path] = make(map[*regexp.Regexp]sets.Set[string])
+		}
+		o.advisoryApprovers[path][re] = advisorySet
+		if o.approvers[path] == nil {
+			o.approvers[path] = make(map[*regexp.Regexp]sets.Set[string])
+		}
+		if existing, ok := o.approvers[path][re]; ok {
+			o.approvers[path][re] = existing.Union(advisorySet)
+		} else {
+			o.approvers[path][re] = advisorySet.Union(sets.New[string]())
```
Prucek, line 768: Union with an empty set?

damdo:
@smg247 do we still have a plan for this one? TY

smg247:
Thanks for the review @Prucek. I have addressed your comments. Ready for another look.

damdo:
@smg247 thanks. Is there some place where we can document this behaviour?

petr-muller:
Thinking about this one a bit, is there a meaningful difference from having an approver that's not a reviewer? Like

```yaml
reviewers: [A, B, C]
approvers: [A, B, C, D]
```

IIRC approvers only ever serve as blunderbuss backfill when there are not enough reviewers => advisory approvers would not be included in that so thats a difference. Also approvers are not auto-assigned, the bot just instructs the author to /assign a selected one after the PR is reviewed, so that's another. are there more? thinking whether these would be enough to introduce a new field. I don't intend to nak it or anything just thinking if this adds enough value.

PR #785, opened 2026-06-30
spyglass/junit: deduplicate test detail row rendering

petr-muller:
Extract a shared "test-detail-rows" sub-template for the expandable test result rows that were duplicated four times across the default failed/flaky and group failed/flaky sections. Unify the CSS class names (failure-*/flaky-*/group-* -> detail-*) and collapse the redundant selectors in junit.css.

Assisted by Claude Code


petr-muller:
/hold

Need to test this

Issue #780, opened 2026-06-29
Smart blunderbuss reviewer selection using git blame data


smg247:
## What would you like to be added?

Enhance the blunderbuss plugin to make smarter reviewer selections instead of choosing randomly from OWNERS files.

### 1. Blame-based reviewer ranking

Replace the current random selection algorithm (`PopRandom()`) with a scoring system that uses GitHub's GraphQL blame API to rank OWNERS-listed reviewers by their actual contributions to the changed files.

**Scoring formula** (inspired by prior art in reviewer suggestion tooling):
```
score = (line_count × 10) + (recency_score × 5) + owner_bonus
```

Where:
- `line_count` = number of blamed lines the candidate owns in the changed regions
- `recency_score` = inverse time-based score for most recent contribution
- `owner_bonus` = +3 for OWNERS approver, +2 for OWNERS reviewer

Only OWNERS-listed reviewers are eligible for assignment. Blame data is used to **rank** existing candidates, not to discover new ones.

### 2. Smart fallback when not enough reviewers found

When the configured `ReviewerCount` cannot be met from OWNERS-listed reviewers for the changed files, look at people listed as reviewers/approvers in OWNERS files elsewhere in the repo, rank them by blame data for the changed files, and assign the best matches. If blame scores are not available, fall back to random assignment.

## Why is this needed?

The current random selection frequently assigns reviewers with no context on the changed code, leading to:
- Slower review cycles (assigned reviewer has to ramp up or defer)
- Reviewer fatigue (random distribution ignores expertise), also leads to reviewers completely ignoring all assignments
- Missed opportunities to route PRs to the people best equipped to review them
- Many people want to completely turn off Blunderbuss

Blame-based selection ensures the people who actually wrote and maintain the code are prioritized.

petr-muller:
TBH I do not exactly understand the proposal here, the text seems to mix up approval powers (approve) with assigning reviewers (blunderbuss)  - where I guess the blame-based weighing could make some sense, although I'm not sure whether landing more reviews on people who touched more code than others is really a good tradeoff, it sounds like making busy people busier and not giving newer reviewers opportunities to review code they may not be yet familiar with.

I do not understand how item 2 is related to the first one, because approval is separate logic in separate plugin: approve, not blunderbuss. I somewhat see the value - give some people approval powers without being automatically assigned (although I consider that to be an organizational flaw, it'd allow people to keep powers without expecting them to use it in standard flow)

[...]

PR #787, opened 2026-06-30
`rifle`: smart reviewer selection; alternative to `blunderbuss`

smg247:
`rifle` acts as a, mutually exclusive, smart alternative plugin to `blunderbuss`. It replaces random reviewer selection with blame-based scoring that ranks OWNERS candidates by code contribution to changed files.

- Add GetBlame to GitHub client (GraphQL blame API with Apps auth)
- Score reviewers: (lines × 10) + (recency × 5) + owner bonus
- Smart fallback: when not enough reviewers found, search broader OWNERS scope ranked by blame scores, fill remaining slots randomly

Implements: kubernetes-sigs/prow#780


pkg/plugins/rifle/rifle.go
```diff
@@ -0,0 +1,461 @@
...
+	for _, c := range ranked {
+		if useStatusAvailability {
+			busy, err := reviewer.IsUserBusy(ghc, c.login)
+			if err != nil {
+				log.WithError(err).WithField("user", c.login).Warn("Failed to check user availability")
+			} else if busy {
+				busyReviewers.Insert(c.login)
+				continue
+			}
+		}
```
Prucek, line 409: I think when all users are busy, it would always return `""` which would cause an infinite loop here: 

```
for reviewers.Len() < minReviewers && unusedLeaves.Len() > 0 {
	if r := selector(&unusedLeaves); r != "" {
		reviewers.Insert(1, r)
	}
}

**Comment by stevekuznetsov** [comment]
What's the additional GitHub API quota load of this approach?

**Comment by smg247** [comment]
> What's the additional GitHub API quota load of this approach?

Up to 20 additional REST API calls for `GetBlame` per PR (capped by `maxBlameFiles`). The `IsUserBusy` GraphQL calls are the same as blunderbuss and are cached via `busyReviewers` to avoid duplicates.

Keep in mind, utilizing `rifle` instead of `blunderbuss` is _optional_, and would not be recommended if API token budget is low.

I will look into adding docs to clarify this.

File: pkg/plugins/rifle/rifle.go [hunk 2]
```diff
@@ -0,0 +1,461 @@
...
+	for _, file := range changes {
+		allReviewerCandidates = allReviewerCandidates.Union(oc.Reviewers(file.Filename).Set())
+		allApproverCandidates = allApproverCandidates.Union(oc.Approvers(file.Filename).Set())
+	}
+
+	scorer := &reviewerScorer{
+		ghc:       ghc,
+		org:       repo.Owner.Login,
+		repo:      repo.Name,
+		ref:       pr.Base.Ref,
```
Prucek: `GetBlame` is called with `pr.Base.Ref` (e.g. main), so blame is taken at whatever the branch tip is now. GitHub’s files API patches use old-side hunk line numbers from the `merge-base` (the last common ancestor of head and base). PRs are often not branched from the current tip of main, so those line numbers will not match the blamed file and we can score the wrong authors. I think we should blame the merge-base commit SHA instead. WDYT?

smg247: Thanks for pointing this out, this would definitely have led to incorrect results, especially as PRs age.

PR #788, opened 2026-06-30
verify_boilerplate: use current year instead of hardcoded range

smg247:
## Summary
- The copyright year range in `verify_boilerplate.py` was hardcoded to `range(2014, 2026)`, which excluded 2026 and would need a manual bump every January.
- Changed to `range(2014, datetime.date.today().year + 1)` so the checker automatically accepts the current year.

## Test plan
- [x] Verified a file with `Copyright 2026` passes the checker after this change
- [x] Verified existing files still pass

### DIFF
--- a/hack/boilerplate/verify_boilerplate.py
+++ b/hack/boilerplate/verify_boilerplate.py
@@ -215,7 +215,7 @@ def get_files(extensions):
 
 
 def get_dates():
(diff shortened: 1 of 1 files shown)


petr-muller:
This is intentional, new files are not expected to have dated copyright boilerplate, see https://github.com/kubernetes-sigs/prow/pull/669

PR #801, opened 2026-07-15
Add nil guard to publisher.Commit and set GitUser in GithubOptions.GitClientFactory

### LINKED ISSUE #800: publisher.Commit panics (nil pointer) when GitUser is unset, and GitHubOptions.GitClientFactory can't set it
Hi there, while writing a prow plugin I ran into the following issue and wondered if it is possible to add the suggestions I stated below.
I think it would improve the Usability of GitHubOptions.

`publisher.Commit` panics (nil pointer) when `GitUser` is unset, and `GitHubOptions.GitClientFactory` can't set it

## Problem

`RepoClient.Commit` calls the `GitUser` getter without a nil check (`pkg/git/v2/publisher.go:53-55`):

```go
name, email, err := p.info()   // publisher.go:55: panics if GitUser was never set
```

A factory built via `flagutil.GitHubOptions.GitClientFactory(...)` (`pkg/flagutil/github.go:338`) never sets `GitUser` and offers no way to set it -> so `.Commit(...)` panics:

```
panic: runtime error: invalid memory address or nil pointer dereference
sigs.k8s.io/prow/pkg/git/v2.(*publisher).Commit(...)  pkg/git/v2/publisher.go:55
```

`GitClientFactory` wires `Username`/`Token` (`github.go:355-356`, for clone/push) but not `GitUser` (for commit). `WithGitUser` exists (#507) but isn't reachable through this helper.

## Workaround

Build two factories on the same working tree: `ghOpts.GitClientFactory(...)` for checkout/push (keeps the existing PAT + GitHub App auth), and a manual `gitv2.NewClientFactory(WithGitUser(...))` used only to commit (local, needs no token).

## Suggested fix

- Nil-guard `publisher.Commit` so it returns an error instead of panicking.
- Let callers set the git user through `GitHubOptions.GitClientFactory`.

Happy to open a PR for both if welcome.

StarMiner99:
as described in #800 here is a proposed modification that would close #800.

*What does this pr do?*
- Adds a nil guard to pkg/git/v2/publisher.go publisher.Commit: p.info previously was directly called, if GitUser was not set p.info would be nil and a nil pointer dereference panic occured. I inserted a nil guard.
- in pkg/flagutil/github.go: now also sets the GitUser when using the GitClientFactory, username is the same as opts.Username and the email is left empty.

### DIFF
--- a/pkg/git/v2/publisher.go
+++ b/pkg/git/v2/publisher.go
@@ -52,6 +52,9 @@ type publisher struct {
 // Commit adds all of the current content to the index and creates a commit
 func (p *publisher) Commit(title, body string) error {
 	p.logger.Infof("Committing changes with title %q", title)
+	if p.info == nil {
+		return fmt.Errorf("error committing %q: GitUser is not set", title)
+	}
 	name, email, err := p.info()
 	if err != nil {
 		return err
--- a/pkg/flagutil/github.go
+++ b/pkg/flagutil/github.go
@@ -356,6 +356,7 @@ func (o *GitHubOptions) GitClientFactory(cookieFilePath string, cacheDir *string
 			return nil, fmt.Errorf("failed to get git authentication: %w", err)
 		}
 		opts.Username = func() (string, error) { return user, nil }
+		opts.GitUser = func() (string, string, error) { return user, "", nil } // pass empty email, git allows this
 		opts.Token = generator
 	}
 	// If the client is for Gerrit we're already set with the cookie filepath.


ylink-lfs:
Several improvement ideas:
- Expect to add tests in `pkg/git/v2/publisher_test.go` for the new behavior.
- Current the returned error doesn't explain the remedy for developers.
- Authentication and commit identity are separated, but the current changes inherit the authentication identity for commit.
- The field validation logic should follow prow's convention, not git's minimal requirement. See https://github.com/kubernetes-sigs/prow/blob/0633879af8026d056e1a5dbe1e29f5a98f6acec3/cmd/generic-autobumper/bumper/bumper.go#L169-L192

pkg/git/v2/publisher.go
```diff
@@ -52,6 +52,9 @@ type publisher struct {
 // Commit adds all of the current content to the index and creates a commit
 func (p *publisher) Commit(title, body string) error {
 	p.logger.Infof("Committing changes with title %q", title)
+	if p.info == nil {
+		return fmt.Errorf("error committing %q: GitUser is not set", title)
```
ylink-lfs, line 56: Current the returned error doesn't explain the remedy for developers.

ylink-lfs (changes requested):

PR #803, opened 2026-07-17
append the release note if there is no release note block

pacoxu:
Here is an example: https://github.com/kubernetes/kubernetes/pull/140600#issuecomment-4999451986.

This PR help to support `/release-note-edit` on no release-note block PR description.

### DIFF
--- a/pkg/plugins/releasenote/releasenote.go
+++ b/pkg/plugins/releasenote/releasenote.go
@@ -493,23 +493,41 @@ func editReleaseNote(gc githubClient, log *logrus.Entry, ic github.IssueCommentE
 			plugins.FormatResponseRaw(ic.Comment.Body, ic.Comment.HTMLURL, user, "/release-note-edit must be used with a release note block."),
 		)
 	}
+	if len(noteMatcherRE.FindAllStringSubmatchIndex(ic.Comment.Body, -1)) != 1 {
+		return gc.CreateComment(
+			org, repo, ic.Issue.Number,
+			plugins.FormatResponseRaw(ic.Comment.Body, ic.Comment.HTMLURL, user, "/release-note-edit must be used with a single release note block."),
+		)
+	}
 
 	// 0: start of release note block
 	// 1: end of release note block
 	// 2: start of release note content
 	// 3: end of release note content
 	i := noteMatcherRE.FindStringSubmatchIndex(ic.Issue.Body)
-	if len(i) != 4 {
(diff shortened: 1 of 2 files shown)


pkg/plugins/releasenote/releasenote.go
```diff
@@ -493,23 +493,41 @@ func editReleaseNote(gc githubClient, log *logrus.Entry, ic github.IssueCommentE
...
+			separator = ""
+		} else if strings.HasSuffix(ic.Issue.Body, lineEnding) {
+			separator = lineEnding
+		}
+		ic.Issue.Body += separator + "```release-note" + lineEnding + strings.TrimSpace(newNote) + lineEnding + "```" + lineEnding
+	} else {
+		// Splice in the contents of the new release note block to the top level comment.
+		// This accounts for all older regex matches.
+		b := []byte(ic.Issue.Body)
+		replaced := append(b[:i[2]], append([]byte("\r\n"+strings.TrimSpace(newNote)+"\r\n"), b[i[3]:]...)...)
```
cblecker, line 528: The new append path above correctly detects `\n` vs `\r\n` and adapts, but this splice path still hardcodes `\r\n`. For a body using LF line endings (e.g., a PR created via API), this would inject CRLF around the note content, creating mixed line endings.

Since you've already written the detection logic, consider extracting the `lineEnding` variable before the `if/else` so both paths share it.

pkg/plugins/releasenote/releasenote_test.go
```diff
@@ -764,6 +764,40 @@ func Test_editReleaseNote(t *testing.T) {
...
+					User: github.User{Login: "user"},
+				},
+				Repo: github.Repo{Owner: github.User{Login: "org"}},
+			},
+			fcFunc: func(fc *fakegithub.FakeClient) {
+				fc.OrgMembers["org"] = []string{"user"}
+				fc.Issues[issueNum] = &github.Issue{Number: issueNum, User: github.User{Login: "user"}}
+			},
+			expectedNote: "```release-note\nThe new note\n```\n",
+		},
```
cblecker, line 800: Both new test cases use LF-only bodies. Since GitHub's web UI typically produces CRLF bodies, consider adding a test with a CRLF body that has no existing release-note block (e.g., `Body: "Top\r\nBelow"`) to exercise the `strings.Contains(ic.Issue.Body, "\r\n")` branch in the append path.

PR #804, opened 2026-07-22
plank: revive pods terminated by kubelet

yuluo-yx:
Treat pods with an active DisruptionTarget condition and a TerminationByKubelet reason as unexpected infrastructure stops so  Plank can recreate them within the configured revival limit.

- For https://github.com/kubernetes-sigs/kueue/issues/12856

/cc @mimowo @mbobrovskyi


pkg/plank/reconciler.go
```diff
@@ -702,6 +707,22 @@ func getPodUnexpectedStopCause(pod *corev1.Pod) PodUnexpectedStopCause {
 	return PodUnexpectedStopCauseNone
 }
 
+func podWasTerminatedByKubelet(pod *corev1.Pod) bool {
```
mimowo, line 714: Let's add a comment here or in the unit test describing what is the scenario. In particular mention that this status is observed in the Kueue builds as the CI jobs run on spot machines which can terminate during build time. When this happens the build needs to be retried without manual intervention.
You can even point to the issue in Kueue where the issue was spot.

mimowo:
```golang
if podWasTerminatedByKubelet(pod) {
  // A kubelet-initiated termination indicates an infrastructure disruption,
  // not a test failure. For example, this can happen when a spot node is
  // reclaimed. Treat it as unexpected so Plank can revive the pod.
  // See https://github.com/kubernetes-sigs/kueue/issues/12856.
  return PodUnexpectedStopCauseTerminationByKubelet
}
```
wdyt?

mimowo:
LGTM, I only left a small readability comment proposal for the future readers: https://github.com/kubernetes-sigs/prow/pull/804#discussion_r3637085102

Thanks again for working on this 👍

mimowo:
/lgtm

@droslean  @Prucek over to you 👍

pkg/plank/reconciler.go
```diff
@@ -702,6 +711,22 @@ func getPodUnexpectedStopCause(pod *corev1.Pod) PodUnexpectedStopCause {
...
+	if pod.Status.Reason != Terminated {
+		return false
+	}
+
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type != corev1.DisruptionTarget {
+			continue
+		}
+
+		return condition.Status == corev1.ConditionTrue && condition.Reason == corev1.PodReasonTerminationByKubelet
```
mimowo, line 724: I think we could safely generalize this check for any "DisruptionTarget" reason, because the aim of this condition in the k8s core is to indicate any k8s-initiated disruptions: https://github.com/kubernetes/website/blob/e6231d08aba100c24da779ef2388c949d73159db/content/en/blog/_posts/2024/pod-failure-policy-for-jobs-goes-ga.md?plain=1#L69-L93

However, I would leave it to maintainers, we can also grow the cases as they show up.

yuluo-yx: I think it's fine. 🤔

mimowo: In that case we may add the following comment:
```golang
if podWasDisrupted(pod) {
	// DisruptionTarget identifies a retriable disruption affecting the Pod,
	// such as preemption, eviction, or termination by the kubelet. Treat it as
	// unexpected so Plank can revive the Pod. This also covers spot node
	// reclamation observed in Kueue CI.
	// See https://kubernetes.io/blog/2024/08/19/kubernetes-1-31-pod-failure-policy-for-jobs-goes-ga/#kubernetes-initiated-pod-disruptions,
	// and https://github.com/kubernetes-sigs/kueue/issues/12856.
	return PodUnexpectedStopCauseDisruption
}
```
plus adjust the tests for more cases. I would like that,  wdyt @droslean?

BenTheElder:
In the past we have not added "auto recreate pods" due to eg carefully scheduled and highly expensive and resource constrained scalability jobs. Is this opt-in?

upodroid:
We have a PDB in place in our clusters.

https://github.com/kubernetes/k8s.io/blob/49404c739d7204f0113436368b2ce241d59fbd77/kubernetes/gke-prow-build/prow/test-pods-poddisruptionbudget.yaml#L15

Are the PDBs not being honoured by Karpenter on the EKS cluster?

mimowo:
@BenTheElder thank you for stopping by, I researched the topic:

Pod revival is a pre-existing Plank mechanism. The global [`plank.max_revivals` setting](https://github.com/kubernetes-sigs/prow/blob/937cca0dd5d26c70a8601af8b1748c4b8c4ddf9f/pkg/config/config.go#L674-L678) defaults to 3, but could be changed per CI job, so is **opt-out**. The mechanism was introduced in April 2025: [#412](https://github.com/kubernetes-sigs/prow/pull/412). There are no CI jobs in test-infra changing the default which suggests the community is happy with the default.

This PR only extends the set of unexpected-stop conditions recognized by the existing mechanism to include `DisruptionTarget`. which was introduced in K8s 1.31 as GA specifically for the reason of indicating retriable failures, see the [blog post](https://kubernetes.io/blog/2024/08/19/kubernetes-1-31-pod-failure-policy-for-jobs-goes-ga/#kubernetes-initiated-pod-disruptions). While we observed such failures in Kueue they are infrequent, so I don't expect a storm of retries due to extending the policy.

From my experience leaving a CI job failed due to infra creates:
- toil for maintainers to investigate flake issues and so "test-infra",
- annoyance for contributors, who need to re-trigger the job.
- alert fatigue in the community  -  people start retriggering jobs without investigating them, which masks real bugs

So, I think extending the existing mechanism the community is happy with seems to make sense. Let me know if we should consider alternative designs.

mimowo:
Actually, apologies - while Plank defines `max_revivals`, defaulting it to 3, there is no corresponding setting on [ProwJobSpec](https://github.com/kubernetes-sigs/prow/blob/0633879af8026d056e1a5dbe1e29f5a98f6acec3/pkg/apis/prowjobs/v1/types.go#L142), which may explain why no CI job re-defines the default in test-infra.

Conceptually, I think it could belong under  ProwJob's `spec.decorator_config`, alongside the timeout-related settings. If introduced there, plumbing it through to Plank should be relatively straightforward. At the moment, however, there does not appear to be a per-job opt-out, AFAIK.

mimowo:
@upodroid thank you for taking a look as well.

IIUC, the PDB with `maxUnavailable: 0` prevents eviction of the Prow Pods to be initiated by Karpenter, eg during scale-down.

The Spot interruption case is different. When AWS decides to reclaim an instance it gives Karpenter only a short interruption-notice window. Karpenter can use that time to launch a replacement node and attempt to drain the affected node, but it cannot defer the EC2 instance termination. 

So I think the PDB protects against Karpenter-initiated voluntary disruption, while Plank revival is still needed to recover from an involuntary Spot interruption.

upodroid:
We shouldn't be using spot instances at all. I'll look at it again

mimowo:
> We shouldn't be using spot instances at all. I'll look at it again

So for reference here is our Kueue [issue](https://github.com/kubernetes-sigs/kueue/issues/12856) where we investigated (see [post]()https://github.com/kubernetes-sigs/kueue/issues/12856#issuecomment-5046910654) the [CI failure](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12777/pull-kueue-test-e2e-sequential-baseline-shard-0-main/2074518954436464640)  from "07/07/2026". 

In particular the attached artifact [podinfo.json](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kueue/12777/pull-kueue-test-e2e-sequential-baseline-shard-0-main/2074518954436464640/podinfo.json) contains the following information: `"karpenter.sh/capacity-type": "spot"`.

Here is the more complete snippet with the node information:
```json
	"node": {
		"metadata": {
			"name": "ip-10-0-68-68.us-east-2.compute.internal",
			"uid": "8e2aa498-0a3a-4bf7-b6d5-c70a71d78c57",
			"resourceVersion": "1331922906",
			"creationTimestamp": "2026-07-07T15:41:06Z",
			"deletionTimestamp": "2026-07-07T15:50:24Z",
			"deletionGracePeriodSeconds": 0,
			"labels": {
				"beta.kubernetes.io/arch": "amd64",
				"beta.kubernetes.io/instance-type": "m6idn.2xlarge",
				"beta.kubernetes.io/os": "linux",
				"failure-domain.beta.kubernetes.io/region": "us-east-2",
				"failure-domain.beta.kubernetes.io/zone": "us-east-2b",
				"k8s.io/cloud-provider-aws": "99afed85e58f0057554bd51076ed54f5",
				"karpenter.k8s.aws/ec2nodeclass": "default",
				"karpenter.k8s.aws/instance-capability-flex": "false",
				"karpenter.k8s.aws/instance-category": "m",
				"karpenter.k8s.aws/instance-cpu": "8",
				"karpenter.k8s.aws/instance-cpu-manufacturer": "intel",
				"karpenter.k8s.aws/instance-cpu-sustained-clock-speed-mhz": "3500",
				"karpenter.k8s.aws/instance-ebs-bandwidth": "25000",
				"karpenter.k8s.aws/instance-encryption-in-transit-supported": "true",
				"karpenter.k8s.aws/instance-family": "m6idn",
				"karpenter.k8s.aws/instance-generation": "6",
				"karpenter.k8s.aws/instance-hypervisor": "nitro",
				"karpenter.k8s.aws/instance-local-nvme": "474",
				"karpenter.k8s.aws/instance-memory": "32768",
				"karpenter.k8s.aws/instance-network-bandwidth": "12500",
				"karpenter.k8s.aws/instance-size": "2xlarge",
				"karpenter.k8s.aws/instance-tenancy": "default",
				"karpenter.sh/capacity-type": "spot",
				"karpenter.sh/do-not-sync-taints": "true",
				"karpenter.sh/initialized": "true",
				"karpenter.sh/nodepool": "amd64-fallback",
				"karpenter.sh/registered": "true",
				"kubernetes.io/arch": "amd64",
				"kubernetes.io/hostname": "ip-10-0-68-68.us-east-2.compute.internal",
				"kubernetes.io/os": "linux",
				"node.kubernetes.io/exclude-from-external-load-balancers": "karpenter",
				"node.kubernetes.io/instance-type": "m6idn.2xlarge",
				"topology.ebs.csi.aws.com/zone": "us-east-2b",
				"topology.k8s.aws/zone-id": "use2-az2",
				"topology.kubernetes.io/region": "us-east-2",
				"topology.kubernetes.io/zone": "us-east-2b"
			},
			"annotations": {
				"alpha.kubernetes.io/provided-node-ip": "10.0.68.68",
				"csi.volume.kubernetes.io/nodeid": "{\"ebs.csi.aws.com\":\"i-076c6bf8eb1daa3be\"}",
				"karpenter.k8s.aws/ec2nodeclass-hash": "4098229930653828644",
				"karpenter.k8s.aws/ec2nodeclass-hash-version": "v5",
				"karpenter.k8s.aws/instance-profile-name": "prow-build-cluster_6795124765534763251",
				"karpenter.sh/nodeclaim-min-values-relaxed": "false",
				"karpenter.sh/nodepool-hash": "6821555240594823858",
				"karpenter.sh/nodepool-hash-version": "v3",
				"node.alpha.kubernetes.io/ttl": "15",
				"volumes.kubernetes.io/controller-managed-attach-detach": "true"
			},
			"ownerReferences": [
				{
					"apiVersion": "karpenter.sh/v1",
					"kind": "NodeClaim",
					"name": "amd64-fallback-vf6wr",
					"uid": "5be498cb-806c-46aa-8ab9-2a1321aeefbc",
					"blockOwnerDeletion": true
				}
			],
			"finalizers": [
				"karpenter.sh/termination"
			],
```
Generally, we are happy with spots - they provide more capacity and are faster than the machines we used before. 

Graceful handling disruptions with `DisrutionTarget` conditions seems like the k8s-native approach to me.  The condition is also added in a couple other scenarios as you can find in the [official documentation](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-conditions).

upodroid:
Spot was never meant to be turned on; it's been turned off now

mimowo:
> Spot was never meant to be turned on; it's been turned off now

But does it mean that our capacity for running CI jobs is reduced and the machines are slower? Or we preserve the same capacity, just more reliable? TBH if the capacity is reduced I would prefer to use spots.

upodroid:
We retain the same capacity; now nodes will scale down when empty instead of an EC2 interruption

mimowo:
Yeah, I found the PR https://github.com/kubernetes/k8s.io/pull/9765.

I'm not sure how it works on AWS exactly, but I feel in practice the obtainability of the declared capacity may be reduced, wdyt?

mimowo:
@upodroid I think we can move the discussion to slack: https://kubernetes.slack.com/archives/CCK68P2Q2/p1784834959652239

mimowo:
@droslean @Prucek @BenTheElder would you like to consider the contribution as it can be generalized for handling spots (any DisruptionTarget)?

I think this is an easy win for the community as it can help handle any type of retrieable failures (non-initiating from Pod), so please evaluate, but we will ofc respect the decision if no, and we can close the PR.

pkg/plank/reconciler.go
mimowo: @yuluo-yx when you have a chance let's generalize the PR and leave it to the testing leads as discussed at slack: https://kubernetes.slack.com/archives/CCK68P2Q2/p1784837273444169?thread_ts=1784834959.652239&cid=CCK68P2Q2

pkg/plank/reconciler.go
```diff
@@ -671,18 +671,27 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
...
 
 func getPodUnexpectedStopCause(pod *corev1.Pod) PodUnexpectedStopCause {
 	if pod.Status.Reason == Evicted {
 		return PodUnexpectedStopCauseEvicted
 	}
 
+	if podWasTerminatedByKubelet(pod) {
+		// A kubelet-initiated termination indicates an infrastructure disruption,
+		// not a test failure. For example, this can happen when a spot node is
+		// reclaimed. Treat it as unexpected so Plank can revive the pod.
```
droslean, line 690: You are assuming that Prow runs only on Kubernetes. I dont think it would be wise to default this behaviour as a first look. I will need some time to investigate how Openshift treats the nodes and what status its using before moving forward.

droslean:
> @droslean @Prucek @BenTheElder would you like to consider the contribution as it can be generalized for handling spots (any DisruptionTarget)?
> 
> I think this is an easy win for the community as it can help handle any type of retrieable failures (non-initiating from Pod), so please evaluate, but we will ofc respect the decision if no, and we can close the PR. 

Thanks a lot for the contribution!. Let me investigate the behaviour of MCO in Openshift to see if this change won't cause any problem on that infra. Otherwise, I would mostly prefer to have the extra reasons configurable by the user which this is directly connected with the corresponding infrastructure, instead of assuming that Prow will always run on Kubernetes only.

dobesv:
I ran into similar issues and instead of modifying prow core I created a separate controller as recommended by the maintainers:

https://gist.github.com/dobesv/91b43c6396a177e9e73cd9c5757b400c

> But does it mean that our capacity for running CI jobs is reduced and the machines are slower? 

Spot instances are not slower in general, for the same instance type they are the same.  So unless your environment uses different (faster) instance types for spot instances you won't get a performance benefit

> Or we preserve the same capacity, just more reliable? TBH if the capacity is reduced I would prefer to use spots.

Spot instances would have less capacity than on-demand instances, for the same instance type(s).  Spot instance requests will be refused if capacity is low for the given instance type/family at times where on-demand or reserved instances would still get launched.

pkg/plank/reconciler.go
yuluo-yx: got it, I haven't joined the K8s Slack channel yet; I'll post an update once I receive the invitation email and join.

Prucek:
The implementation treats all kubelet-initiated terminations as retriable infrastructure glitches — not just spot reclamation. Kubelet terminates pods for many legitimate reasons:
- Resource pressure (memory, disk, PID eviction) 
- Graceful node shutdown
- Priority-based preemption at the node level
- Node-level OOM

Kubelet does those things for a reason and I don't think we should be going in this direction. 
/lgtm cancel

mimowo:
@Prucek have you had a chance to read documentation for the DisruptionTarget reason on the  condition? That one was designed specifically to mark retriable terminations. Please read :
https://kubernetes.io/blog/2024/08/19/kubernetes-1-31-pod-failure-policy-for-jobs-goes-ga/#kubernetes-initiated-pod-disruptions
and
https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-conditions

The dcoumentation specifically says: *A dedicated Pod DisruptionTarget condition is added to indicate that the Pod is about to be deleted due to a disruption. The reason field of the condition additionally indicates one of the following reasons for the Pod termination:* for the retriable scenarios.

While later for other scenarios:

*In all other disruption scenarios, like eviction due to exceeding [Pod container limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/), Pods don't receive the DisruptionTarget condition because the disruptions were probably caused by the Pod and would reoccur on retry.*

So the reason for the condition is to indicate failures which are likely not to occur on retry.
So for these cases contributors need to now retry build manually.

The purpose of the PR is to reuse an existing max_revivals mechanism to retry automatically without creating a toil on contributors to retry manually.

Ideally we dont retry only on real code bugs, right?

However this is hard to discriminate on all cases. DisruptionTarget is used to say: this PR failuer is worthy retry because likely just an infra failure, not a core bug.

Prucek:
@mimowo. I agree the condition was designed to signal retriable disruptions, and retrying on spot reclamation makes sense.

  However, `TerminationByKubelet` covers both spot/graceful-node-shutdown and node-pressure eviction (memory, disk, PID pressure). For node-pressure eviction, the pod was terminated because the node ran out of
  resources — retrying is likely to hit the same problem again, and auto-reviving could make things worse by adding load to an already-pressured cluster.

mimowo:
>  For node-pressure eviction, the pod was terminated because the node ran out of
resources — retrying is likely to hit the same problem again, and auto-reviving could make things worse by adding load to an already-pressured cluster.

Maybe, but in this case:
1. the mechanism is anyway capped  by max_revivals=3
2. contributors don't check if a PR failed maybe due to PID pressure on node and carefully setting up their times to retry in 10h, most of them just randomly do "retry". Relying on humans being slow is not a good strategy.

yuluo-yx:
I can briefly summarize the points currently under discussion regarding this PR:

> 1. The PR proposes expanding the scope of node shutdown retries to include `TerminationByKubelet`. However, the review feedback notes that `TerminationByKubelet` also encompasses normal behaviors—such as eviction due to node pressure or preemption—and that excessive retries could exacerbate node pressure. 
> 2. Generalizing the scope to `DisruptionTarget` is a point requiring further discussion.
> 3. Cloud infrastructure behaviors may vary across different cloud providers.

In summary, the key questions are: **Should Prow automatically recover from such infrastructure disruptions by default? And how can we provide sufficiently granular control for expensive tasks, different clusters, and various causes of disruption?**

yuluo-yx:
@mimowo @Prucek 

What are our next steps regarding this issue? It seems we haven't reached a consensus—or perhaps I missed some key information. 👀

mimowo:
Yeah, since our flake issues are addressed by no longer using "spot" VMs for our CI jobs we no longer experience flakes. 

I propose to keep the PR open for a week, but if there is no will to move it forward from prow maintainers, then I would just suggest to close it as lazy consensus.

Prucek:
@mimowo @yuluo-yx I would be maybe ok with approving it, if it was a opt-in feature as Ben mentioned. We can ask the maintainers about their opinion @petr-muller @smg247 WDYT?

petr-muller:
From a very high-level PoV I like the feature and would not mind merging it, but I admit I was not paying too much attention because it seemed enough people are engaged here.

mimowo:
Hi folks, maybe to address the concerns about wrong decisions of the new policy we could have a phased approach:
1. introduce the code but only log the decisions - skip retrying
2. evaluate the results - search for logs and ask some expert or AI to evaluate the decisions of the new mechanism as "true positives", but also "false positives"; observe for a month
3. review the stats and decide on the flipping the logs into actionable code

The only issue is - I'm not sure how easy / hard it would be to collect the decisions of the new mechanism from the last month easily, but maybe some AI could help with that.

wdyt?

petr-muller:
@mimowo I like the approach

yuluo-yx:
If there are no issues, I think we can proceed with the merger. 👀 

@petr-muller /cc @mimowo

Issue #789, opened 2026-07-01
Override plugin: follow-up cleanup from sticky override work


smg247:
Follow-up items identified during review of #778:

**1. Consider removing synthetic ProwJob creation in override**

  Now that overrides embed BaseSHA: in the status description (and Tide reads it via BaseSHAFromContextDescription), the synthetic passing ProwJob creation in /override may no longer be necessary.
  Previously, overrides relied on these synthetic ProwJobs to avoid retesting, but they'd get reaped by Sinker after max_prowjob_age, causing unwanted retests.

  With BaseSHA now embedded in the status description, Tide can recognize the override result without needing the synthetic ProwJob. Worth investigating whether the ProwJob creation can be removed as a
  simplification.

  Ref: https://github.com/kubernetes-sigs/prow/pull/778#discussion_r2178175703

**2. Reuse baseSHA from original status description instead of fetching at command time**

  When /override is run, it calls baseSHAGetter to get the current base branch SHA at command time. This may not match the baseSHA that the original job ran against (which is embedded in the job's status
  description via BaseSHA:).

  It would be more correct to extract the baseSHA from the original status description when one exists, and only fall back to fetching the current base SHA when the original status has no embedded
  baseSHA.

  Ref: https://github.com/kubernetes-sigs/prow/pull/778#discussion_r2178173553

PR #808, opened 2026-07-27
override: preserve status base SHA when overriding jobs

SaaiAravindhRaja:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When `/override` handles a failed presubmit status that already carries a `BaseSHA:` marker, it now reuses that original base SHA for both the replacement status description and the synthetic successful ProwJob.

This avoids replacing the job's original base SHA with the base branch SHA at command time, which can differ if the base branch moved after the job ran.

#### Which issue(s) this PR fixes:

Part of https://github.com/kubernetes-sigs/prow/issues/789

#### Special notes for your reviewer:

This keeps the existing synthetic ProwJob behavior intact and only changes the base SHA source used when creating the override result.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

### DIFF
--- a/pkg/plugins/override/override.go
+++ b/pkg/plugins/override/override.go
@@ -515,15 +515,9 @@ If you are trying to override a checkrun that has a space in it, you must put a 
 		return oc.CreateComment(org, repo, number, plugins.FormatResponseRaw(e.Body, e.HTMLURL, user, resp))
 	}
 
-	baseSHA, err := baseSHAGetter()
-	if err != nil {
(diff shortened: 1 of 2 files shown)


SaaiAravindhRaja:
hi! is there anything i can improve on?

Prucek:
I am not sure about this; isn't this solving the same issue: https://github.com/kubernetes-sigs/prow/pull/778 ?

PR #817, opened 2026-07-29
plank: enforce pending timeout when build cluster is unreachable

Prucek:
## Summary

When a build cluster becomes unreachable, `syncPendingJob` enters a retry loop that never terminates:

1. `r.pod()` does an informer cache lookup — pod not found (cache is empty after plank restart while the cluster was down)
2. Enters `!podExists` branch → calls `startPod()` → `client.Create()` (direct API write)
3. Cluster unreachable → network error (e.g. connection refused)
4. `isRequestError()` returns `false` — it only matches 4xx API responses, not raw network errors
5. Returns error → controller-runtime requeues with backoff → repeats forever

The `pod_pending_timeout` is never evaluated because the timeout check only runs after the pod is found in the informer cache and observed in a PodPending phase. When the pod is missing from the cache entirely, neither the scheduling nor the pending timeout code paths are reached.

**Fix:** In the `!podExists` path, when `startPod()` fails with a non-request error, check the ProwJob's `pendingTime` against the configured `pod_pending_timeout`. If the timeout has been exceeded, mark the job as errored instead of retrying forever. If the timeout hasn't elapsed yet, continue retrying (the cluster may recover).

Also extracts `maxPodPendingTimeout` helper to deduplicate the timeout resolution logic between the pod-missing and pod-pending code paths (addresses review feedback from the previous revision).

**Example:** https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-build-farm-canary-build06/2080695972983214080 — this canary job was stuck in pending state since July 24 because build06 was temporarily unreachable. Despite `pod_pending_timeout: 30m` being configured globally, it was never enforced. The prowjob.json confirms: `state: "pending"`, `pod_name` set (pod was originally created), no `finished.json` or `build-log.txt`.

### Why the informer cache read doesn't help

As confirmed in review discussion: `buildClient.Get()` reads from the informer cache — a purely in-memory lookup that never fails with network errors. When a build cluster goes down:
- If plank keeps running: the stale pod stays in cache, existing timeout logic works
- If plank restarts: fresh cache can't list from the unreachable cluster → pod is missing → `startPod()` network error loop (this fix)

## Changes

- **`pkg/plank/reconciler.go`**: Add pending timeout check in the `!podExists` + `startPod` failure path for non-request errors. Extract `maxPodPendingTimeout` helper.
- **`pkg/plank/controller_test.go`**: Add `ExpectError` field to test struct. Add two test cases: "build cluster unreachable, pending timeout exceeded" (expects ErrorState) and "build cluster unreachable, pending timeout not yet exceeded" (expects retry error).

### DIFF
--- a/pkg/plank/reconciler.go
+++ b/pkg/plank/reconciler.go
@@ -465,12 +465,20 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
 		id, pn, err := r.startPod(ctx, pj)
 		if err != nil {
 			if !isRequestError(err) {
-				return nil, fmt.Errorf("error starting pod for PJ %s: %w", pj.Name, err)
+				if pj.Status.PendingTime != nil && r.clock.Since(pj.Status.PendingTime.Time) >= r.maxPodPendingTimeout(pj) {
+					pj.SetComplete()
+					pj.Status.State = prowv1.ErrorState
+					pj.Status.Description = fmt.Sprintf("Pod pending timeout: failed to create pod in build cluster: %v", err)
+					r.log.WithFields(pjutil.ProwJobFields(pj)).WithError(err).Info("Marked job as errored: pod creation failed and pending timeout exceeded.")
+				} else {
+					return nil, fmt.Errorf("error starting pod for PJ %s: %w", pj.Name, err)
+				}
+			} else {
+				pj.Status.State = prowv1.ErrorState
+				pj.SetComplete()
+				pj.Status.Description = fmt.Sprintf("Pod can not be created: %v", err)
+				r.log.WithFields(pjutil.ProwJobFields(pj)).WithError(err).Warning("Unprocessable pod.")
 			}
-			pj.Status.State = prowv1.ErrorState
-			pj.SetComplete()
-			pj.Status.Description = fmt.Sprintf("Pod can not be created: %v", err)
-			r.log.WithFields(pjutil.ProwJobFields(pj)).WithError(err).Warning("Unprocessable pod.")
 		} else {
 			pj.Status.BuildID = id
 			pj.Status.PodName = pn
@@ -548,10 +556,7 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
 
 		case corev1.PodPending:
 			var requeueAfter time.Duration
-			maxPodPending := r.config().Plank.PodPendingTimeout.Duration
-			if pj.Spec.DecorationConfig != nil && pj.Spec.DecorationConfig.PodPendingTimeout != nil {
-				maxPodPending = pj.Spec.DecorationConfig.PodPendingTimeout.Duration
-			}
+			maxPodPending := r.maxPodPendingTimeout(pj)
 			maxPodUnscheduled := r.config().Plank.PodUnscheduledTimeout.Duration
 			if pj.Spec.DecorationConfig != nil && pj.Spec.DecorationConfig.PodUnscheduledTimeout != nil {
 				maxPodUnscheduled = pj.Spec.DecorationConfig.PodUnscheduledTimeout.Duration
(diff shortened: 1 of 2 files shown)


pkg/plank/reconciler.go
```diff
@@ -456,6 +456,24 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
 
 	pod, podExists, err := r.pod(ctx, pj)
 	if err != nil {
+		maxPodPending := r.config().Plank.PodPendingTimeout.Duration
+		if pj.Spec.DecorationConfig != nil && pj.Spec.DecorationConfig.PodPendingTimeout != nil {
+			maxPodPending = pj.Spec.DecorationConfig.PodPendingTimeout.Duration
+		}
```
smg247: this is duplicated from original lines 551-553. I think we could use a helper here incase the logic needs to merge something else in the future.

pkg/plank/reconciler.go
```diff
@@ -456,6 +456,24 @@ func (r *reconciler) syncPendingJob(ctx context.Context, pj *prowv1.ProwJob) (*r
...
+			pj.Status.State = prowv1.ErrorState
+			pj.Status.Description = fmt.Sprintf("Pod pending timeout: could not get pod from build cluster: %v.", err)
+			r.log.WithFields(pjutil.ProwJobFields(pj)).WithError(err).Info("Marked job as errored: build cluster unreachable and pending timeout exceeded.")
+			pj.Status.URL, err = pjutil.JobURL(r.config().Plank, *pj, r.log)
+			if err != nil {
+				r.log.WithFields(pjutil.ProwJobFields(pj)).WithError(err).Warn("failed to get jobURL")
+			}
+			if patchErr := r.pjClient.Patch(ctx, pj.DeepCopy(), ctrlruntimeclient.MergeFrom(prevPJ)); patchErr != nil {
+				return nil, fmt.Errorf("patch prowjob: %w", patchErr)
+			}
```
smg247: This also seems like something that must be duplicated. Can we get a helper to error a pj that is used here?

petr-muller:
Isn't this too strict? Single failed `Get` - where the condition can be transient  - will make Plank _assume_ the Pod never actually started and kill any job that went `Pending` (=we created a pod) before a fairly short period of time. I guess the problem is that we do not distinguish the Pending jobs where we never saw the payload pod started (where we'd want to apply `pod_pending_timeout`) from the Pending jobs where we did (where we'd want to apply a different timeout, like the job timeout itself).

Still I feel something like a grace period is in order.

pkg/plank/reconciler.go
```diff
@@ -450,12 +450,64 @@ func (r *reconciler) terminateDupes(ctx context.Context, pj *prowv1.ProwJob) err
...
 
+func (r *reconciler) maxPodPendingTimeout(pj *prowv1.ProwJob) time.Duration {
+	maxPodPending := r.config().Plank.PodPendingTimeout.Duration
+	if pj.Spec.DecorationConfig != nil && pj.Spec.DecorationConfig.PodPendingTimeout != nil {
+		maxPodPending = pj.Spec.DecorationConfig.PodPendingTimeout.Duration
+	}
+	return maxPodPending
+}
+
+func (r *reconciler) updateProwJobStatus(ctx context.Context, pj *prowv1.ProwJob, prevPJ *prowv1.ProwJob) error {
```
petr-muller: This helper could use a godoc that describes how exactly is the the prowjob status updated, because it is absolutely unclear without reading the code. It should probably retain the comment that was at its original location because the wait could be surprising and a reader can just guess why it is there.

petr-muller (commented):
I dug into the `Get()` failure signal and have several more thoughts - I don't consider them blocking but I'd like them to be considered.

1. I wonder if instead of handling the error in-situ it should be re-translated to a `TerminalError` that's handled almost identically the call chain? We already have the "stop reconciling, set to abort" behavior, just at different place.
2. We don't have the "wait for the change to propagate into the informer cache behavior" in the `TerminalError` handler though -> that kinda defeats (1) but I wonder if the `TerminalError` handler should do the wait too (and then we could do (1))

And lastly, I really wonder if the Pod `Get()` really actually errors if a build cluster dies - if I read the code correctly it is an informer cache-backed read, which is local and not affected by build farm outages - so the `Get` IMO never errors, you just stop getting new state and you stop getting reconcile events about the Pod.

PR #818, opened 2026-07-29
Fix sticky overrides being clobbered by retriggered ProwJobs

smg247:
Two issues prevent /override-sticky statuses from persisting. This takes care of the issues with `trigger`.

Trigger plugin: skip presubmits whose context already has a SUCCESS status in buildAll(). Previously, when LGTM was applied to an untrusted PR, buildAll would create new ProwJobs for all presubmits regardless of current status, clobbering sticky overrides when the new jobs completed.

### DIFF
--- a/pkg/plugins/trigger/pull-request.go
+++ b/pkg/plugins/trigger/pull-request.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	klabels "k8s.io/apimachinery/pkg/labels"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	prowapi "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	"sigs.k8s.io/prow/pkg/config"
@@ -389,11 +390,43 @@ func buildAllButDrafts(c Client, pr *github.PullRequest, eventGUID string, baseS
 	return buildAll(c, pr, eventGUID, baseSHA, presubmits)
 }
 
-// buildAll ensures that all builds that should run and will be required are built
+// buildAll ensures that all builds that should run and will be required are built.
+// It skips presubmits whose context already has a successful GitHub status or
+// check run to avoid redundantly re-running jobs that have already passed.
 func buildAll(c Client, pr *github.PullRequest, eventGUID string, baseSHA string, presubmits []config.Presubmit) error {
 	org, repo, number, branch := pr.Base.Repo.Owner.Login, pr.Base.Repo.Name, pr.Number, pr.Base.Ref
 	changes := config.NewGitHubDeferredChangedFilesProvider(c.GitHubClient, org, repo, number)
-	toTest, err := pjutil.FilterPresubmits(pjutil.NewTestAllFilter(), changes, branch, presubmits, c.Logger)
+
+	successContexts := sets.New[string]()
+	combinedStatus, err := c.GitHubClient.GetCombinedStatus(org, repo, pr.Head.SHA)
+	if err != nil {
+		c.Logger.WithError(err).Warn("Failed to get combined status; running all presubmits")
+	} else if combinedStatus != nil {
+		for _, status := range combinedStatus.Statuses {
+			if status.State == github.StatusSuccess {
+				successContexts.Insert(status.Context)
+			}
+		}
+	}
+	checkRunList, err := c.GitHubClient.ListCheckRuns(org, repo, pr.Head.SHA)
+	if err != nil {
+		c.Logger.WithError(err).Warn("Failed to list check runs; some passing checks may be re-triggered")
+	} else if checkRunList != nil {
+		for _, cr := range checkRunList.CheckRuns {
+			if strings.EqualFold(cr.Conclusion, "success") || strings.EqualFold(cr.Conclusion, "neutral") {
+				successContexts.Insert(cr.Name)
+			}
+		}
+	}
+
+	var filter pjutil.Filter
+	if successContexts.Len() > 0 {
+		filter = pjutil.NewTestAllWithExistingStatusFilter(successContexts)
+	} else {
+		filter = pjutil.NewTestAllFilter()
+	}
+
+	toTest, err := pjutil.FilterPresubmits(filter, changes, branch, presubmits, c.Logger)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/plugins/override/override.go
```diff
@@ -363,6 +377,42 @@ func parseOverrideInput(in string) []string {
...
+			continue
+		}
+		job.SetComplete()
+		job.Status.State = prowapi.AbortedState
+		job.Status.Description = "Aborted by /override-sticky."
+		if _, err := oc.Update(context.TODO(), &job, metav1.UpdateOptions{}); err != nil && !apierrors.IsConflict(err) {
+			log.WithError(err).WithField("prowjob", job.Name).Warn("Failed to abort ProwJob during /override-sticky")
+		} else {
+			log.WithField("prowjob", job.Name).WithField("context", jobContext).Info("Aborted running ProwJob for /override-sticky")
+		}
```
petr-muller: This does not look right to me - this sets a jobs to a final state (`SetComplete`) but there is an underlying Pod running somewhere and Plank needs to terminate it. It won't do anything for a Prowjob stomped like this.

But I guess you did this to prevent racing with crier reporting status for an aborted job? I think override needs to to exactly the same thing like `/test` does when something triggers a job that is already running for the same revision - IIRC it aborts first, waits for Plank to actually reconcile the old job, then it triggers the new one. And maybe I remember this wrong and this is all handled by plank itself which makes your intent much harder.

But i'm pretty sure the running pod leak is a problem so holding.

/hold

pkg/plugins/trigger/pull-request.go
```diff
@@ -389,11 +390,43 @@ func buildAllButDrafts(c Client, pr *github.PullRequest, eventGUID string, baseS
 	return buildAll(c, pr, eventGUID, baseSHA, presubmits)
 }
 
-// buildAll ensures that all builds that should run and will be required are built
+// buildAll ensures that all builds that should run and will be required are built.
+// It skips presubmits whose context already has a successful GitHub status to
+// avoid re-running passing jobs and clobbering sticky overrides.
```
petr-muller: This looks just partially related to sticky override - previously we'd re-run even a normal job-provided PASS and now we don't. Which seems correct but it's a wider change than "avoid rerunning sticky overrides".

petr-muller (commented):
This looks like two separate changes and should be done in two commits (and I'd actually prefer two PRs myself). I think the override doing an abort is trickier than it seems, there needs to be some handover to Plank to terminate the actual workload.

pkg/plugins/trigger/pull-request.go
petr-muller: Actually, does this silently break `/test all`? That's an explicit trigger and I'd expect this to trigger a run over (even a sticky) override.

pkg/plugins/override/override.go
smg247: Yes, it is to prevent the race. I will re-work this part in a separate PR. I think the `trigger` issue is the more important one to solve.

pkg/plugins/trigger/pull-request.go
smg247: Yes, this is more of a fundamental optimization that happens to fix the issue with `sticky-overrides`. I will update the comment to be less specific.

As for breaking `/test all`, I don't think this path is utilized there:

> /test all goes through generic-comment.go's `handleGenericComment`, which calls `pjutil.PresubmitFilter` to build the filter, then `pjutil.FilterPresubmits`. It never calls `buildAll`.
  `buildAll` is only called from the PR event handler — specifically `buildAllButDrafts` on `LGTM/ok-to-test` label events and `buildAllIfTrusted` on `open/reopen/sync/ready-for-review`.

smg247: I still want to dig into this a bit more to make sure I'm not overlooking something...

smg247: looks like the above is correct. It uses the `TestAllFilter` which only filters out tests that match `NeedsExplicitTrigger`. It is unaffected by this change

smg247:
@petr-muller can you give this another look?

pkg/plugins/trigger/pull-request.go
```diff
@@ -389,11 +390,43 @@ func buildAllButDrafts(c Client, pr *github.PullRequest, eventGUID string, baseS
...
+	checkRunList, err := c.GitHubClient.ListCheckRuns(org, repo, pr.Head.SHA)
+	if err != nil {
+		c.Logger.WithError(err).Warn("Failed to list check runs; some passing checks may be re-triggered")
+	} else if checkRunList != nil {
+		for _, cr := range checkRunList.CheckRuns {
+			if strings.EqualFold(cr.Conclusion, "success") || strings.EqualFold(cr.Conclusion, "neutral") {
+				successContexts.Insert(cr.Name)
+			}
+		}
+	}
```
petr-muller: I'm curious why we need this; will there ever be a name that will filter out a would-be-triggered Prowjob? AFAIK Prowjobs successes will only be reported via statuses, at least until someone ports crier to Checks API?

Not a blocker but it's still a potentially useless API call.

pkg/plugins/trigger/pull-request.go
```diff
@@ -389,11 +390,43 @@ func buildAllButDrafts(c Client, pr *github.PullRequest, eventGUID string, baseS
 	return buildAll(c, pr, eventGUID, baseSHA, presubmits)
 }
 
-// buildAll ensures that all builds that should run and will be required are built
+// buildAll ensures that all builds that should run and will be required are built.
+// It skips presubmits whose context already has a successful GitHub status or
+// check run to avoid redundantly re-running jobs that have already passed.
 func buildAll(c Client, pr *github.PullRequest, eventGUID string, baseSHA string, presubmits []config.Presubmit) error {
```
petr-muller: I interrogated an LLM for all cases that end running `buildAll` and most are fine, but there's one that I'm not sure about - when PR gets retargeted to a different branch. Previously that triggered all the new branch presubmits, now the old branch presubmit pass stays around if the two share the context name. It is arguably a corner case and Tide will eventually retest the stale job (because the old context will carry the old branch basesha), but now when we toy with the idea of not letting Tide retest in some cases (for example, a sticky override would be carried over, despite overriding potentially a different job's results).

Would it be feasible/desirable for the retarget case to keep the old `buildAll` semantics?

pkg/plugins/trigger/pull-request.go
smg247: It was some future proofing, agreed that it can be dropped.

smg247:
@petr-muller, I have addressed your remaining comments here

petr-muller (approved):

PR #731, opened 2026-05-25
Bump Kubernetes dependencies to v0.34.x

ameukam:
Update Kubernetes/client-go/controller-runtime dependencies and related transitives. Adjust tests to use k8s.io/utils/diff.Diff and refresh fixtures for updated Kubernetes YAML serialization.

Assisted by GPT-5.5

PR #823, opened 2026-08-09
chore(deps): bump k8s & controller-runtime + compat fixes

petr-muller:
I took Arnaud's compat fixes from #731, just dropped the dependency bump part of the commit. Then I took Dependabot commit 173c78494ff382b34526fe58ed9742b6dc14ea8f from #810. The resulting state needed a small additional compat fix to build.

...and a massive failing lint patch mostly because of the modernize wanting to replace `ptr.To`-like helpers with `new`. I'm not sure I want to do all this in a single PR though, so holding.

Supersedes #731 #810

Assisted by Claude Code


petr-muller:
/hold

The modernize lint patch is massive, I'm not sure I want to mash everything together like this :/

stmcginnis:
This looks very similar to what I ended up working with an agent to adjust for the breaking changes in the newer libraries. It is quite large, but unfortunately I don't think there's a great way to break it down into smaller pieces.

/lgtm

stmcginnis:
Oh, it does look like this missed updating `.go-version`. That should be raised to 1.26 too since the updated use of `new()` was introduced with that version.

petr-muller:
/close 

superseded by #842

PR #825, opened 2026-08-10
announcement: notify transfer-issue consolidation under issue-management plugin

Amulyam24:
With the change introduced via #702, add announcement for the transfer-issue plugin being moved into the
new issue-management plugin.


upodroid:
We should merge this now as the change is active.

@cblecker

petr-muller (approved):

PR #828, opened 2026-08-10
chore(deps): bump vulnerable dependencies

cblecker:
## Summary

- Bumps dependencies across `go.mod`, `hack/tools/go.mod`, and `site/package-lock.json` to resolve **13 of 20** open Dependabot security alerts

### `go.mod` (3 alerts fixed)
| Package | Bump | Alerts |
|---------|------|--------|
| `go-git/go-git/v5` | v5.19.1 → v5.19.2 | #160 (HIGH: symlink follow in worktree ops), #161 (MED: ref name path traversal) |
| `google.golang.org/grpc` | v1.79.3 → v1.82.1 | #155 (HIGH: xDS RBAC and HTTP/2 vulns) |
| `google/cel-go` | v0.26.0 → v0.29.0 | #157 (MED: JSON private fields via NativeTypes) |

### `hack/tools/go.mod` (7 alerts fixed)
| Package | Bump | Alerts |
|---------|------|--------|
| `google.golang.org/grpc` | v1.79.3 → v1.82.1 | #156 (HIGH) |
| `sigstore/rekor` | v1.5.0 → v1.5.2 | #124 (HIGH: OOM via gzip in APK parsing) |
| `sigstore/sigstore-go` | v1.1.4 → v1.2.1 | #154 (MED), #159 (LOW) |
| `sigstore/timestamp-authority/v2` | v2.0.6 → v2.1.2 | #125 (MED: OOM from unbounded metrics) |
| `go-tuf/v2` | v2.3.1 → v2.4.2 | #76 (MED: path traversal) |
| `golang.org/x/net` | already at fix version | #127 (MED: HTML parser DoS) |

### `site/package-lock.json` (3 alerts fixed)
| Package | Fix | Alerts |
|---------|-----|--------|
| `postcss` | ≥8.5.23 | #158 (HIGH: path traversal), #162 (MED: sourceMappingURL read) |
| `picomatch` | ≥4.0.4 | #89 (MED: method injection in POSIX classes) |

PR #830, opened 2026-08-10
prometheus dependency bump to 0.70.1

elmiko:
This change bumps the prometheus dependency and related packages. It also fixes some linting issues related to the update.

It is worth noting that the conversion from `strings.Split` and `strings.SplitN` to `strings.Cut` are specifically ignoring the boolean value to detect if the separate character is found. This is intentional to preserve the original behavior of `Split` and `SplitN` which will return the entire string if the separator is not found.

replaces #811


elmiko:
cc @petr-muller

petr-muller:
it's obviously fine but it seems we're kinda stepping on each other's toes, @cblecker has the linter fixes in https://github.com/kubernetes-sigs/prow/pull/828

elmiko:
doh! my bad. it's fine to close this.

cblecker:
If yours is passing CI and mine isn't, you can take it @elmiko :)

petr-muller (approved):
:shipit:

PR #832, opened 2026-08-11
spyglass/junit: add copy-to-clipboard widget for test names

petr-muller:
Test names in the JUnit lens are rendered as "ClassName: Name", which is awkward to select by hand and yields a string that needs editing before it can be used as a test filter or a search term.

 Add a small copy icon next to every test name that puts the bare test name on the clipboard. The name is carried in a data attribute rather than scraped from the cell, since the cell also holds the expander arrow and, for skipped tests, the skip reason. The icon only appears on row hover.

Click handling is delegated to a single listener on the container, since a run can have thousands of tests and thus thousands of copy icons. It runs in the capture phase: the row's own expand/collapse handler is bound directly to the row, which sits between the icon and the container, so a bubble-phase listener here would run after the row's handler already fired. Capturing lets us stop the click before it gets there.

Assisted by Claude Code

### DIFF
--- a/pkg/spyglass/lenses/junit/junit.css
+++ b/pkg/spyglass/lenses/junit/junit.css
@@ -96,6 +96,19 @@ table.flaky-layout tbody tr.flaky-text:hover {
   vertical-align: middle;
 }
 
+.copy-test-name {
+  cursor: pointer;
+  font-size: 1em;
+  vertical-align: middle;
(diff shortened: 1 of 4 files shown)


BenTheElder (commented):
Can you share an example screenshot?

My TS is rusty but this looks OK.

PR #847, opened 2026-08-12
deck: drop stale Prow-managed contexts with no matching ProwJob

### LINKED ISSUE #824: Deck PR status page shows stale contexts for renamed or removed presubmit jobs.
Deck's PR status page can show a context for a presubmit job that no longer exists, for example after the job was renamed or removed from the repo's config. GitHub still reports the old status against the commit, and since there's no matching ProwJob for it anymore, it stays visible indefinitely in the PR status list.

Contexts belonging to jobs that no longer exist should eventually stop showing up on the PR status page, instead of persisting forever.

**How to reproduce**:

1. Have a presubmit job run and report a status on a PR.
2. Rename or remove that job from the repo's Prow config.
3. Push a new commit to the same PR, or open a new PR against the same repo/branch.
4. Open the PR's Deck status page. The old job's status is still listed, with no current ProwJob backing it.

The root cause is in `getFullPRContext` in `cmd/deck/static/pr/pr.ts`. It builds its list of unified contexts by putting every GitHub-reported status context into a map first, then merging in current ProwJob data on top. Nothing in that flow ever removes an entry, so a context with no corresponding current ProwJob just stays in the map and gets rendered.

This is related to kubernetes/test-infra#36399, but that issue is actually describing a different, though related, problem: stale results in the prow bot's PR comment, which is a separate reporting path from this one. Opening this issue to track the Deck-specific problem accurately.

cc @BenTheElder @petr-muller

Champbreed:
This PR drops such stale contexts when the name matches Prow's job naming convention (pull-, branch-ci-, ci- prefixes), so non-Prow checks are never touched.

getFullPRContext merges GitHub's reported contexts with Prow's live ProwJob list, but never drops a context GitHub still reports if no matching ProwJob exists for it. This happens when a Prow job is renamed or removed. GitHub keeps the old context indefinitely since nothing tells it to retract it.

Known limitation: this matches by naming convention, not by comparing against Prow's actual job config, so it won't catch a renamed job with a non-standard name. Happy to follow up with a config-based check if that's preferred.

Fixes #824
 
cc @BenTheElder @petr-muller

### DIFF
--- a/cmd/deck/static/pr/pr.ts
+++ b/cmd/deck/static/pr/pr.ts
@@ -332,7 +332,7 @@ function createSearchCard(): HTMLElement {
  * all pr contexts and only replaces contexts that have existing Prow Jobs. Tide
  * context will be omitted from the list.
  */
-function getFullPRContext(builds: ProwJob[], contexts: Context[]): UnifiedContext[] {
+export function getFullPRContext(builds: ProwJob[], contexts: Context[]): UnifiedContext[] {
   const contextMap: Map<string, UnifiedContext> = new Map();
   if (contexts) {
     for (const context of contexts) {
@@ -348,6 +348,7 @@ function getFullPRContext(builds: ProwJob[], contexts: Context[]): UnifiedContex
     }
   }
 
+  const buildContextNames = new Set<string>();
   for (const build of builds) {
     const {
       spec: {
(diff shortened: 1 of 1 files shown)


petr-muller:
/hold

I am not convinced we should do this, posted a https://github.com/kubernetes-sigs/prow/issues/824#issuecomment-5279863638 to the issue.

PR #849, opened 2026-08-14
fix: status-reconciler track own config state for dropped deltas

### LINKED ISSUE #848: status-reconciler can permanently miss config changes when delta delivery times out
## Problem

`status-reconciler` can permanently miss a Prow configuration change when it is busy processing an earlier notification.

`config.Agent.Set()` updates the agent's internal current config (`ca.c`) before delivering the corresponding delta to subscribers. Delivery uses a channel with a one-minute timeout. If `status-reconciler` is still busy—for example, listing pull requests or making GitHub API calls—the channel can remain blocked until the timeout expires, and that delta is dropped.

Because the agent's internal config has already advanced, a later delta's `Before` value includes the dropped change. When `status-reconciler` trusts that `Before` value, it has no way to discover that it never processed the intermediate change. The missed change is therefore permanently invisible to it.

This can cause actions tied to config transitions to be skipped. For example, when a new blocking presubmit is added, `status-reconciler` may never call `triggerNewPresubmits` for already-open pull requests. Those pull requests can remain blocked by the newly required context until someone manually triggers the job.

This issue isolates the `status-reconciler` problem discussed in #841; the GitHub-ruleset context-retirement scenario and Tide status-description behavior from that issue are separate concerns.

## Expected behavior

`status-reconciler` eventually processes every effective configuration change, even if one or more intermediate notifications cannot be delivered while it is busy.

Dropped notifications should also be visible to operators rather than failing silently.

## Proposed approach

- Have `status-reconciler` maintain its own `lastProcessed` configuration.
- On each notification, compare `lastProcessed` with the agent's current live configuration from `Config()`, rather than relying on the notification delta's `Before` value.
- Advance `lastProcessed` only after the relevant reconciliation work has been handled successfully.
- Log a warning from `config.Agent.Set()` whenever delivery of a config delta times out and the delta is dropped.

With this approach, if intermediate notifications are lost, the next successfully delivered notification causes a diff from the last configuration actually processed by `status-reconciler` to the current configuration, recovering all missed changes.

## Acceptance criteria

- A test demonstrates that a config notification can time out while `status-reconciler` is busy and that the next notification still causes the skipped config transition to be reconciled.
- Adding a blocking presubmit during such a dropped notification still eventually triggers the new presubmit on applicable open pull requests.
- A timed-out/dropped config delta emits a warning with enough context for operators to diagnose the failure.
- Normal processing does not reconcile the same config transition more than intended.

Prucek:
## Summary

- `config.Agent.Set()` silently drops deltas when the subscriber channel is blocked for over a minute. When this happens, the agent's config advances but the subscriber never sees the notification, making those changes permanently invisible to `status-reconciler`.
- Fix `status-reconciler` to diff against its own `lastProcessed` config rather than trusting the delta's `Before` field, so the next delivered delta captures all missed changes.
- Add a warning log in the config agent when a delta delivery times out.

Fixes #848

Generated with the help of AI (Claude Code)

### DIFF
--- a/pkg/config/agent.go
+++ b/pkg/config/agent.go
@@ -415,6 +415,7 @@ func (ca *Agent) Set(c *Config) {
 			select {
 			case sub <- delta:
 			case <-end.C:
+				logrus.Warn("Failed to deliver config delta to subscriber within 1 minute, delta dropped. Subscriber may miss config changes if it does not track its own state.")
 			}
 			if !end.Stop() { // prevent new events
 				<-end.C // drain the pending event
--- a/pkg/statusreconciler/controller.go
+++ b/pkg/statusreconciler/controller.go
@@ -165,14 +165,35 @@ func (c *Controller) Run(ctx context.Context) {
 		return
 	}
 
+	// lastProcessed tracks the config we last reconciled against. We maintain
+	// this ourselves rather than trusting the delta's Before field, because
+	// config.Agent.Set() silently drops deltas when the subscriber channel is
+	// blocked for over a minute. When that happens, ca.c advances but we never
+	// receive the delta, so subsequent deltas have a Before that already
+	// includes the dropped changes. By diffing against our own lastProcessed,
+	// the next delta that does get through captures everything we missed.
+	var lastProcessed *config.Config
+
 	for {
 		select {
 		case change := <-changes:
 			start := time.Now()
-			log := logrus.WithField("old_config_revision", change.Before.ConfigVersionSHA).WithField("config_revision", change.After.ConfigVersionSHA)
-			if err := c.reconcile(change, log); err != nil {
+
+			if lastProcessed == nil {
+				// First notification: use the delta's Before (the saved state
+				// from --status-path, or empty config if none was configured).
+				lastProcessed = &change.Before
+			}
+
+			current := c.statusClient.Config()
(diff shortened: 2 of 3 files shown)


Prucek:
This seems like a flake: [pull-prow-unit-test](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_prow/849/pull-prow-unit-test/2088177384883752960)

kaovilai (approved):

petr-muller (commented):
I am not a fan of this approach, degrading the delta channel to be just a notification mechanism for the destination to do its own config fetch feels like it breaks a fairly nice abstraction and separation of concern.

To me the issue is the coupling of the channel drain with reconciliation - I believe *that* needs to be fixed.  So either we put some kind of dedicated reader to status-reconciler to manage the work to be done but async to the reconciliation (so that the send never has chance to block and drop), or I think we can actually make the delta sending smarter: make it buffered with a single slot, then if we need to send another delta, we would drain it ourselves (if there's something to read, the consumer have not read it yet), widen the delta (old before, new after) and send it to the channel.

The first one is a bit more complicated, second is simpler but changes the interface (arguably for the better). I have prototypes for both in the works, we can discuss what we like best.

/hold

PR #859, opened 2026-08-17
Migrate to Pub/Sub v2

cblecker:
## Summary

- Migrate the Pub/Sub reporter and subscriber to `cloud.google.com/go/pubsub/v2` v2.6.0.
- Use type-specific service-account credential loading for explicit GCS and Cloud Build credential files.
- Update the integration fake to use v2 publishers and protobuf admin requests.
- Make Spyglass credential and missing-object tests resilient to wrapped dependency errors.

This replaces #858 while preserving the existing Pub/Sub resource names, message payloads, missing-topic user-error classification, and single-stream subscriber behavior.

## Testing

- Targeted unit tests for the Pub/Sub reporter/subscriber, GCS opener, Cloud Build client, Spyglass, and integration fake.
- Targeted race tests for the same packages.
- `make go-unit`
- `make go-lint`
- `make verify-go-deps`
- `make verify` (using Podman for the boilerplate container)

PR #861, opened 2026-08-18
pkg/config/secret: fix TestAddWithParser flake

### LINKED ISSUE #603: flaking test: `TestAddWithParser`
I have seen this twice over the weekend, seems like this test started flaking recently:

```
=== Failed
=== FAIL: pkg/config/secret TestAddWithParser/1 (2.00s)
    agent_test.go:169: expected value 2 from generator, got 1
time="X0X6-0X-X5TXX:04:44Z" level=warning msg="Error loading secret file." error="stat /tmp/547X33X59: no such file or directory" secret-path=/tmp/547X33X59
time="X0X6-0X-X5TXX:04:44Z" level=warning msg="Error loading secret file." error="stat /tmp/XX3958X448: no such file or directory" secret-path=/tmp/XX3958X448
time="X0X6-01-X5T11:04:45Z" level=warning msg="Error loading secret file." error="stat /tmp/1X39581448: no such file or directory" secret-path=/tmp/1X39581448
time="X0X6-01-X5T11:04:45Z" level=warning msg="Error loading secret file." error="stat /tmp/547X33X59: no such file or directory" secret-path=/tmp/547X33X59
time="X0X6-01-X5T11:04:45Z" level=error msg="Error loading secret." error="strconv.Atoi: parsing \"not-a-number\": invalid syntax" secret-path=/tmp/TestAddWithParser030X0674637/001/secret
=== FAIL: pkg/config/secret TestAddWithParser (0.00s)
```

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_prow/601/pull-prow-unit-test/2015379343022231552
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_prow/525/pull-prow-unit-test/2015813939778031616

cblecker:
## What this PR does / why we need it

`pkg/config/secret TestAddWithParser` flakes with `expected value 2 from generator, got 1`. It has been hitting unrelated PRs since at least January, and is currently the sole failure on `pull-prow-unit-test-race-detector-nonblocking` for #859.

This is a real defect in the test, not infrastructure noise: **the test synchronizes on an event that happens before the thing it asserts.**

### Root cause

`parsingSecretReloader.reloadSecret` publishes a new value in two distinct steps:

```go
raw, parsed, err := loadSingleSecretWithParser(p.path, p.parsingFN)  // :72  calls the user's parser
if err != nil { ...; continue }                                      // :73-76  no publish on parse failure
p.lock.Lock()
p.rawValue = raw
p.parsed = parsed                                                    // :80  value becomes visible to get()
p.lock.Unlock()
reloadCensor()                                                       // :82  runs after the store
```

The test's parsing function pushed the parsed value into a channel, and `checkValueAndErr` treated receiving from that channel as "the reload finished", then immediately called the generator. But the parser runs at `:72` and the store lands at `:80`, with no happens-before edge between them. Reading the old value is therefore legal — not a bug in the reloader. `-race` slows goroutine scheduling enough to widen the window, which is why only the race-detector job fails while `pull-prow-unit-test` passes.

### The fix

Synchronize on `reloadCensor` instead of on the parser. It is invoked at `reloader.go:44` and `:82`, in both cases *after* `p.lock.Unlock()`, so a channel send from inside it establishes a genuine happens-before edge with the store. Once the test receives that signal, the Go memory model guarantees `get()` observes the new value — deterministic, not merely likely.

`AddWithParser` hardwires `a.refreshCensorer` as that callback, so the test now constructs `parsingSecretReloader` directly and calls `start()` with its own callback. The public-API coverage that gives up is recovered by a new `TestAddWithParserRegistersSecret`, which has no timing dependence at all — everything it asserts is established synchronously before `AddWithParser` returns.

For the bad-parse case, the test now asserts on the **sequence of published values** rather than sampling the getter at a moment in time. The reload loop is a single serialized goroutine, so every publish emits exactly one signal, in order. Writing a subsequent good value (`3`) and requiring the next published value to be `3` proves nothing was published for the failed parse — any such publish would have been signalled strictly earlier. That converts an unprovable safety property ("never publishes a bad parse") into a liveness property that can be waited on, with **no observation window or sleep anywhere in the test**.

The new `writeSecret` helper bumps the file's mtime explicitly, so two writes can never share a timestamp on a coarse-granularity filesystem — which would otherwise hide the change until the 600-tick forced reload ten minutes later.

The `TestAddWithParser` wrapper that runs the subtest twice in parallel is kept exactly as-is; that is what makes `-race` exercise concurrent `get()` against the reload goroutine.

**No production code changes.**

### What each assertion still catches

| Regression | Caught by |
|---|---|
| Reloader stops noticing file changes | `awaitPublish(2)` / `awaitPublish(3)` time out |
| Wrong value stored (parse result dropped, stale value published) | `awaitPublish` value mismatch, and the `loader.get()` check after each publish |
| A failed parse clobbers the previously good value | `awaitPublish(3)` sees the bad publish first and fails with the actual value |
| `reloadCensor` moved before the store (censorer refreshed with a stale secret) | `loader.get()` check inside `awaitPublish` |
| Data races on `parsed` / `rawValue` | unchanged — two parallel subtests under `-race` |

### Verification

All green locally:

```
go test -race -count=20 ./pkg/config/secret/...                 # ok, 61s
go test -race -count=20 -cpu=1,2,4 ./pkg/config/secret/...      # ok, 182s
go test ./pkg/config/secret/...                                 # ok, 3.6s
```

I also confirmed the test still fails on a real regression rather than merely passing: temporarily deleting the `continue` at `reloader.go:75`, so a failed parse publishes the zero value, makes it fail with exactly `expected published value 3, got 0`. Reverted afterwards.

Runtime is ~3s per subtest (each publish waits up to one 1s reload tick), versus ~2s before.

### Known warts, deliberately not bundled into a flake fix

- `reloadSecret` uses `time.Tick` (`reloader.go:55`) and never stops; the goroutine leaks for process lifetime. Harmless for a daemon, noise in tests.
- `lastModTime` starts at the zero value, so the loop's first tick always re-reads and republishes the current value. The test tolerates this rather than changing behavior.
- Making the reload interval injectable would speed the test up, but adds production surface purely for tests and leans harder on filesystem mtime resolution.

## Which issue(s) this PR fixes

Fixes #603

(Reopened — it was closed as rotten by the lifecycle bot without a fix, and the identical error is still flaking.)

## Special notes for your reviewer

Kept separate from #859 (Pub/Sub v2 migration) so that PR stays scoped. Once this merges, `/test pull-prow-unit-test-race-detector-nonblocking` on #859 should go green.

```release-note
NONE
```

### DIFF
--- a/pkg/config/secret/agent_test.go
+++ b/pkg/config/secret/agent_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package secret
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


petr-muller (approved):
sweet, nice to get rid of this

PR #862, opened 2026-08-19
fix(githuboauth): fall back to fresh session on undecodable cookie

psalajova:
Deck treated an undecodable gorilla session cookie as a 500. CookieStore.New()/Get() still return an empty session in that case (stale cookie after cookie-secret rotation, leftover from another deck, etc.), so OAuth could not finish and private-deck job reruns failed with
```
500 Internal server error Create new session: securecookie: the value is not valid
```

Login, the new token session, logout, and GetLogin now log and continue with a fresh session so a new cookie can be written. The OAuth state cookie on redirect still fails closed (CSRF).

(Also an issue was opened for this in DPTP jira: https://redhat.atlassian.net/browse/DPTP-4383)

### DIFF
--- a/pkg/githuboauth/githuboauth.go
+++ b/pkg/githuboauth/githuboauth.go
@@ -156,8 +156,8 @@ func (ga *Agent) HandleLogin(client OAuthClient, secure bool) http.HandlerFunc {
 		oauthSession.Options.Secure = secure
 		oauthSession.Options.HttpOnly = true
 		if err != nil {
-			ga.serverErrorAndPrint(w, "Creating new OAuth session", err)
-			return
+			// New always returns a session; a decode error only means the existing cookie is invalid (e.g. stale signing key). Continue with this empty session so we can overwrite it.
+			ga.logger.WithError(err).Warning("Failed to decode existing OAuth session cookie, using new session.")
 		}
 		oauthSession.Options.MaxAge = 10 * 60
 		oauthSession.Values[stateKey] = state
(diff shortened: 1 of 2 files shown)


pkg/githuboauth/githuboauth.go
```diff
@@ -156,8 +156,7 @@ func (ga *Agent) HandleLogin(client OAuthClient, secure bool) http.HandlerFunc {
 		oauthSession.Options.Secure = secure
 		oauthSession.Options.HttpOnly = true
 		if err != nil {
-			ga.serverErrorAndPrint(w, "Creating new OAuth session", err)
-			return
+			ga.logger.WithError(err).Warning("Failed to decode existing OAuth session cookie, using new session.")
```
petr-muller, line 160: Continuing here is surprising to anyone who reads the code, and we make an assumption that `New` returns a new session even on errors, which actually [seems to be the case](https://github.com/gorilla/sessions/blob/4355a998706e83fe1d71c31b07af94e34f68d74a/store.go#L69-L99), but it would be good to place an explicit comment that documents the assumption and why is it fine. Having that comment (at all changed places) would clearly communicate continuing on error is intentional and not an omission. The log kinds hints at it, but IMO not strong enough.

pkg/githuboauth/githuboauth.go
```diff
@@ -179,7 +178,7 @@ func (ga *Agent) HandleLogin(client OAuthClient, secure bool) http.HandlerFunc {
 func (ga *Agent) GetLogin(r *http.Request, identifier AuthenticatedUserIdentifier) (string, error) {
 	session, err := ga.gc.CookieStore.Get(r, tokenSession)
 	if err != nil {
-		return "", err
+		ga.logger.WithError(err).Warning("Failed to decode existing token session cookie.")
 	}
 	token, ok := session.Values[tokenKey].(*oauth2.Token)
```
petr-muller, line 185: won't we fail on this type assertion on empty `Values` anyway? and then error out, just with a different, more generic error.

psalajova: ah, you are right

PR #870, opened 2026-08-21
statusreconciler: never drop config deltas

### LINKED ISSUE #848: status-reconciler can permanently miss config changes when delta delivery times out

petr-muller:
config.Agent.Set delivered each config delta to subscribers via a goroutine that waited up to a minute and then silently dropped the delta. A subscriber busy longer than that (the status-reconciler, while reconciling PR statuses) would miss the drop, and because it trusts the next delta's Before field, every config change hidden by the dropped delta was skipped permanently.

Make delivery lossless at the source. Subscribe now owns the channel: it hands back a single-slot buffered channel and returns a receive-only view. Set coalesces into that slot instead of blocking or dropping -- an undelivered {Before, _} is taken back and replaced with {Before, newAfter}. Deltas chain (each Before equals the previous After), so the merged delta is exactly the transition the subscriber still owes, and Before stays trustworthy no matter how long the subscriber was busy.

This removes the per-send goroutine and one-minute timer entirely and fixes the bug for every subscriber, not just the status-reconciler. moonraker and pubsub already track their own last config and are unaffected.

- config.Agent: Subscribe() returns an owned single-slot DeltaChan; Set uses deliverDelta to coalesce; DeltaChan is now receive-only (<-chan Delta)
- moonraker, pubsub, statusreconciler subscribers updated to the new Subscribe

Fixes kubernetes-sigs/prow#848
Alternative to #849 / #871

Assisted by Claude Code (Opus 4.8)

### DIFF
--- a/pkg/config/agent.go
+++ b/pkg/config/agent.go
@@ -35,15 +35,18 @@ type Delta struct {
 	Before, After Config
 }
 
-// DeltaChan is a channel to receive config delta events when config changes.
-type DeltaChan = chan<- Delta
+// DeltaChan is a receive-only channel handed out by Subscribe to deliver config
+// delta events when config changes.
+type DeltaChan = <-chan Delta
 
 // Agent watches a path and automatically loads the config stored
 // therein.
 type Agent struct {
-	mut           sync.RWMutex // do not export Lock, etc methods
-	c             *Config
-	subscriptions []DeltaChan
+	mut sync.RWMutex // do not export Lock, etc methods
+	c   *Config
+	// subscriptions are single-slot buffered channels owned by the Agent; Set
+	// coalesces into them (see deliverDelta) and hands out a receive-only view.
+	subscriptions []chan Delta
 }
 
 // IsConfigMapMount determines whether the provided directory is a configmap mounted directory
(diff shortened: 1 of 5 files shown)


Prucek:
I really like this fix. Simple, clean, and fixes it for all subscribers.
/lgtm
Before merge, I want to mention that this is a breaking API change for anyone vendoring `pkg/config` and calling `Subscribe` directly. I checked `ci-tools` and it doesn't call `Subscribe`/use `DeltaChan`, so it's unaffected, but I haven't checked other consumers.
Given that and the fact #871 is still an open alternative, I'd be fine holding this a bit longer if there's more discussion on which approach to land.
/hold

PR #878, opened 2026-08-25
test/integration: deflake TestDeck against parallel deck restarts

cblecker:
## Summary

`pull-prow-integration` intermittently fails `TestDeck` with a 504 from the ingress. Root cause: `TestRerun` restarts the single-replica deck deployment while running in parallel (to force kubelet to re-sync the updated `job-config` configmap volume), and until the replacement pod is ready, ingress-nginx answers 5xx for deck. `TestDeck` performed a single, unretried `http.Get` and lost that race.

One commit per change so any CI regression is attributable:

- **test/integration: retry TestDeck front page fetch** — `TestDeck` now polls for up to 180s, the same pattern its sibling `TestDeckTenantIDs` already uses for exactly this reason. The `<title>Prow Status</title>` assertion stays outside the poll: a served-but-wrong front page is a genuine failure, not a transient.
- **test/integration: fix nil deref in TestRerun rerun helper** — on a transport-level error, `res` is nil and `res.Body.Close()` panicked, killing the whole test binary. Transport errors are now retried like non-200 responses.
- **test/integration: wait for replacement pods in refreshProwPods** — `refreshProwPods` returned immediately after deleting pods, leaving callers to race the downtime window. It now records the deleted pods' UIDs and polls (up to 2 min) until no old pod remains and a new pod reports Ready. Also handles the previously ignored `labels.Parse` error. This lengthens how long `prowComponentsMux` is held (~10-15s for deck); both callers are long-running parallel tests with minute-scale polls, so this is absorbed.
- **test/integration: fix deck readiness probe timeout** — both deck deployments had readinessProbe `timeoutSeconds: 600` with `periodSeconds: 3`; kubelet runs probes serially per container, so one hung probe could freeze the Ready condition for 10 minutes. Now `3`; deck's `/healthz/ready` is trivial, so flapping risk is negligible.

## Out of scope / known limitations

- `prowComponentsMux` only serializes pod-deleters against each other; it cannot protect parallel readers from restarts. The TestDeck retry loop is what addresses that.
- The integration test binary runs with Go's default 10-minute `-timeout`, which is close to the suite's real runtime (`TestRerun` alone can exceed 3 minutes).

## Test plan

- [x] `go build ./test/integration/...` and `go vet ./test/integration/test/` pass; gofmt clean
- [ ] Rerun `pull-prow-integration` several times (the flake is intermittent, so individual passes are weak evidence). In runs where deck is mid-restart, `TestDeck`'s new retry `t.Logf` lines are positive evidence the retry engaged and absorbed the window.

## AI usage disclosure

This PR was created with the assistance of AI tooling (Claude Code). The author has reviewed and is responsible for all changes.

### DIFF
--- a/test/integration/config/prow/cluster/deck_deployment.yaml
+++ b/test/integration/config/prow/cluster/deck_deployment.yaml
@@ -93,7 +93,7 @@ spec:
             port: 8081
           initialDelaySeconds: 10
           periodSeconds: 3
-          timeoutSeconds: 600
+          timeoutSeconds: 3
       volumes:
       - name: oauth-config
         secret:
--- a/test/integration/config/prow/cluster/deck_tenant_deployment.yaml
+++ b/test/integration/config/prow/cluster/deck_tenant_deployment.yaml
@@ -95,7 +95,7 @@ spec:
             port: 8081
           initialDelaySeconds: 10
           periodSeconds: 3
-          timeoutSeconds: 600
+          timeoutSeconds: 3
       volumes:
       - name: oauth-config
         secret:
--- a/test/integration/test/deck_test.go
+++ b/test/integration/test/deck_test.go
@@ -227,17 +227,31 @@ func getSpecs(pjs *prowjobv1.ProwJobList) []prowjobv1.ProwJobSpec {
 func TestDeck(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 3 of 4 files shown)


cblecker:
/cc @petr-muller
another deck test deflake 🔪

Prucek (approved):

PR #907, opened 2026-08-26
tide: do not mutate ProwJobs in the cache

petr-muller:
Avoid mutating the ProwJob object in the cache while computing an index key. It is probably benign here because the semantics should be the same, but the mutation is a suprising side effect and informer cache objects sould not be mutated by the reader.

Added a commit modernizing sorts in neighboring Tide codel

#904 side quest

Assisted by Claude Code

### DIFF
--- a/pkg/tide/blockers/blockers.go
+++ b/pkg/tide/blockers/blockers.go
@@ -17,10 +17,11 @@ limitations under the License.
 package blockers
 
 import (
+	"cmp"
 	"context"
 	"fmt"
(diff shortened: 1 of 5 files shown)


petr-muller:
hmm maybe we could do the sort in 904 transform function and then the index key does not need to sort at all

/hold

petr-muller:
/hold cancel

nah this is better; transforming the actual resources could result in some spurious updates or something

PR #909, opened 2026-08-27
blunderbuss, rifle: count existing eligible reviewers toward the reviewer count

cblecker:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When a PR moves from draft to ready-for-review (or `/auto-cc` is used), blunderbuss selects `reviewer_count` reviewers without looking at who is already requested on the PR. A reviewer who is eligible per OWNERS but was already added — e.g. `/cc`'d while the PR was a draft — does not count, so blunderbuss still picks a full quota of new reviewers and over-assigns the PR. Repeated draft→ready cycles keep adding reviewers until the pool is exhausted. Rifle's handler is a copy of blunderbuss's selection logic and had the identical bug.

This PR makes both plugins treat pending requested reviewers (`requested_reviewers`) that are in the OWNERS reviewer pool for the changed files as counting toward `reviewer_count` and `max_reviewer_count`, and excludes them from being picked again. When `exclude_approvers: false`, members of the approver pool also count, because approvers are legitimate picks via the existing fallback path. Concretely:

* `pkg/reviewer`: `GetReviewers` gains an `excluded` set that is never selected (in addition to the author), and a new shared `EligibleRequestedReviewers` helper computes the already-requested reviewers that are eligible per OWNERS (logins normalized with `NormLogin`).
* `blunderbuss`/`rifle` `handle()`: the number of new picks is reduced by the existing eligible reviewers, the max cap accounts for them too, and rifle's blame-based fallback excludes them as well.
* OWNERS `required_reviewers` are still requested even when the quota is already met by existing reviewers.
* A configured `reviewer_count: 0` keeps today's no-op semantics.

Users who already submitted a review are out of scope — only pending requested reviewers count.

#### Which issue(s) this PR is related to:

N/A

#### Special notes for your reviewer:

* This is a default behavior change (bug fix), not a new config option.
* This PR was written in part with the assistance of generative AI. I have reviewed and verified every change and can explain each one.
* Suggested verification: `go test ./pkg/reviewer/... ./pkg/plugins/blunderbuss/... ./pkg/plugins/rifle/...` — the new blunderbuss test cases include the reported scenario end-to-end (draft→ready with an eligible reviewer already `/cc`'d picks only the shortfall; a second run adds nobody).

### DIFF
--- a/pkg/plugins/blunderbuss/blunderbuss.go
+++ b/pkg/plugins/blunderbuss/blunderbuss.go
@@ -62,7 +62,7 @@ func helpProvider(config *plugins.Configuration, _ []config.OrgRepo) (*pluginhel
 		logrus.WithError(err).Warnf("cannot generate comments for %s plugin", PluginName)
 	}
 	pluginHelp := &pluginhelp.PluginHelp{
-		Description: "The blunderbuss plugin automatically requests reviews from reviewers when a new PR is created. The reviewers are selected randomly from the OWNERS files that apply to the files modified by the PR.",
+		Description: "The blunderbuss plugin automatically requests reviews from reviewers when a new PR is created. The reviewers are selected randomly from the OWNERS files that apply to the files modified by the PR. Reviewers that are already requested on the PR and present in those OWNERS files count toward the configured reviewer count.",
 		Config: map[string]string{
 			"": reviewer.ConfigString(PluginName, reviewCount),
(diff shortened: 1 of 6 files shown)


cblecker:
/hold
for reviews

Prucek (approved):

PR #929, opened 2026-09-08
Added a case for tarball files on gcs

hector-vido:
`WriterOptionsFromFileName` guesses upload attributes from a file's name. For a file ending in `.gz` it uploads the compressed bytes, strips the extension, and sets `Content-Encoding: gzip` - so `build-log.txt.gz` is stored as `build-log.txt` and GCS decompresses on read. That is the behavior we want for logs: Deck fetches `build-log.txt` and gets plain text, without anything in the chain needing to know it was gzipped.

Archives gain nothing from that trade, `must-gather.tar.gz` is stored as `must-gather.tar` holding gzip bytes; no viewer renders a tar inline, the user simply wants to download the artifact the job produced. What they get is an object whose name no longer matches what was uploaded, and whose bytes depend on whether the client sends `Accept-Encoding: gzip` - a client that does receives a gzip stream in a file named `.tar`.

This change adds an exception for tarballs: when the segment before `.gz/.gzip` is `tar`, the file keeps its full name and is stored with `Content-Type: application/gzip` and no `Content-Encoding`, so GCS serves the bytes untouched.

This will help KDE [Ark](https://apps.kde.org/ark/) to correctly handle these files.
About GCS [transcoding](https://docs.cloud.google.com/storage/docs/transcoding) process.

**Note** that artifacts previously landing at `foo.tar` will now land at `foo.tar.gz`.

### DIFF
--- a/pkg/pod-utils/gcs/metadata.go
+++ b/pkg/pod-utils/gcs/metadata.go
@@ -39,12 +39,15 @@ func WriterOptionsFromFileName(filename string) (string, io.WriterOptions) {
 	segment := segments[index]
 
 	// https://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding
-	switch segment {
-	case "gz", "gzip":
-		attrs.ContentEncoding = new("gzip")
-	}
+	isGzip := segment == "gz" || segment == "gzip"
+	// A .tar.gz is a gzipped archive, not a gzip-encoded .tar: declaring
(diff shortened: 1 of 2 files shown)


dtantsur:
I'd really appreciate if we could fix this bug. As a KDE user, I have to manually rename each and every file I download from the CI.

Prucek (approved):

petr-muller (approved):

Issue #914, opened 2026-09-01
`trigger` cannot evaluate file-dependent test selection on diffs too large


petr-muller:
When a presubmit's `should_run` evaluation needs the changed-file list, a GitHub `422 Unprocessable Entity` from `GET /repos/{owner}/{repo}/pulls/{number}/files` prevents the Prow `trigger` plugin from completing test selection. The observed instance is an `issue_comment` retrigger, which aborts without feedback, but the same limitation should affect every trigger path that needs the changed-file list, including automatic selection on a PR push.

This was observed against a very large open PR where GitHub reports that the diff takes too long to generate.

```text
pull-ci-openshift-origin-main-lint: should run: error getting pull request changes:
return code not 2XX: 422 Unprocessable Entity
```

The GitHub API response for the PR is:

```json
{
  "message": "Server Error: Sorry, this diff is taking too long to generate.",
  "errors": [{
    "resource": "PullRequest",
    "field": "diff",
    "code": "not_available"
  }],
  "status": "422"
}
```

The PR has only two changed files, but is very large: 529,312 additions and 534,905 deletions. GitHub therefore cannot provide its file list.

[`config.NewGitHubDeferredChangedFilesProvider`](https://github.com/kubernetes-sigs/prow/blob/b9f26284d47bc7e7ba0b726c3b13144b4982ab29/pkg/config/jobs.go#L561-L580) is the shared lazy provider for PR changed files. On its first use it calls [`GetPullRequestChanges`](https://github.com/kubernetes-sigs/prow/blob/b9f26284d47bc7e7ba0b726c3b13144b4982ab29/pkg/github/client.go#L2527-L2551), which calls the GitHub `GET /repos/{org}/{repo}/pulls/{number}/files` endpoint. A `run_if_changed` or `skip_if_only_changed` presubmit calls this provider from [`Presubmit.ShouldRun`](https://github.com/kubernetes-sigs/prow/blob/b9f26284d47bc7e7ba0b726c3b13144b4982ab29/pkg/config/jobs.go#L518-L530).

[`pjutil.FilterPresubmits`](https://github.com/kubernetes-sigs/prow/blob/b9f26284d47bc7e7ba0b726c3b13144b4982ab29/pkg/pjutil/filter.go#L209-L246) stops immediately on a `Presubmit.ShouldRun` error; it does not skip that presubmit or continue evaluating the rest:

```go
shouldRun, err := presubmit.ShouldRun(branch, changes, forced, defaults)
[...]

PR #930, opened 2026-09-09
tide: exclude PR from subpool when its changed files cannot be fetched

KR-Ravindra:
## Problem

When the changed files of a single PR cannot be retrieved (GitHub answers `422 Unprocessable Entity` for `GET /repos/{owner}/{repo}/pulls/{n}/files` when the diff is too large to generate), tide drops the **entire** subpool for that org/repo/branch on every sync, not only the affected PR. Every other merge-eligible PR sharing that subpool is blocked from merging for as long as the unreadable PR stays eligible, and tide logs `Error initializing subpool.` each loop.

## Root cause

`pkg/tide/tide.go`, `presubmitsByPull` (function at line 1685 on base `104d452f`): the error from `ps.ShouldRun(sp.branch, c.changedFiles.prChanges(&pr), forceRun, false)` at line 1713 is returned as the error of the whole function (`return nil, err` at line 1715).

`initSubpoolData` (function at `pkg/tide/tide.go:694`) wraps it into `error determining required presubmit prowjobs` at line 698, and `filterSubpools` (function at `pkg/tide/tide.go:666`) logs `Error initializing subpool.` at line 675 and returns without adding the subpool to the filtered pool map.

The same function already tolerates a per-PR failure a few lines earlier: when `GetPresubmits` fails (broken inrepoconfig) the PR is logged and excluded from `sp.prs` and the loop continues. A changed-files failure is the same kind of per-PR problem but was treated as fatal for the subpool.

## Fix

Treat a changed-files failure the same way as a failure to load the presubmits of that PR:

- when `ShouldRun` returns an error, stop evaluating presubmits for that PR, log a warning, drop any presubmits already collected for it, and `continue`;
- only append the PR to `filteredPRs` (which becomes `sp.prs`) after all its presubmits were evaluated successfully.

Other PRs in the subpool keep their required presubmits and the subpool is still initialized. The excluded PR is rediscovered on the next sync and retried, as before.

The `trigger` and `approve` plugins hit the same 422 (see the issue). This PR only addresses the tide side, where the blast radius is the whole merge pool; what the plugins should do when the file list is unavailable is a separate behavior decision and is left for a follow-up.

## How tested

New test case `changed files error excludes only the affected PR from the subpool` in `TestPresubmitsByPull` (`pkg/tide/tide_test.go`): two PRs in one subpool, `GetPullRequestChanges` returns a 422-style error for PR 1 via the existing `ghcInterceptor`. It asserts that PR 100 still gets its `always_run` and `run_if_changed` presubmits, that its changed files are cached, and that only PR 100 remains in `sp.prs`.

The `always_run` presubmit is listed before the `run_if_changed` one on purpose: PR 1 then has one presubmit collected before `ShouldRun` fails on the second, so the assertion on the presubmit map also covers the `delete(presubmits, pr.Number)` cleanup. With that line removed the case fails with an extra `"1": [{"always_run": true, "context": "always"}]` entry in the map.

Before the fix:

```
--- FAIL: TestPresubmitsByPull (0.00s)
    --- FAIL: TestPresubmitsByPull/changed_files_error_excludes_only_the_affected_PR_from_the_subpool (0.00s)
        tide_test.go:3752: unexpected error from presubmitsByPull: error getting PR changes for #1: failed get PR changes: return code not 2XX: 422 Unprocessable Entity
FAIL
FAIL	sigs.k8s.io/prow/pkg/tide	0.079s
```

After the fix:

```
$ go test ./pkg/tide/
ok  	sigs.k8s.io/prow/pkg/tide	1.921s
$ go vet ./pkg/tide/ && gofmt -l pkg/tide/ && ./_bin/golangci-lint run ./pkg/tide/...
0 issues.
$ go build ./pkg/tide/... ./cmd/tide/
```

## Links

- Issue: https://github.com/kubernetes-sigs/prow/issues/914

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/pkg/tide/tide.go
+++ b/pkg/tide/tide.go
@@ -1681,7 +1681,8 @@ func refGetterFactory(ref string) config.RefGetter {
 }
 
 // presubmitsByPull creates a map pr -> requiredPresubmits and will filter out all PRs
-// where we failed to find out the required presubmits (can happen if inrepoconfig is enabled).
+// where we failed to find out the required presubmits (can happen if inrepoconfig is enabled
+// or if the changed files of the PR cannot be retrieved).
 func (c *syncController) presubmitsByPull(sp *subpool) (map[int][]config.Presubmit, error) {
 	presubmits := make(map[int][]config.Presubmit, len(sp.prs))
 
@@ -1696,9 +1697,9 @@ func (c *syncController) presubmitsByPull(sp *subpool) (map[int][]config.Presubm
 			log.WithError(err).Debug("Failed to get presubmits for PR, excluding from subpool")
 			continue
 		}
-		filteredPRs = append(filteredPRs, pr)
 		log.WithField("num_possible_presubmit", len(presubmitsForPull)).Debug("Found possible presubmits")
 
+		var changedFilesErr error
 		for _, ps := range presubmitsForPull {
 			if !c.provider.jobIsRequiredByTide(&ps, &pr) {
 				continue
@@ -1712,7 +1713,8 @@ func (c *syncController) presubmitsByPull(sp *subpool) (map[int][]config.Presubm
 			forceRun := (requireManuallyTriggeredJobs && ps.ContextRequired() && ps.NeedsExplicitTrigger()) || ps.RunBeforeMerge
 			shouldRun, err := ps.ShouldRun(sp.branch, c.changedFiles.prChanges(&pr), forceRun, false)
 			if err != nil {
-				return nil, err
+				changedFilesErr = err
+				break
 			}
 			if !shouldRun {
 				continue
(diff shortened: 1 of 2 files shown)


petr-muller:
@KR-Ravindra this looks good to me, makes things better because a big PR wont get a whole subpool stuck. any thoughts about how to surface the individual failure so the too large PR will not get infinitely stuck without any feedback except maybe the Tide log?

KR-Ravindra:
@petr-muller thanks. Agreed that today the only trace is the `Error initializing subpool.` line, and with this change the excluded PR is simply absent from `sp.prs`, so the status controller treats it as "not in pool" and falls through the query-diff logic; as far as I can tell it never gets a description that names the real cause.

What I would do, in a follow-up so this PR stays a pure blast-radius fix:

1. In `presubmitsByPull`, record the excluded PR and the error on the subpool (e.g. `sp.excluded map[int]string`) instead of just logging it, and raise the log to Warning.
2. Hand that map to the status controller through `c.statusUpdate` next to `poolPRs` and `blocks`, and have `expectedStatus` return `StatusError` with `Not mergeable. Could not determine required presubmits: <err>` for those PRs before the query-diff branch. Same path the blocker issues use, so the reason lands on the PR's tide context and is visible to the author.
3. Append the same text to `Pool.Error` so the tide dashboard shows it for the subpool too.

Happy to fold it into this PR instead if you prefer one change.

petr-muller:
@KR-Ravindra sounds good to me! I also prefer a followup PR, keep the PRs small and focused. I was just curious if we see a path forward.

:shipit:

petr-muller (approved):

Issue #937, opened 2026-09-14
tide: PR excluded for unfetchable changed files gets no status and is retried forever


KR-Ravindra:
## What happened?

Since #930, when GitHub cannot return the changed files of a PR (`GET /repos/{owner}/{repo}/pulls/{n}/files` answers `422 Unprocessable Entity` / `Sorry, this diff is taking too long to generate` for very large diffs, see #914), tide excludes only that PR from the subpool instead of dropping the whole subpool. That fixes the blast radius, but the excluded PR itself now disappears silently:

- `presubmitsByPull` logs a single Warning, deletes the PR from `presubmits` and drops it from `sp.prs` (https://github.com/kubernetes-sigs/prow/blob/94b6cafe5c50e23b53b0727076e957c6cee5eafd/pkg/tide/tide.go#L1725-L1735). Nothing about the exclusion is kept on the `subpool` struct (https://github.com/kubernetes-sigs/prow/blob/94b6cafe5c50e23b53b0727076e957c6cee5eafd/pkg/tide/tide.go#L1909-L1926).
- The sync controller hands the status controller only `blocks`, `poolPRs`, `baseSHAs` and `requiredContexts` (https://github.com/kubernetes-sigs/prow/blob/94b6cafe5c50e23b53b0727076e957c6cee5eafd/pkg/tide/tide.go#L581-L585; struct at https://github.com/kubernetes-sigs/prow/blob/94b6cafe5c50e23b53b0727076e957c6cee5eafd/pkg/tide/status.go#L97-L110). `expectedStatus` therefore treats the PR as "not in pool" and falls into the query-diff branch (https://github.com/kubernetes-sigs/prow/blob/94b6cafe5c50e23b53b0727076e957c6cee5eafd/pkg/tide/status.go#L300-L347). Because the PR fulfils every tide query, `hasFulfilledQuery` is true and the code falls through to the ProwJob lookup, so the `tide` context ends up as `success` / `In merge pool.` or `pending` / `Not mergeable. Retesting: ...` — neither names the real cause, and the first one is actively wrong.
[...]

PR #938, opened 2026-09-14
tide: report excluded PRs on their tide status context

### LINKED ISSUE #777: tide: context checker error silently drops PR instead of reporting via status context
## What happened?

When Tide's `TideContextPolicy.Validate()` finds a context in both `RequiredContexts` and `RequiredIfPresentContexts`, `filterSubpools` errors out and silently skips the entire subpool. No `tide` status context is ever set on the PR, leaving the author with no signal that anything is wrong.

## Error from logs

```json
{"error":"error setting up context checker for pr <N>: contexts ci/prow/e2e-binary-build-success are defined as required and required if present",
 "file":"/src/pkg/tide/tide.go:674",
 "func":"sigs.k8s.io/prow/pkg/tide.(*syncController).filterSubpools.func1",
 "level":"error","msg":"Error initializing subpool."}
```

## How to reproduce

Configure a repo with `from-branch-protection: true` in tide context_options, add a context to both `branch-protection.required_status_checks.contexts` and `tide.context_options.required-if-present-contexts`, then open a PR matching the tide query. Tide logs the error above but sets no status on the PR.

## Expected behavior

**Context checker errors should be reported via the tide status context**, not silently swallowed. When `filterSubpools` hits this error, the PR gets no `tide` status at all. The author has no way to know Tide is broken for their PR. Tide should set a `pending` or `error` status with a description like "Error evaluating merge requirements" so the problem is visible on the PR.

/kind bug
/area tide

KR-Ravindra:
## Problem

Fixes https://github.com/kubernetes-sigs/prow/issues/777
Part of https://github.com/kubernetes-sigs/prow/issues/937 (follow-up to #930 as agreed with @petr-muller)

When `GetTideContextPolicy` fails for one PR (for example the `contexts X are defined as required and required if present` collision from #777), tide drops the whole subpool and the PR never gets a `tide` status context. Nothing tells the author why the PR does not merge; the only trace is the `Error initializing subpool.` log line.

PRs that #930 excludes from a subpool because their changed files cannot be fetched have the milder form of the same problem (#937): the status controller treats them as "not in pool", finds every query fulfilled and reports `In merge pool.` or `Retesting: ...`, neither of which names the cause.

## Root cause

- `pkg/tide/tide.go:717-719` (`initSubpoolData`): a per-PR context checker error is returned as the error of the whole subpool, and `filterSubpools` (`tide.go:675`) logs it and drops the subpool.
- Nothing about an excluded PR reaches the status controller: `statusUpdate` (`pkg/tide/status.go:97`) only carries `blocks`, `poolPRs`, `baseSHAs` and `requiredContexts`.
- The status controller builds its own context checker with the same `GetTideContextPolicy` call (`status.go:789`), so for #777 `expectedStatus` fails with `failed to set up context register` as well and the PR is skipped. That is why no status is set at all, not even a query diff.

## Fix

- `initSubpoolData` excludes only the affected PR, the same way the presubmit and changed-files failures are already handled, and records the reason in a new `subpool.excluded` map. `presubmitsByPull` records the changed-files reason in the same map.
- `Sync` hands the map to the status controller as `statusUpdate.excluded`. It is collected from the unfiltered subpools, because a subpool whose only PR was excluded does not survive `filterSubpools`.
- `expectedStatus` returns `error` / `Not mergeable. Could not determine merge requirements: <err>` (or `... required presubmits: <err>`) for excluded PRs. The check sits after the mergeability check, so a merge conflict is still reported first, and before the context checker is built, since that can fail for the very same reason.

## How tested

`TestInitSubpoolDataExcludesPRWithContextCheckerError` reproduces the issue's mechanism: `conditional` is explicitly required through the tide context options while the head of PR 1 carries an inrepoconfig presubmit with that context and `run_if_changed`. Before the fix:

```
--- FAIL: TestInitSubpoolDataExcludesPRWithContextCheckerError
    tide_test.go:3846: unexpected error from initSubpoolData: error setting up context checker for pr 1: contexts conditional are defined as required and required if present
```

After the fix PR 2 stays in `sp.prs`, PR 1 has no context checker or presubmits, `sp.excluded` holds the reason, and `excludedPRMap` still reports it when the subpool is dropped by `filterSubpools`.

New `TestExpectedStatus` cases and `TestSetStatusReportsExcludedPRs` fail without the `expectedStatus` change:

```
--- FAIL: TestExpectedStatus/excluded_from_pool_reports_the_reason (0.00s)
    status_test.go:910: Expected status state "error", but got "success".
    status_test.go:913: Expected status description "Not mergeable. Could not determine merge requirements: contexts conditional are defined as required and required if present", but got "In merge pool.".
--- FAIL: TestExpectedStatus/excluded_from_pool_reports_the_reason_even_if_the_context_checker_cannot_be_set_up (0.00s)
    status_test.go:907: error calling expectedStatus(): failed to set up context register: contexts conditional are defined as required and required if present
--- FAIL: TestSetStatusReportsExcludedPRs (0.00s)
    status_test.go:2383: Expected state to be "error", was "pending"
```

The existing `TestPresubmitsByPull` case from #930 additionally asserts the recorded reason.

```
$ go test ./pkg/tide/
ok  	sigs.k8s.io/prow/pkg/tide	93.804s
$ go vet ./pkg/tide/ && gofmt -l pkg/tide/ && ./_bin/golangci-lint run ./pkg/tide/...
0 issues.
$ go build ./pkg/tide/... ./cmd/tide/
```

## Links

- https://github.com/kubernetes-sigs/prow/issues/777
- https://github.com/kubernetes-sigs/prow/issues/937
- Follow-up to https://github.com/kubernetes-sigs/prow/pull/930

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

Not in this PR: surfacing the exclusion in `Pool.Error` (item 3 of the plan on #930). Deck does not render that field today, so I left it for a separate change.


KR-Ravindra:
Round 1 self-review.
1. Re-checked against current main (94b6caf): `initSubpoolData` still fails the whole subpool on one PR's context-checker error (`tide.go:717`) and `filterSubpools` drops it with only a log line (`tide.go:674`); the status controller has no way to learn about it since `statusUpdate` carries no per-PR error.
2. The excluded map is built from the unfiltered subpools on purpose: a subpool whose only PR is excluded is dropped by `filterSubpools`, and the second half of `TestInitSubpoolDataExcludesPRWithContextCheckerError` covers that.
3. With the `expectedStatus` check and the `initSubpoolData`/`presubmitsByPull` exclusion removed, the four new or extended tests fail as quoted in the description; with them, `go test ./pkg/tide/`, vet and gofmt are clean locally.
4. Filled in the missing "before" output in the How tested section. `Pool.Error` surfacing stays out of scope since Deck does not read that field.
Marking ready for review; no prow jobs have run yet, so this needs /ok-to-test.

PR #942, opened 2026-09-16
feat(deck): add dark mode support for main prow status dashboard

vovakrk:
- Add FOUC prevention script in head to check localStorage and system preferences
- Add dark mode toggle button in header for main dashboard
- Add theme switcher logic persisting choice in localStorage and syncing logo
- Add dark mode CSS styling for builds table, filter box, histogram, modals, drawer, and headers


vovakrk:
deck dark mode on

Qqkyu (commented):
/lgtm

cmd/deck/template/base.html
```diff
@@ -31,6 +34,17 @@
   <script type="text/javascript" src="/static/extensions/script.js?v={{deckVersion}}"></script>
   <script defer src="https://cdnjs.cloudflare.com/ajax/libs/material-design-lite/1.3.0/material.min.js"></script>
   {{block "scripts" .Arguments}}{{end}}
+  {{if eq .PageName "index"}}
+  <script type="text/javascript">
+    (function() {
+      var theme = localStorage.getItem('prow-theme');
+      var prefersDark = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;
```
michelle192837, line 41: nit: I think this change is fine, but note that there is already a `.DarkMode` argument, and it feels a bit janky to have both.

I like picking up the user/system preferences though and don't think the existing use should necessarily block this. Maybe, can you file an issue noting we have both, and to consolidate them later?

michelle192837 (approved):
/lgtm

/hold

(Holding for comment)

cmd/deck/template/base.html
vovakrk: Good catch! I've created a new issue with a problem statement & proposed consolidation plan.

Issue #5, opened 2022-03-03
Implement doc checkers


listx:
Reference: https://github.com/kubernetes/test-infra/issues/24821#issuecomment-1057524854

As per the [design doc](https://docs.google.com/document/d/17_gyhQipmK_roYqUtwx-hTwi1tn2sd0aG6vdkBn6fKM/edit?usp=sharing), we should implement automated checkers to make sure we uphold minimum quality standard for both existing documentation as well as newly proposed documentation.

There have been some discussions in the design doc already, but feel free to comment on here about which checker(s) to implement first.

jihoon-seo:
/reopen
/lifecycle frozen

Here are tools I saw:

1. K8s doc's [Internal link checking tool](https://github.com/kubernetes/website/tree/main/content/en/docs/doc-contributor-tools/linkchecker)

According to the doc, the tool
- Builds the site and generates output HTML in the `/public` directory of your local `kubernetes/website` repository
- Pulls the `wdjp/htmltest` Docker image
- Mounts your local `kubernetes/website` repository to the Docker image
- Scans the files generated in the `/public` directory and provides command line output when it encounters broken internal links

and its workflow seems similar to the current link checker in this Prow doc repo.
https://github.com/kubernetes-sigs/prow/blob/0d8946d8dccdcbbb1dc00c3c20641b202b0494a8/site/Makefile#L26-L27
https://github.com/kubernetes-sigs/prow/blob/0d8946d8dccdcbbb1dc00c3c20641b202b0494a8/site/check-broken-links.sh#L22-L30

---

2. K8s doc's [`linkchecker.py`](https://github.com/kubernetes/website/tree/main/scripts#linkcheckerpy)

While the above linkcheckers analyze whether the links in the Hugo-generated HTML files are working or not (404, ...),
this linkchecker performs sort of static analysis on markdown files.

---

And yes, it would be great to implement a new linkchecking Prow workflow
- for every PR, and/or
- executed periodically

petr-muller:
/reopen

Does not feel like an issue that should rot - either we think this is useful, or we should explicitly decide we don't want that.

listx:
I think the first option (https://github.com/kubernetes/website/tree/main/content/en/docs/doc-contributor-tools/linkchecker) could be adopted. It is simpler as it already uses the same underlying tool (`htmltest`) that we currently use.

[...]

PR #952, opened 2026-09-21
fix(site): make the docs link checker check links

dklima:
This PR is the first step for issue #5, "Implement doc checkers".

## The problem

The link checker does not check links. No CI job runs it.

`site/Makefile` ran htmltest one time for each HTML file. For a file argument,
htmltest sets `DirectoryPath` to the folder of that file. See
[main.go:133-137](https://github.com/wjdp/htmltest/blob/v0.17.0/main.go#L133-L137).

The site root became the folder of the page. Each absolute link then failed with
`target does not exist`.

To hide those false errors, `site/.htmltest.yml` ignored `^/docs/` and
`^/community/`. Those two rules cover all 196 internal links in
`site/content/en`. `IgnoreExternalBrokenLinks: true` covers the other 355
external links. Almost nothing stayed under test.

htmltest was also not pinned. `make check-broken-links` failed for everyone.

## The changes

| Change | Reason |
|---|---|
| Run htmltest one time on the full tree, with `DirectoryPath: public` | Absolute links now resolve |
| Remove the `^/docs/` and `^/community/` ignores | They hid the real link surface |
| Delete `site/check-broken-links.sh` | The per-file loop was the root cause |
| Pin htmltest in `hack/tools` | Same method as misspell and ko |
| Add `hack/make-rules/verify/links.sh` and a `verify-links` target | Builds the site and the checker from zero |
| Add a `check-broken-links-internal` target with `-s` | Internal links are fast and deterministic |
| Set `IgnoreSSLVerify: false` | It hid an expired certificate on `prow.kubesphere.io` |
| Ignore the Slack links in the footer | 403, then 429 on retry, on all 122 pages |

Two notes:

- The script runs `npm ci`. The Docsy theme sends SCSS through PostCSS, and hugo
  calls PostCSS through `npx`.
- `verify-links` is not in `hack/make-rules/verify/all.sh`. That file runs in
  `pull-prow-verify-lint`, and its image has no hugo and no node.

## Two more commits

`make verify` failed on this branch for two reasons. Both are in separate
commits.

**`Accept the current year in the boilerplate check`.** `get_dates` built its
year list from `range(2014, 2026)`, so the last accepted year was 2025. Every
file added in 2026 fails the check. The upper bound now comes from the current
year.

**`Skip the generated site in the verify checks`.** The site build writes `.sh`
files into `site/public` and generated CSS into `site/resources`.
`verify-file-perms` then failed on `migrate.sh`, and `verify-spelling` failed on
a typo inside generated CSS. This PR makes a site build part of the normal
workflow, so both checks now prune those folders, as they already prune
`node_modules` and `_bin`.

## Test result

```
$ make verify-links
htmltest started at 02:31:04 on public
✔✔✔ passed in 1m39s
tested 122 documents
PASS: No broken internal links detected

$ make verify-go-deps
SUCCESS: ran go mod tidy in root and hack/tools folders
```

`make verify` passes on this branch. The internal-only run tests 122 documents
in 0.57s, with zero errors.

## Next steps, not in this PR

**CI jobs.** Internal and external links fail for different reasons. Internal
links are deterministic, so they can block a build. External links break for
third-party reasons, so they belong in a periodic job. This matches the
[proposal in the issue](https://github.com/kubernetes-sigs/prow/issues/5#issuecomment-1325612367).
The periodic job needs an image with hugo extended, node and Go. Maintainers
must choose that image. I can then open the `kubernetes/test-infra` PR.

**16 dead external links.** The checker found them. They are warnings today.
Most are `kubernetes/test-infra/config/prow/*` paths that moved when Prow became
this repository. The full list is in the comment below. This is content work for
a separate PR. I can do it.

## Notes

Two things to know about how it was made:

1. The root cause was proven, not guessed. Run the old per-file invocation
   against the new config. htmltest prints
   `started ... on public/docs/components/core/tide` and then fails on each
   absolute link.
2. The review found the `// indirect` entry in `hack/tools/go.mod` and the
   `IgnoreSSLVerify` problems.

You can reproduce the whole result with `make verify-links`.

/kind bug


dklima:
The 16 dead external links, for the follow-up PR. They are warnings today, not errors.

**Moved when Prow became this repository** (`github.com/kubernetes/test-infra`):

- `/tree/master/config/prow/cluster`
- `/blob/master/config/prow/cluster/ghproxy.yaml`
- `/blob/master/config/prow/cluster/kubernetes_external_secrets.yaml`
- `/blob/master/config/prow/cluster/prow_controller_manager_deployment.yaml`
- `/blob/master/config/prow/cluster/prow_controller_manager_rbac.yaml`
- `/blob/master/config/prow/cluster/prowjob-crd/legacy/prowjob-schemaless_customresourcedefinition.yaml`
- `/blob/master/config/prow/cluster/build/build_kubernetes-external-secrets_customresource.yaml`
- `/blob/master/config/prow/experimental/controller_manager.yaml`
- `/blob/master/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L84`
- `/blob/master/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L527`
- `/tree/master/gencred`
- `/tree/master/gencred/`

**Other projects:**

- `github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md`
- `github.com/Azure/aks-engine/tree/master/.prowci`
- `github.com/GoogleCloudPlatform/compute-image-tools/tree/master/test-infra`

**Dead hosts, in the adopter list:**

- `https://prow.build-infra.jetstack.net/` (no DNS record)
- `https://prow.kubesphere.io` (certificate expired on 2025-09-26)

To get this list again, run `make verify-links` and read the warnings.

Prucek (approved):
/ok-to-test

PR #954, opened 2026-09-21
docs: repair the dead links in the site content

dklima:
This PR repairs the dead links that issue #5 asked us to find. It does not
depend on #952, and it touches no file that #952 touches.

## What broke

14 links in `site/content/en` answered 404.

The manifests for prow.k8s.io left `kubernetes/test-infra` in March 2026. See
[`e1669ccf2`](https://github.com/kubernetes/test-infra/commit/e1669ccf2), which says: *"this was used by the google operated prow cluster, current
config is in github.com/kubernetes/k8s.io"*. The `gencred` tool left the same repository
in [`a47feb714`](https://github.com/kubernetes/test-infra/commit/a47feb714). Paths in kube-state-metrics, aks-engine and compute-image-tools
also changed.

## Three rules for the new targets

| Rule | Where | Target |
|---|---|---|
| A current page points at the current file | `ghproxy/_index.md`, `components/core/prow-controller-manager.md` | `kubernetes/k8s.io`, under `kubernetes/gke-prow/prow` |
| A historical page points at a pinned commit | `announcements.md`, `prow-secrets.md` | test-infra at `1f8084bc`, the commit before the removal |
| A line anchor becomes a plain file link | `components/cli-tools/generic-autobumper.md` | the job file, with no `#L` |

`announcements.md` is a changelog, so each link must show the file as it was at
that time. `prow-secrets.md` already pins two of its links to a commit, so the
three repaired links in it use the same style.

The two autobumper links broke for two reasons at once: the file moved, and the
line numbers changed. The job name is the stable identifier, so the `#L` anchors
are gone.

## Two changes that need your judgement

1. `generic-autobumper.md` named `post-test-infra-deploy-prow` as the example
   postsubmit. That job no longer exists anywhere in test-infra. I used
   `post-k8sio-deploy-prow-build-trusted-resources`, which is in the same file
   as the periodic example. Tell me if you prefer a different job.
2. The adopter entries for `Azure AKS Engine` and `Daisy (Google Compute Image
   Tools)` now point at the project, because the old sub-paths are gone. Note
   that `Azure/aks-engine` is archived, and its description calls it a legacy
   tool. You may prefer to remove the entry.

## Two links I did not touch

Both are in the adopter list. Neither has a correct URL for me to write, so a
maintainer must decide.

- `https://prow.build-infra.jetstack.net/` has no DNS record. The host is gone.
- `https://prow.kubesphere.io` answers, but its certificate expired on
  2025-09-26. The URL is right. The site is broken.

## Testing

Every new URL answers 200. Extract the links from the diff and check them:

```bash
git diff --no-color -U0 origin/main -- site/content/en \
  | grep '^+' | grep -v '^+++' | grep -oE 'https://[^)> ]+' | sort -u \
  | while read -r u; do
      printf '%s  %s\n' "$(curl -s -o /dev/null -w '%{http_code}' -L --compressed "$u")" "$u"
    done
```

All 17 URLs in the diff return 200.

## Notes

The dead links come from a run of `htmltest` over the built site. #952 makes
that run work and is where the list came from, but this PR stands alone: it only
edits Markdown under `site/content/en`.

/kind bug
/area documentation


site/content/en/docs/components/core/prow-controller-manager.md
```diff
@@ -30,7 +30,7 @@ $ go run ./cmd/prow-controller-manager --help
 
 ### Configuration
 
-* [Deployment manifest](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/prow_controller_manager_deployment.yaml)
+* [Deployment manifest](https://github.com/kubernetes/k8s.io/blob/main/kubernetes/gke-prow/prow/prow-controller-manager.yaml)
 * [RBAC manifest](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/prow_controller_manager_rbac.yaml)
```
Prucek: The RBAC manifest link should be updated as well

dklima: Fixed, thanks.

`kubernetes/k8s.io` has no separate RBAC file. The manifest `kubernetes/gke-prow/prow/prow-controller-manager.yaml` holds the Deployment, the Service, the ServiceAccount, the Role and the RoleBinding.

So the two bullets are now one:

```markdown
* [Deployment and RBAC manifest](https://github.com/kubernetes/k8s.io/blob/main/kubernetes/gke-prow/prow/prow-controller-manager.yaml)
```

Pushed in ce331cee9.

dklima:
@Prucek that link was dead, and your catch was worth more than one line. I swept all 444 external links in `site/content` to see what else the first pass missed.

Three more repairs are now in the branch:

| File | Problem | Fix |
|---|---|---|
| `components/core/prow-controller-manager.md` | RBAC manifest, 404 | The `kubernetes/k8s.io` file holds the Deployment, the ServiceAccount, the Role and the RoleBinding, so the two bullets became one |
| `metrics/_index.md` | `pushgateway_deployment.yaml`, 404 | `kubernetes/gke-prow/prow/pushgateway.yaml` in `kubernetes/k8s.io` |
| `components/deprecated/tackle.md` | ProwJob CRD raw URL, 404 | The same file in `kubernetes-sigs/prow` |

The PR count went from 14 dead links to 17. All 19 URLs in the diff answer 200. The 94 internal page links are clean.

The sweep also found ten URLs that answer incorrectly and have no correct replacement. They are in the PR description, under **Links that need a maintainer decision**. Two of them need your call: the `kubernetes-jenkins` bucket now refuses anonymous reads, and the `upstreamURLBase` example points at a repository that holds no prow config today.

I used `curl` for the sweep, because `site/check-broken-links.sh` needs #952 to run.

Prucek (approved):

PR #956, opened 2026-09-21
feat(trigger): approve workflow runs on push

### LINKED ISSUE #194: Allow `ok-to-test` label to approve GitHub workflow runs for new contributors #25210
_Original Issue: https://github.com/kubernetes/test-infra/issues/25210_

---

**What would you like to be added**:
The ability for Prow to approve GitHub workflow runs after the `ok-to-test` label has been added to a PR via GitHub's API https://docs.github.com/en/rest/reference/actions#approve-a-workflow-run-for-a-fork-pull-request

**Why is this needed**:
New contributors are blocked from running GitHub workflows even after a maintainer adds the `ok-to-test` label to a PR and contributors shouldn't need to rely on GUI access to approve workflow runs knative/test-infra/issues/3057

dklima:
Closes #194.

> **Depends on #955.** This branch contains the commit of that pull request, because the client methods that this change uses come from there. Review only the second commit. GitHub removes the first one from the diff after #955 merges.

GitHub holds the workflow runs of a fork pull request at the `action_required` gate. Prow approved them only for the `/ok-to-test` comment. A new commit creates new pending runs, and nobody approved those.

### Changes

- Approve the pending runs for these pull request events:

| action | approve | reason |
|---|---|---|
| `synchronize` | yes | a new commit creates new pending runs |
| `reopened` | yes | GitHub creates the runs again |
| `ready_for_review` | yes | same trust rule as the build path |
| `edited` with a base change | yes | a base change starts the Actions again |
| `labeled` ok-to-test, human sender | yes | no comment event exists for this |
| `labeled` ok-to-test, bot sender | no | the comment handler approved the runs already |
| `labeled` lgtm | no | the build-once of LGTM runs an untrusted pull request on purpose |
| `opened` | no | an untrusted author waits for the `/ok-to-test` |

- Run the approval in a `defer` after the switch of `handlePR`. The abort of the old jobs and the creation of the ProwJobs do not wait for the poll, and the approval also runs in a repository that has no presubmits. That repository is the case that the issue reports: prow for automation, Actions for CI, zero prow presubmits.
- Poll with `wait.ExponentialBackoff`, 5 attempts over about 30 seconds, because GitHub creates the runs after it sends the webhook. Stop when the runs exist and no run waits for approval. Approve each run one time only.
- Read the pull request and the trust again before each approval. A push or an `/ok-to-test cancel` can arrive during the poll.
- Delete `approveGitHubActionsWorkflowRuns`, which returned a `*sync.WaitGroup` that the caller dropped. The approval now runs in the handler goroutine, which the `WaitGroup` of hook tracks. A graceful shutdown no longer cuts an approval in the middle.

### Operational change to review

A push handler in a repository with `trigger_github_workflows: true` stays in the `WaitGroup` of hook for up to 30 seconds. `GracefulShutdown` waits for that same group, so a shutdown can take up to half a minute more. A burst of pushes collects pending goroutines.

This is not a correctness regression, but the maintainers must accept it. A work queue with a limit is the answer if a deployment measures a problem. I did not build one here, because no other prow plugin has one and it needs shared state between handlers.

### Alternative that I did not take

The `workflow_run` webhook event removes the poll and the race, because the payload carries the head commit. It needs a new handler type in `pkg/plugins/plugins.go`, a new case in `pkg/hook/server.go`, and a change of the GitHub App subscription in every deployment. That is a maintainer decision, not a contributor one.

### Known limit

The check of the head commit and the trust is not atomic. A push can arrive between the check and the approval. The `ok-to-test` model already works this way: after the label exists, every later push is trusted until someone revokes it.

### Testing

```bash
go test ./pkg/plugins/trigger/... ./pkg/github/... -count=1
go test ./pkg/plugins/trigger/... -count=1 -race
make go-lint
```

22 new cases: the event table above, the poll (a late run, a late sibling run, a push without runs, the exact count of list calls), the two revalidation races, and the order of the abort against the poll. An end to end test needs a prow deployment with the `actions: write` permission on a repository that has blocked Actions. The maintainers must validate that part.

### Note


dklima:
A note on the `size/XXL` label: this branch also contains the commit of #955, which adds 274 of the 1231 changed lines. After #955 merges, the diff of this PR drops to about 960 lines, and the size plugin changes the label to `size/XL`.

PR #961, opened 2026-09-21
`peribolos`: use REST for direct collaborators to avoid GraphQL resource limits

hoxhaeris:
**What this PR does / why we need it**:

`ListDirectCollaboratorsWithPermissions` listed direct collaborators using the GraphQL `collaborators(affiliation: DIRECT)` connection with `first: 100`. On repositories in large organizations GitHub rejects that query with `Resource limits for this query exceeded`, because resolving the `DIRECT` affiliation is expensive on GitHub's side regardless of the page size. This caused peribolos to fail configuring collaborators for such repositories, e.g.:

```
failed to configure <org>/<repo> collaborators: failed to list direct collaborators for <org>/<repo>: GraphQL query failed: Resource limits for this query exceeded.
```

This switches the implementation to the REST collaborators endpoint with `affiliation=direct`, which returns the same set of directly-added collaborators (excluding org/team-inherited access) but is not subject to the GraphQL per-query resource budget. Permissions are mapped with the existing `LevelFromPermissions` helper. The now-unused `directCollaboratorsQuery` type and `mapGraphQLPermissionToRepoLevel` helper are removed.

REST and GraphQL were verified to return identical members and permissions on a real large repository (280 direct collaborators, correctly excluding team/org-inherited access).


bradmwilliams (commented):
/lgtm
Confirmed that the changes match the existing behavior, without the overhead and/or limitations, of the GraphQL calls.

pkg/github/client_test.go
```diff
@@ -2407,6 +2407,44 @@ func TestListCollaborators(t *testing.T) {
 	}
 }
 
+func TestListDirectCollaboratorsWithPermissions(t *testing.T) {
```
cblecker, line 2410: This test serves a single response with no `Link` header, so it never exercises the pagination path — which is the whole point of the change. `readPaginatedResultsWithValuesWithContext` breaks out of the loop as soon as `Link: rel="next"` is absent (`client.go:2043`), so this passes just as happily against an implementation that fetches one page and stops. The 280-collaborator repo in the description is three pages at `per_page=100`, and a silently truncated list feeds `configureCollaborators`, which removes anyone not in config — worth pinning down.

Two other assertions worth adding: `r.Method`, and `per_page=100`. Note that `readPaginatedResultsWithValues` doesn't inject a `per_page` default the way `readPaginatedResults` does — it passes `values` straight through — so dropping it from `client.go:4371` would quietly fall back to 30 per page with nothing failing.

`TestListIssues` just above (`client_test.go:981`) has the two-page shape to copy: return the first batch with

```go
w.Header().Set("Link", fmt.Sprintf(`<blorp>; rel="first", <http://%s/someotherpath>; rel="next"`, r.Host))
```

serve a different user from `/someotherpath`, and assert the union. One thing to watch: with two requests the shared `gotAffiliation` variable becomes last-write-wins, so it's better to assert `affiliation` inside the handler.

cblecker (commented):
The REST switch itself looks right to me — reusing `readPaginatedResultsWithValues` and `LevelFromPermissions` keeps this consistent with `ListCollaborators` right above it, `org` is still threaded through so GitHub App installation-token scoping is preserved, and replacing the bespoke `mapGraphQLPermissionToRepoLevel` with the table-tested shared helper drops its `default: return Read` fallback, which is a bonus. One comment inline on the test.

One stray thing outside the diff: `cmd/peribolos/main_test.go:3837` still says `// In real usage, this would only return direct collaborators via GraphQL` on the `fakeCollaboratorClient` stub. Now that the four comments in `main.go` have been updated it's the last GraphQL reference left around this method, and it sits on the test double for exactly the method you reimplemented. Worth folding into the same sweep.

Nothing blocking.

pkg/github/client_test.go
hoxhaeris: Good catch, fixed in d2b1e56a0. The test now serves two pages via a Link: rel="next" header and asserts the union, plus r.Method, affiliation=direct, and per_page=100 inside the handler. Also swept the last stray `via GraphQL` comment on the fakeCollaboratorClient stub in main_test.go. Thanks!

pkg/github/client_test.go
```diff
@@ -2407,6 +2407,59 @@ func TestListCollaborators(t *testing.T) {
...
+				t.Fatalf("Marshal error: %v", err)
+			}
+			fmt.Fprint(w, string(b))
+		default:
+			t.Errorf("Bad request path: %s", r.URL.Path)
+		}
+	}))
+	defer ts.Close()
+	c := getClient(ts.URL)
+	perms, err := c.ListDirectCollaboratorsWithPermissions("org", "repo")
```
cblecker: Now that pagination is actually exercised, it'd be worth pinning the failure path too. `result` is accumulated across pages in the closure, so the only thing keeping a partial map from escaping is the `return nil, err` at client.go:4386 — nothing here would catch a refactor that returned `result, err` instead, which is a natural-looking mistake given the map is already built.

That matters because `configureCollaborators` treats this map as the complete current set. A subtest that serves page 1 with the `next` link and then 500s on page 2, asserting both `err != nil` and that `reader` is absent from the returned map, would lock it down. Non-blocking, but cheap while you're in here.

pkg/github/client.go
```diff
@@ -4349,46 +4349,15 @@ func (c *client) ListCollaborators(org, repo string) ([]User, error) {
...
+// ListDirectCollaboratorsWithPermissions gets direct repository collaborators with their permissions.
 // This only returns users who were explicitly added as collaborators, not those with inherited org/team access.
 //
-// See GraphQL schema: repository.collaborators(affiliation: DIRECT)
+// It uses the REST collaborators endpoint with affiliation=direct rather than the GraphQL
+// collaborators(affiliation: DIRECT) connection: on repositories in large organizations the GraphQL
+// query exceeds GitHub's per-query resource budget ("Resource limits for this query exceeded") because
+// resolving the DIRECT affiliation is expensive, while the REST endpoint is not subject to that limit.
+//
+// See https://docs.github.com/en/rest/collaborators/collaborators#list-repository-collaborators
```
cblecker, line 4362: I'd rather this comment carried the evidence than the doc link. The page it points at defines `direct` as "all collaborators with permissions to an organization-owned repository, regardless of organization membership status", and says outright that there's no way to differentiate an organization-level grant from a repository-level one in the response — so it doesn't actually support the direct-only claim on the line above, which is what `configureCollaborators` leans on when it removes everyone not in config (cmd/peribolos/main.go:1305-1318).

What does support it is the check in your PR description: REST and GraphQL returning identical sets on a 280-collaborator repo. That's the load-bearing fact, and it's the one thing that won't be in the tree. Could you fold it in? Roughly: note that the REST reference is loose here and can't distinguish org-level from repo-level grants, and that the direct-only behaviour was confirmed empirically against the GraphQL results.

Two smaller things in the same block: "because resolving the DIRECT affiliation is expensive" and "the REST endpoint is not subject to that limit" are both claims about GitHub's internals we can't verify from here. I'd phrase the first as observed behaviour ("has been observed to fail with ...") and drop the guarantee in the second.

Worth keeping the asymmetry in mind while wording it — returning too few collaborators just leaves stale access in place, but returning too many means we actively call `RemoveCollaborator` on org owners and base-permission members.

pkg/github/client_test.go
```diff
@@ -2407,6 +2407,59 @@ func TestListCollaborators(t *testing.T) {
...
+		}
+		switch r.URL.Path {
+		case "/repos/org/repo/collaborators":
+			if got := r.URL.Query().Get("affiliation"); got != "direct" {
+				t.Errorf("Expected affiliation=direct, got %q", got)
+			}
+			if got := r.URL.Query().Get("per_page"); got != "100" {
+				t.Errorf("Expected per_page=100, got %q", got)
+			}
+			b, err := json.Marshal([]User{
```
cblecker: The fixture covers Read/Write/Admin but not Maintain or Triage — which are the two the deleted `mapGraphQLPermissionToRepoLevel` handled by explicit string match, and the two where `LevelFromPermissions` is order-sensitive (helpers.go:74-88 checks Admin → Maintain → Push → Triage → Pull, and a real maintain payload sets `pull`, `triage`, `push` and `maintain` all true). A reordered branch there would silently downgrade Maintain to Write, which means peribolos never matches config and re-issues `UpdateCollaborator` on every run.

Adding a `triager` and a `maintainer` to the page-2 fixture covers it. Separately, it'd be worth serving the pages as raw JSON literals instead of `json.Marshal([]User{...})` — right now the fixture round-trips through the same struct tags it's meant to exercise, so a wrong tag on `RepoPermissions.Triage`/`Maintain` would be invisible. That's less of a live worry since `ListCollaborators` already decodes the same endpoint into `[]User`, but the literal is better documentation of the REST shape either way.

cblecker (commented):
Thanks for the REST switch — reusing `readPaginatedResultsWithValues` and `LevelFromPermissions` instead of hand-rolling a second mapping table is the right shape, and it brings this call under the client's throttling/retry path that the GraphQL version bypassed.

A few things on test coverage and on the doc comment's justification, none of them blocking.

pkg/github/client_test.go
cblecker: Thanks, the two-page version is what I was after.

One correction to the comment, and it's on me for how I phrased the original: truncation here doesn't cause wrongful removal. The removal loop iterates the *fetched* set (cmd/peribolos/main.go:1307), so a short list produces *fewer* `github.None` actions — stale collaborators silently kept — plus redundant PUTs for the config entries that fell off the missing page (main.go:1277-1295). There is a wrongful-removal path, but it runs through `--dump` writing a truncated config that gets applied later, which is a different chain than the one described here.

Also, `configureCollaborators` is unexported in cmd/peribolos and this is pkg/github, so a rename over there leaves this comment stale with no compiler signal. Maybe phrase it in terms this package owns:

```go
// Serve two pages so results are actually accumulated across pages. A truncated
// list would silently under-report current collaborators, so callers reconciling
// against a desired config would drift instead of converging.
```

pkg/github/client.go
hoxhaeris: Reworded. It now states the load-bearing fact (affiliation=direct returned the same set as GraphQL `collaborators(affiliation: DIRECT)` on the 280-collaborator repo) and notes the REST reference is loose and cannot distinguish an org-level grant from a repo-level one. Phrased the GraphQL failure as observed behavior and dropped the claim about REST internals. In 49e12d33e.

pkg/github/client_test.go
hoxhaeris: Added a subtest that serves page 1 with the `next` link, 500s on page 2, and asserts both `err != nil` and that `reader` is absent from the returned map. Mutation-checked it: switching the wrapper to `return result, err` makes it fail with `got map[reader:read]`, so it pins the `return nil, err`. In 49e12d33e.

pkg/github/client_test.go
hoxhaeris: Done. Page 2 now includes a triager and a maintainer so all five levels are covered, exercising the order-sensitive Admin -> Maintain -> Push -> Triage -> Pull path. Also switched both pages to raw JSON literals so a wrong `RepoPermissions` tag surfaces here instead of round-tripping through the tags under test. In 49e12d33e.

pkg/github/client_test.go
hoxhaeris: You are right, thanks. Reworded to the package-owned version (no `configureCollaborators` reference): it now describes truncation as silently under-reporting current collaborators, so callers reconciling against a config drift instead of converging. In 49e12d33e.

petr-muller:
Skimmed & LGTM but I'll leave this for @cblecker

cblecker (approved):

PR #962, opened 2026-09-22
config: reject partial job config directory walks

petr-muller-author:
When a live job-config directory changes during filepath.Walk, the callback receives filesystem errors. ReadJobConfig currently logs and discards those errors, then returns and publishes whatever partial job set was loaded. For destructive consumers such as sinker, that can make active periodics appear removed and allow their last ProwJobs to be garbage-collected; horologium then immediately recreates them.

Collect walk errors with the other load errors and reject the partial result. The config agent already retains its last good config after a failed reload. Continue walking after an error so one reload reports all observed failures.

Add a regression test that loads a valid file, injects a subsequent checkout-disappearance error, and verifies the partial config is rejected.

Tests:
- `go test ./pkg/config -run "TestReadJobConfig(RejectsWalkErrors|ProwIgnore)$" -count=1`
- `go test ./cmd/sinker ./pkg/config`


petr-muller:
/close

Sorry this was open by mistake

